PR #182, opened 2022-11-10
Convert AllowedIPs to be a slice

nerdalert:


nerdalert:
Since this is slightly invasive I want to run through a few more deployments. Ive tested over 50 nodes with no issues but want to let it sit overnight.

nerdalert:
I think this is g2g after testing with the team.

vishnoianil (approved):

PR #185, opened 2022-11-12
Update new peers only from peer listings

### LINKED ISSUE #139: Implement the routes as a new peer join as opposed to downing wg0 on join
This will prevent a re-convergence every time there is a new node joining. It was fine for a small number of nodes but convergence in the hundreds is painful so it makes sense to bump this one up and try and hit it before the demo.

nerdalert:
- resolves #139
- nodes now only update new peers, the result is zero interruption for existing peers.
- added 2 seconds to some e2e due to more pubkey lookups but this is for initial provisions only and negligable compared to the improvement of leveraging the peer cache by the agent.
- change wgconfig.peer to peers with a new ini annotation. OSX and windows still use wg conf file while linux no longer does.

### DIFF
--- a/internal/apex/apex.go
+++ b/internal/apex/apex.go
@@ -20,8 +20,6 @@ const (
 	wgBinary     = "wg"
 	wgGoBinary   = "wireguard-go"
 	wgWinBinary  = "wireguard.exe"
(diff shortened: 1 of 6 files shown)


nerdalert:
Validated on 50 nodes as well as OSX.

PR #192, opened 2022-11-17
ui: Drop ID from Zone list

russellb:
Zones already have a friendly name, so drop the ID from the Zone list UI. The ID can still be viewed on the Zone detail view after a Zone is selected.


russellb:
Screenshot from 2022-11-17 09-30-36

PR #193, opened 2022-11-17
apex: Send controller the hostname for a device

russellb:
Have the apex client send the Hostname to the controller when creating a new Device.

In the UI, replace the Device ID with the Device's Hostname in the Device list. The ID is still visible in the Device details page.

### DIFF
--- a/internal/models/device.go
+++ b/internal/models/device.go
@@ -12,6 +12,7 @@ type Device struct {
 	PublicKey string      `gorm:"uniqueIndex" json:"public_key"`
 	Peers     []*Peer     `json:"-"`
 	PeerList  []uuid.UUID `gorm:"-" json:"peers" example:"97d5214a-8c51-4772-b492-53de034740c5"`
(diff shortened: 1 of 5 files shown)


russellb:
Screenshot from 2022-11-17 10-48-54


Screenshot from 2022-11-17 10-49-20

internal/models/device.go
```diff
@@ -12,6 +12,7 @@ type Device struct {
 	PublicKey string      `gorm:"uniqueIndex" json:"public_key"`
 	Peers     []*Peer     `json:"-"`
 	PeerList  []uuid.UUID `gorm:"-" json:"peers" example:"97d5214a-8c51-4772-b492-53de034740c5"`
+	Hostname  string      `json:"hostname"`
```
dave-tucker, line 15: Please also add `example:"fedora"` so we get a pretty representation in the API docs

dave-tucker (approved):

PR #196, opened 2022-11-17
Stash usernames in our DB and show them in the UI

russellb:
This patch stashes the UserName in the request context in the auth middleware, and then users it when creating a user in our database. The UI is also updated to show the username in the User list instead of IDs. The ID is still visible in the user detail view.

Screenshot from 2022-11-17 17-49-41
Screenshot from 2022-11-17 17-49-58

### DIFF
--- a/internal/handlers/user.go
+++ b/internal/handlers/user.go
@@ -11,6 +11,9 @@ import (
 	"gorm.io/gorm"
 )
 
+// key for username in gin.Context
+const AuthUserName string = "_apex.UserName"
+
 func (api *API) CreateUserIfNotExists() gin.HandlerFunc {
 	return func(c *gin.Context) {
 		userID := c.GetString(gin.AuthUserKey)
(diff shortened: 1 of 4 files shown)


dave-tucker:
@russellb I thought about this, but I'm not sure it's the right approach.
We're effectively building a cache as we're not going to be notified of username changes from the auth provider.
A more robust approach would be for the backend to somehow populate userinfo periodically.
I know this makes the UI nice for now - so I'm supportive of this overall - but I want to make sure that we capture in issues that this isn't the intended state long term.

russellb:
Thanks @dave-tucker. Yeah, I meant to add a comment here that this felt like a bad hack. 

I think I could update it to validate that it’s still up to date every time the auth middleware checks to make sure we have the user in the DB. That might be a nice compromise over a periodic task. Inactive users would still go stake though.

russellb:
I updated this so it updates our user cache if we detect that the username has changed. It will still go stale if an updated user never comes back to the apex UI.

dave-tucker:
LGTM

Issue #199, opened 2022-11-18
IT for symmetric NAT testing


nerdalert:
We need to be able to test symmetric NAT in IT. I think I have a working topology setup via containers that will be deployable in CI. Just need to codify it. #25 is a start for it.

PR #200, opened 2022-11-19
Added e2e IT for NAT

nerdalert:
- Provisions a apex five-node deployment with seven containers total. The additional two containers are doing ip masquerading to setup a "easy" NAT scenario.
- This emulates the topology of two remote sites, both with firewalls facing the Internet with no inbress traffic allowed or any static NAT mappings from the outside to the inside interface of the NAT nodes.
- What is validated in the test is in order for spoke traffic traversal to occur between net1 and net2, spoke nodes are relayed through the relay node.
- Also validated is local address ICE pairing is reachable between nodes.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -115,8 +115,25 @@ jobs:
         name: ${{ matrix.deployment }}-logs
         path: docker-compose.log
 
+  go-e2e:
+    needs: lint
+    name: go-e2e-integration
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    strategy:
+      fail-fast: false
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: 1.19
+      - name: Run go-e2e
+        run: |
+          make go-e2e
+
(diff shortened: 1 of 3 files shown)


nerdalert:
Resolves #199 

Example output from a relay node dump where you can see two sockets opened up from each network with the same IP but different port which enables the connectivity:

```
        wW6HhRQxY4vhhzOZFgHtXZ+842quEyOH8byAoxzXfmU=	(none)	172.17.0.3:11097	10.29.0.4/32	1668824491	948	860	off
        HhZK5yMqh4T73zqEHIT0qWl1S7ZfunrUfa5lcn1Y4Qs=	(none)	172.17.0.3:51820	10.29.0.2/32	1668824491	1204	1116	off
        uoLerL9yx7RwIa+RQGr6LvMZGsdRuFD/DyJhggfwMlw=	(none)	172.17.0.4:26600	10.29.0.5/32	1668824491	692	604	off
        7ZwvxmSp5Fxv6rK+LkXlkNsmyi+KdepJ1e+qBIDcLHs=	(none)	172.17.0.4:51820	10.29.0.3/32	1668824491	1460	1372	off


```

PR #202, opened 2022-11-22
ui: Use friendly names in more places

russellb:
This PR updates the UI to use friendly names for resources in more places.

### DIFF
--- a/ui/src/App.tsx
+++ b/ui/src/App.tsx
@@ -77,9 +77,9 @@ const App = () => {
       title="Controller"
       layout={Layout}
     >
-      <Resource name="users" list={UserList} show={UserShow} icon={UserIcon} />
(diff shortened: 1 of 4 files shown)


russellb:
I just added another commit which adds some additional updates to the Devices page that were a bit more complicated. Instead of a list of Peer IDs, it lists the zones the device is peered with by name. If you click the zone name, it will take you to the Peer details.

Screenshot from 2022-11-22 11-44-06

russellb:
and now with this last change, there are no more IDs shown on any of the Resource list pages. The IDs can be seen if you go into the resource detail pages. Later it may be useful to add the ability to filter on IDs, even if we're not showing them.

Screenshot from 2022-11-22 12-17-43

PR #206, opened 2022-11-28
[POC] Add an IPSec provider

nerdalert:
- Adds support for strongswan which is a decent initial candidate as it supports draft-brunner-ikev2-mediation for mediating between left and right peers behind NAT for UDP hole punching.


nerdalert:
This was a POC, will be closing to have it as needed if an IPSec driver is decided to be of interest.

Issue #170, opened 2022-11-08
apex: Add Keepalives


dave-tucker:
We should add a `PATCH /api/peer/:id` endpoint to the API.
The payload would be something like `{ "last_seen": 123457676 }`
The client SHOULD ping this API once every N minutes - lets say 10 minutes as a starting point.
This would update a `LastSeen` timestamp on the `Peer`.

On the controller side, we should have a SQL stored procedure that gets called once every 30 mins.
If `Peer.LastSeen` is at `time.Now() - (30 * time.Minutes)` then we go ahead and remove that `Peer` from the database.

This is related to #153 since the code for Peer deletion will need to be written - ensuring that any IP addresses or child-prefixes are released.

Issue #113, opened 2022-10-30
[Bug] Discovery flagged a node as reachable when it was not


nerdalert:
Hit the first bug on mesh discovery. In this scenario `DEBU[1818] peer [ 192.168.208.5 ] is reachable` was not reachable but came back reachable from the probe in the debug log if the client that could not reach it. 

I know that the probe/discovery is temporary but I wanted to capture it anyhow in case anyone hits a random node that gets populated as a peer instead of getting bounced to a hub. Below is what to look for in the logs as the probe is looking for reachable endpoint IPs (`ax.EndpointIP`).

```
DEBU[1818] peer [ 172.31.88.123 ] is not reachable
DEBU[1818] peer [ 10.150.0.23 ] is not reachable
DEBU[1818] peer [ 192.168.1.65 ] is reachable
DEBU[1818] peer [ 192.168.208.5 ] is reachable
DEBU[1818] peer [ 3.86.167.187 ] is reachable
DEBU[1818] peer [ 10.150.0.25 ] is not reachable
```

This is also a good example of how easy this is to troubleshoot. If a node in the zone is unreachable, you simply check the`wg0.conf` file on that node and see if there is a peer entry. See the node's config below that couldn't ping the endpoint.

The troubleshooting flow is pretty simple:

- If there **IS** a peer entry, check the `Endpoint` IP and see if you can ping it. If you can't which was the case here, I knew discovery bugged since that is a private address `192.168.208.5:51820` and unreachable from the node and should never have gotten populated in the wg0.conf file.

- If there **IS NOT** a peer entry, you would start troubleshooting the hub router that the traffic would be draining to for the subnet, in this example `10.185.0.0/20` pointing to the hub router at `3.86.167.187:51820`. The most likely cause would be no connectivity to the hub-router endpoint IP.

```
sudo cat /etc/wireguard/wg0.conf
[Interface]
PrivateKey = OIoAtn6qwl9mWLsYgWQUHaqFpbo66/7ZMS69BOZd93U=
Address    = 10.185.0.8
ListenPort = 51820
SaveConfig = false
PostUp     =
PostDown   =

[Peer]
PublicKey           = 6H9Hr1QfydUgCbfECxoxNUOWhHCDm1z3oOQhoSYwCD8=
Endpoint            = 3.86.167.187:51820
AllowedIPs          = 10.185.0.0/20
PersistentKeepAlive = 25

[Peer]
[...]

PR #208, opened 2022-11-30
Adds full mesh capabilities via reflexive addresses

nerdalert:
- Adds a keepalive managed by the agent and disables wireguard keepalives #170
- Now supports full mesh via reflexive addresses. State is pushed from the relay node since all nodes can establish a connection there and then reconciled by the agent depending on the agent environment. #113
- Discovers symmetric NAT and builds peers to only relay and local peers.
- The agent now manages what port wireguard will listen on to help with reflexive
  NAT socket mapping and avoid overlap to be more compatible with more routers.
  The exception is the hub-router/relay node that needs to have a well-known port
  for all nodes to be able to dial into for state distribution until apex can
  do NAT whole-punching independent of a relay node.
- Adds a `--relay-only` bool switch in the case of nodes that Apex is unable to
  discern is behind a symmetric NAT. Temporary but may be useful for unforeseen
  use cases/environments.
- Adds support for child-prefixes in hub/relay zones.

### DIFF
--- a/cmd/apex/main.go
+++ b/cmd/apex/main.go
@@ -42,7 +42,7 @@ func main() {
 			},
 			&cli.IntFlag{
 				Name:     "listen-port",
-				Value:    51820,
+				Value:    0,
 				Usage:    "port wireguard is to listen for incoming peers on",
 				EnvVars:  []string{"APEX_LISTEN_PORT"},
 				Required: false,
@@ -78,6 +78,12 @@ func main() {
 				Usage:    "set if this node is to be the hub in a hub and spoke deployment",
 				Value:    false,
 				EnvVars:  []string{"APEX_HUB_ROUTER"},
+				Required: false,
+			},
+			&cli.BoolFlag{Name: "relay-only",
+				Usage:    "set if this node is unable to NAT hole punch in a hub zone (Apex will set this automatically if symmetric NAT is detected)",
+				Value:    false,
+				EnvVars:  []string{"APEX_RELAY_ONLY"},
 				Required: false,
 			},
 			&cli.StringFlag{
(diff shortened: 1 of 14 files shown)


nerdalert:
For testing and a bit more info:

### x86
```
wget https://apex-net.s3.amazonaws.com/apex-dev-linux-amd64
chmod +x apex-dev-linux-amd64
sudo ./apex-dev-linux-amd64 http://<running_controller>:8080/
Login: kitteh1
Pass: floofykittens
```

### ARM
```
wget https://apex-net.s3.amazonaws.com/misc/apex-dev-linux-arm64
chmod +x apex-dev-linux-arm64
sudo ./apex-dev-linux-arm64  http://<running_controller>:8080/
Login: kitteh1
Pass: floofykittens
```

### Darwin/OSX
```
wget https://apex-net.s3.amazonaws.com/misc/apex-dev-darwin-amd64
chmod +x apex-dev-darwin-amd64
sudo ./apex-dev-darwin-amd64  http://<running_controller>:8080/
Login: kitteh1
Pass: floofykittens
```

### Verify
- Once joined, watch convergence over ~60s with using state from the relay (the timer is pretty loose, to see convergence occur to help with debugging).

```
sudo fping -s -g  10.185.0.1 10.185.0.28 -r 1
```

- Now you will see peers for all EC2 nodes since the agent is directly peering to them
and vice versa through the port opened via the connection to the relay/hub node:

```
sudo wg show wg0 dump
```

- If you have a nested hypervisor it will peer directly to nodes on your network
running above that NAT cone (e.g. not inside the hypervisor) since the nested
node will open to the local machine. (I have only tested that with one nested VM though).

- If you have a node that is in a symmetric NAT cone, the agent builds peers to other nodes
local to your host and one connection to the relay node. If it is not local or a relay node,
it will not build a peer (e.g. EC2 nodes are reachable via the relay only since it is symmetric NAT)

- Also works on OSX.

dave-tucker (approved):
LGTM

Issue #169, opened 2022-11-08
Support forwarding without re-encrypting traffic


russellb:
The current design for supporting hub-and-spoke connectivity assumes that the hub is a node participating in the zone. It would be incredibly valuable to support an architecture where the service can provide traffic forwarding without needing to re-encrypt the traffic. This would allow the service to provide connectivity without the service needing access to the actual traffic in a zone.

russellb:
implemented in #1778

PR #209, opened 2022-11-30
README.md: Add Guiding Principles

russellb:
This change introduces a set of guiding principles for the project. These principles should be things that the whole team agrees on and feels passionately about. These principles should be used to guide decision making as the project continues to evolve.

There are a few other minor cleanups to other text included, as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,17 +1,15 @@
 # ![Apex](assets/logo.svg)
 
 > **Warning**: This project is currently a PoC. It's an experiment to try out
-> some new ideas. Do not rely on this as it could be abandoned at any time.
+> some new ideas. Do not rely on this, as it could be abandoned at any time.
 
 [![build](https://github.com/redhat-et/apex/actions/workflows/build.yml/badge.svg)](https://github.com/redhat-et/apex/actions/workflows/build.yml)
-
-> *Roads? Where we're going, we don't need roads - Dr. Emmett Brown*
 
 This project demonstrates an approach for building an IP connectivity-as-a-service solution that provides isolated zone-based connectivity using Wireguard for tunneling.
 
 ## Project Vision
 
-This project aims to provide connectivity between nodes deployed across heterogeneous environments (Edge, Public, Private, and Hybrid Cloud) with different visibilities (nodes in a Cloud VPC, nodes behind NAT, etc.). This solution is not specific to any infrastructure or application platform but focuses on providing IP connectivity between nodes and the container or VM workloads running on those nodes. This service is complementary to platforms-specific networking, as it can be used to expand connectivity to places that the platform could not reach otherwise.
+This project aims to provide connectivity between nodes deployed across heterogeneous environments (Edge, Public, Private, and Hybrid Cloud) with different visibilities (nodes in a Cloud VPC, nodes behind NAT, etc.). This solution is not specific to any infrastructure or application platform but focuses on providing IP connectivity between nodes and the container or VM workloads running on those nodes. This service is complementary to platforms-specific networking, as it can expand connectivity to places the platform could not reach otherwise.
 
 Some of the features and use cases that this project aims to support are:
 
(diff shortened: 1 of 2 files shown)


russellb:
Before merging this, I would prefer a review from @nerdalert, @dave-tucker, and @vishnoianil.

README.md
```diff
@@ -11,17 +11,25 @@ This project demonstrates an approach for building an IP connectivity-as-a-servi
...
+2. **Simple UX Above Features** -- Networking technology is often made incredibly complex. One key value that the Apex service can provide is a simplified user experience for a much more challenging problem. When deciding how or if to add a feature, we value retaining the simplicity of the user experience over functionality.
+3. **Optimized Connectivity** -- Forcing traffic through a central hub or other intermediary does not provide ideal network connectivity. Instead, we will implement various techniques to provide direct, mesh connectivity wherever possible, even in places where NAT or firewalls would typically prevent it.
+4. **Platform Agnostic** - Apex will work independently from the infrastructure platform and support multiple operating systems (Linux, Mac, Windows). We value making the service easy to use with different infrastructure platforms (Kubernetes, for example), but we will avoid changes that tie the service to any particular platform.
```
vishnoianil, line 32: @russellb looks good to me. Thanks for putting this together.
"Zero Trust Security" should also be a core principle ? Zero Touch Device Provisioning/Onboarding (Seamless onboarding) probably the other one, but not sure.

vishnoianil (approved):

README.md
nerdalert: +1 this is great, ty! I agree with Anil on a security bullet if it doesn't overload it. I could see it being generic sec bullet with mentions of ZTN and even SASE. Could potentially wedge in observability in that bullet as well.

nerdalert (approved):

README.md
russellb: Good point about security. I think that one is worth adding.  How about this:

---

5. **Secure Traffic** -- All traffic must be encrypted and the service should never have the ability to decrypt it. Even in the case that traffic needs to be relayed to provide connectivity, the relay must not be able to decrypt the traffic it is relaying[^1].

[^1]: See [Issue #169](https://github.com/redhat-et/apex/issues/169). Our first forwarding implementation is by using a normal apex node on the network which decrypts, makes an IP routing decision, and sends it back out over another encrypted tunnel.

russellb: I wanted to capture the aspirational goal here, but also acknowledge that we haven't fully achieved it yet.

README.md
```diff
@@ -1,7 +1,7 @@
 # ![Apex](assets/logo.svg)
 
 > **Warning**: This project is currently a PoC. It's an experiment to try out
-> some new ideas. Do not rely on this as it could be abandoned at any time.
+> some new ideas. Do not rely on this, as it could be abandoned at any time.
 
 [![build](https://github.com/redhat-et/apex/actions/workflows/build.yml/badge.svg)](https://github.com/redhat-et/apex/actions/workflows/build.yml)
```
dave-tucker, line 7: While you're here, removing the quote on line 8 would be appreciated since it only applied when we had a different name 😆

README.md
dave-tucker: I would go further and add a bullet that says something like "Secure By Design" or somesuch...

This is my take, but it's too many bullets - probably should live in a `docs/security.md` or somesuch.

5. Zero Trust Architecture -- The service has been designed with the tenets of Zero Trust Networking in mind.
- Continuous monitoring and validation: Devices check in to the service regularly for peer updates. User logins are managed via OIDC which allows the use of short-lived access tokens that can be periodically refreshed. In effect, both Devices and Users check in regularly and their Peers or Tokens can be quickly revoked if needed.
- Device Authorization: Our device on-boarding process associates devices with authenticated users.
- Least Privilege: Users are assigned only the privileges they require (add footnote about oauth scopes)
- Microsegmentation: Users are broken into smaller segments using Zones and may only communicate within the same zone

The points about encryption and traffic security are good ones! I also feel like that needs to be in a dedicated security doc.

.... so maybe that's a long way of saying the bullet should just be that Security is a guiding principle.

README.md
russellb: done!

README.md
russellb: Thanks, everyone.

I added a generic security bullet that refers to a dedicated security document. The security document includes my bullet on encrypted traffic, as well as all of the ones proposed by @dave-tucker.

docs/security.md
```diff
@@ -0,0 +1,13 @@
...
+2. **Zero Trust Architecture** -- The service has been designed with the tenets of Zero Trust Networking in mind.
+3. **Continuous monitoring and validation** -- Devices check in to the service regularly for peer updates. User logins are managed via OIDC which allows the use of short-lived access tokens that can be periodically refreshed. In effect, both Devices and Users check in regularly and their Peers or Tokens can be quickly revoked if needed.
+4. **Device Authorization** -- Our device on-boarding process associates devices with authenticated users.
+5. **Least Privilege** -- Users are assigned only the privileges they require[^2].
+6. **Microsegmentation** -- Users are broken into smaller segments using Zones and may only communicate within the same zone.
+
+[^1]: See [Issue #169](https://github.com/redhat-et/apex/issues/169). Our first forwarding implementation is by using a normal apex node on the network which decrypts, makes an IP routing decision, and sends it back out over another encrypted tunnel. This was based on convenience, but we know an alternative is required.
+[^2]: ... oauth scopes ...
```
russellb, line 13: @dave-tucker you had "add footnote about oauth scopes," but this is all I have so far :-D

Issue #216, opened 2022-12-01
HTTPS all of the things!


dave-tucker:
TLS termination should happen on the ingress controller - we're assuming that HTTP (or gRPC) between services in K8s is secure enough.
We should use a self-signed certificate for local development.
Operate-first should probably use a nicer domain and a real certificate.
K8s cert-manager should be able to help with this.

Issue #212, opened 2022-12-01
kustomize: Add operate-first overlay


dave-tucker:
Add an overlay for operate-first (copy the dev one and change hostname overrides)

This should include a `production` dex config.
Related to https://github.com/operate-first/support/issues/1154

Since we'd require the use of `externalsecrets` to securely store credentials used for GitHub authentication.

russellb:
@dave-tucker I think this is complete. Is there anything else you want to track with this?

PR #217, opened 2022-12-03
HTTPS ALL THE THINGS!

dave-tucker:
Adds HTTPS config to the dev overlay via a Kustomize Component. Start adding the scaffold for an operate-first overlay - we'll need to figure out how to use cert-manger there.
Updates dev README on how to add the self-signed CA to the root store. This will push us to update the deployment on K8s vs. up/down of cluster to avoid having to re-trust the cert, but that seems a fair tradeoff.

Minikube will break as it will need to patch the ingress controller to add the necessary flag for SSL passthrough.

In staging/production this shouldn't be needed since the auth server will be reachable via HTTPS and have a valid certificate.

Fixes: #216 
Updates: #212 


nerdalert (approved):
LGTM! one doc fix, we can get with the other fixes in there later (join updates).

PR #222, opened 2022-12-05
Makefile: remove apiserver from all target

russellb:
A recent change dropped the apiserver targets from the Makefile. The
    'all' target still had an 'apiserver' dependency, but no 'apiserver'
    target existed. We exclusively run the apiserver in a container, so we
    don't need to build it directly here.
    
    Re-add APISERVER_DEPS, as it's used as a dependency for the go-lint
    target, and that still makes sense.
    
### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 .PHONY: all
-all: go-lint apex apiserver
+all: go-lint apex
 
 .PHONY: apex
 apex: dist/apex dist/apex-linux-arm dist/apex-linux-amd64 dist/apex-darwin-amd64 dist/apex-darwin-arm64 dist/apex-windows-amd64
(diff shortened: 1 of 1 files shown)


russellb:
> We should remove apiserver from `all` since its now exclusively run from inside the container

@dave-tucker updated the patch to remove the target from `all`

Makefile
```diff
@@ -8,6 +8,8 @@ COMMON_DEPS=$(wildcard ./internal/**/*.go) go.sum go.mod
 
 APEX_DEPS=$(COMMON_DEPS) $(wildcard cmd/apex/*.go)
 
+APISERVER_DEPS=$(COMMON_DEPS) $(wildcard cmd/apiserver/*.go)
```
dave-tucker, line 11: I don't think this is required either now.

dave-tucker (changes requested):

Makefile
russellb: It’s used as dependencies for the go-lint target, so that’s why I added it back. On second thought though I don’t think the dependencies are useful there since go-lint is a PHONY target.

dave-tucker (approved):

PR #232, opened 2022-12-07
selfsigned-ca: Don't replace full volumeMount list

russellb:
The patch for the apiserver StatefulSet replaced the entire list of volumeMounts instead of adding a volumeMount to the list. I noticed this after adding a second volumeMount in another branch.

### DIFF
--- a/deploy/components/selfsigned-ca/kustomization.yaml
+++ b/deploy/components/selfsigned-ca/kustomization.yaml
@@ -64,8 +64,8 @@ patchesJson6902:
       Name: apiserver
     patch: |-
       - op: add
-        path: /spec/template/spec/containers/1/volumeMounts
-        value: [{"name":"trust","mountPath":"/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem","readOnly":true,"subPath":"tls-ca-bundle.pem"}]
+        path: /spec/template/spec/containers/1/volumeMounts/-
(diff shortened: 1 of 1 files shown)


deploy/components/selfsigned-ca/kustomization.yaml
```diff
@@ -64,8 +64,8 @@ patchesJson6902:
       Name: apiserver
     patch: |-
       - op: add
-        path: /spec/template/spec/containers/1/volumeMounts
-        value: [{"name":"trust","mountPath":"/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem","readOnly":true,"subPath":"tls-ca-bundle.pem"}]
+        path: /spec/template/spec/containers/1/volumeMounts/-
+        value: {"name":"trust","mountPath":"/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem","readOnly":true,"subPath":"tls-ca-bundle.pem"}
```
russellb, line 68: There are similar changes needed in a couple other places. I should have looked for those before posting!

russellb: and actually this change only worked on my branch where volumeMounts already exists. Using only this change will break.

russellb:
This needs to be included in my branch and not submitted separately. You can only change to append like this when the item already exists. The current patch is right for creating it when it does not exist. It just also overwrites it if it exists. We'll just have to change these over time depending on how the base resources look.

Issue #231, opened 2022-12-07
CA trust is broken for the apiserver


russellb:
Presumably after #216, trust of other CAs other than our own self-signed one is broken. My feature-flags branch has the API server pulling feature flag config from an https URL, which now results in:

```
$ kubectl logs -n apex apiserver-0 -c apiserver
2022/12/07 16:36:42 error: impossible to retrieve flags from the config file: Get "https://raw.githubusercontent.com/russellb/apex/feature-flags/deploy/default_flags.yaml": x509: certificate signed by unknown authority
```

Issue #214, opened 2022-12-01
Put Zones behind a feature flag


russellb:
Depends on https://github.com/redhat-et/apex/issues/213

Zones support is incomplete so it should be turned off by default and put behind a feature flag. For our first cut at the running service, we should have a single, static zone.

The zone concept is still important so it should be visible in the UI. Zone creation from the UI and API should be disabled.

Issue #213, opened 2022-12-01
Add feature flags!


dave-tucker:
We should be able to enable and disable features for deployments

PR #229, opened 2022-12-07
Instrument all the things!

dave-tucker:
Adds:

- Open Telemetry tracing for apiserver (including http and database)
- Prometheus Metrics for apiserver - just the basics for now

Fixes: #172 #173 #174

Signed-off-by: Dave Tucker <dave@dtucker.co.uk>

PR #234, opened 2022-12-09
[WIP] Feature flags

russellb:
for issues #213 and #214

TODO:
- [x] Add Ingress for the feature flag proxy
- [ ] Update UI to disable multi-zone support based on feature flag status
- [ ] Move the apiserver back to polling an https URL for feature flags after issue #231 is resolved
- [ ] Document feature flags properly in docs/feature-flags.md


deploy/base/goff-proxy/kustomization.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+generatorOptions:
```
dave-tucker, line 3: I've just been through removing this in #229 - disableNameSuffixHash - makes the configmap immutable.
That means we'd have to delete/re-create it to pick up changes which I don't think is the behaviour we want.

internal/handlers/zone.go
```diff
@@ -33,6 +34,17 @@ const (
...
 	ctx := c.Request.Context()
+
+	user := ffuser.NewUser(c.GetString(gin.AuthUserKey))
+	multiZoneEnabled, err := ffclient.BoolVariation("multi-zone", user, false)
+	if err != nil {
+		c.JSON(http.StatusInternalServerError, models.ApiError{Error: err.Error()})
+		return
+	}
+	if !multiZoneEnabled {
+		c.JSON(http.StatusForbidden, models.ApiError{Error: "multi-zone support is disabled"})
```
dave-tucker, line 45: http.StatusMethodNotAllowed would be more accurate

deploy/base/goff-proxy/kustomization.yaml
russellb: ah, ok, yeah I'll remove it. It's just from copy/paste.

russellb:
I'll squash all of these random commits once it's no longer a WIP.

russellb:
I'm going to close this WIP PR in favor of introducing feature flags in smaller stages through multiple PRs.

PR #251, opened 2022-12-16
vscode: Set schema for deploy.yaml

dave-tucker:


dave-tucker:
Another trivial one... to trigger another deploy in ArgoCD since we don't have powers to do that manually yet.

PR #261, opened 2022-12-18
Add back windows support

nerdalert:
- the windows wireguard utility requires a config file to bring up the wg interface. It is only there to bind the private key to the wg0 interface. This also used it to add ip to the iface. Peers are all managed with the same library across all platforms now.
- broke out interface operations into a seperate file.

### DIFF
--- a/internal/apex/apex.go
+++ b/internal/apex/apex.go
@@ -26,7 +26,7 @@ const (
 	darwinIface       = "utun8"
 	WgDefaultPort     = 51820
 	wgIface           = "wg0"
-	WgWindowsConfPath = "C:/"
+	WgWindowsConfPath = "C:/apex/"
 )
 
 type Apex struct {
@@ -403,7 +403,6 @@ func checkOS(logger *zap.SugaredLogger) error {
 			return fmt.Errorf("unable to create the wireguard config directory [%s]: %v", WgDarwinConfPath, err)
 		}
 	case Windows.String():
-		logger.Fatal("Windows is temporarily unsupported")
 		logger.Debugf("[%s] operating system detected", nodeOS)
 		// ensure the windows wireguard directory exists
 		if err := CreateDirectory(WgWindowsConfPath); err != nil {
@@ -527,19 +526,21 @@ func (ax *Apex) findLocalEndpointIp() (string, error) {
 
 // binaryChecks validate the required binaries are available
 func binaryChecks() error {
+	// Windows userspace binary
 	if GetOS() == Windows.String() {
 		if !IsCommandAvailable(wgWinBinary) {
 			return fmt.Errorf("%s command not found, is wireguard installed?", wgWinBinary)
 		}
-	} else {
-		if !IsCommandAvailable(wgBinary) {
-			return fmt.Errorf("%s command not found, is wireguard installed?", wgBinary)
-		}
 	}
+	// Darwin wireguard-go userspace binary
 	if GetOS() == Darwin.String() {
 		if !IsCommandAvailable(wgGoBinary) {
 			return fmt.Errorf("%s command not found, is wireguard installed?", wgGoBinary)
 		}
+	}
+	// all OSs require the wg binary
+	if !IsCommandAvailable(wgBinary) {
+		return fmt.Errorf("%s command not found, is wireguard installed?", wgBinary)
 	}
 	return nil
 }
--- a/internal/apex/utils_darwin.go
+++ b/internal/apex/utils_darwin.go
@@ -3,6 +3,7 @@
 package apex
 
 import (
+	"fmt"
 	"net"
 
 	"github.com/vishvananda/netlink"
@@ -14,8 +15,13 @@ func RouteExists(s string) (bool, error) {
 	return false, nil
 }
 
-// AddRoute currently only used for darwin build purposes
+// AddRoute adds a route to the specified interface
 func AddRoute(prefix, dev string) error {
+	_, err := RunCommand("route", "-q", "-n", "add", "-inet", prefix, "-interface", dev)
+	if err != nil {
+		return fmt.Errorf("route add failed: %v", err)
+	}
+
 	return nil
 }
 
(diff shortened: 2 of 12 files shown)


internal/apex/apex.go
```diff
@@ -527,19 +526,21 @@ func (ax *Apex) findLocalEndpointIp() (string, error) {
...
-			return fmt.Errorf("%s command not found, is wireguard installed?", wgBinary)
-		}
 	}
+	// Darwin wireguard-go userspace binary
 	if GetOS() == Darwin.String() {
 		if !IsCommandAvailable(wgGoBinary) {
 			return fmt.Errorf("%s command not found, is wireguard installed?", wgGoBinary)
 		}
 	}
+	// all OSs require the eg binary
```
russellb, line 541:
```suggestion
	// all OSs require the wg binary
```

internal/apex/utils_darwin.go
```diff
@@ -16,6 +17,11 @@ func RouteExists(s string) (bool, error) {
 
 // AddRoute currently only used for darwin build purposes
```
russellb, line 18:
```suggestion
// AddRoute adds a route to the specified interface
```

internal/apex/wg-iface.go
```diff
@@ -0,0 +1,195 @@
...
+						return iface.Name, nil
+					}
+				} else {
+					continue
+				}
+			}
+		} else {
+			continue
+		}
+	}
```
russellb, line 32: This is just a style choice, but I find this nesting to be a bit deep. We can reduce it a bit by flipping the `err` checks like this. It also lets you drop the `else` clauses.

```suggestion
	for _, iface := range interfaces {
		if addrs, err := iface.Addrs(); err != nil {
			continue
		}
		for _, addr := range addrs {
			if ifaceIP, _, err := net.ParseCIDR(addr.String()); err != nil {
				continue
			}
			if ifaceIP.Equal(ip) {
				return iface.Name, nil
			}
		}
	}
```

internal/apex/wg-iface.go
```diff
@@ -0,0 +1,195 @@
...
+
+	var wgOut string
+	var err error
+	if ifaceExists(logger, wgIface) {
+		wgOut, err = RunCommand("wireguard.exe", "/uninstalltunnelservice", wgIface)
+		if err != nil {
+			logger.Debugf("failed to down the wireguard interface (this is generally ok): %v\n", err)
+		}
+		time.Sleep(time.Second * 1)
```
russellb, line 127: I was wondering why the sleep was here, but then I see the comment a few lines down.  It looks like it should be moved up?

The comment also makes it sound like there's probably a better way to do this than sleep. You could just wait until there's no longer a wireguard interface present by calling ifaceExists() again?

internal/apex/wg-iface.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+	logger.Debugf("stopped windows tunnel svc:%v\n", wgOut)
+	// sleep for one second to give the wg async exe time to tear down any existing wg0 configuration
+	wgOut, err = RunCommand("wireguard.exe", "/installtunnelservice", windowsWgConfigFile)
+	if err != nil {
+		return fmt.Errorf("failed to start the wireguard interface: %v\n", err)
+	}
+	logger.Debugf("started windows tunnel svc: %v\n", wgOut)
+	time.Sleep(time.Second * 1)
```
russellb, line 136: Why is the final sleep necessary?

internal/apex/wg-iface.go
```diff
@@ -0,0 +1,195 @@
...
+					switch ip := addr.(type) {
+					case *net.IPNet:
+						if ip.IP.DefaultMask() != nil {
+							return ip.IP
+						}
+					}
+				}
+			}
+		}
+	}
```
russellb, line 183: Similar to an earlier comment, my style preference would be to cut down on the nesting depth like this, but up to you if you want to take it or not.

```suggestion
for _, inter := range interfaces {
	if inter.Name != ifname {
		continue
	}
	if addrs, err := inter.Addrs(); err != nil {
		continue
	}
	for _, addr := range addrs {
		switch ip := addr.(type) {
		case *net.IPNet:
			if ip.IP.DefaultMask() != nil {
				return ip.IP
			}
		}
	}
}
```

internal/apex/wg-iface.go
nerdalert: I ran into a bit of delay in windows setting up the interface < 1s. Ran into an issue where it started trying to add peers before the iface was ready. Same scenario as above.

internal/apex/wg-iface.go
nerdalert: Added iface checks to both before a sleep. Will be curious if anyone else runs into this on Windows. The only machine I have is the one running in EC2. I added a debug statement in case anyone else is ever interested. 

Ty for the review!

internal/apex/wg-iface.go
```diff
@@ -0,0 +1,212 @@
...
+	wgOut, err = RunCommand("wireguard.exe", "/installtunnelservice", windowsWgConfigFile)
+	if err != nil {
+		return fmt.Errorf("failed to start the wireguard interface: %v\n", err)
+	}
+	logger.Debugf("started windows tunnel svc: %v\n", wgOut)
+	// ensure the link is created before adding peers
+	if !ifaceExists(logger, wgIface) {
+		logger.Debugf("windows iface %s has not been created yet, pausing for 1 second", wgIface)
+		time.Sleep(time.Second * 1)
+	}
```
russellb, line 145: It would be a bit more robust to turn this into a loop, but eventually, give up and return an error.  You could use https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait#ExponentialBackoff

I would do the same thing in the check above

russellb: From private chat, Brent would like to merge this as-is. We should come back and update this to make it more robust whenever we expect it to be run in prod.

PR #265, opened 2022-12-21
Control multiple zone support with a feature flag

### LINKED ISSUE #213: Add feature flags!

### LINKED ISSUE #214: Put Zones behind a feature flag

russellb:
This PR introduces the first phase of feature flags and uses a
feature flag to disable support for multiple zones (multi-zone).

The feature-flag backend is represented by the FFlags API, which, for 
now is just a simple backend with defaults that can be changed with
an environment variable. Eventually, we can replace this with one of
several other options that will allow us to do more advanced feature
controls.

Feature flags are exposed via our own API. They may be listed, or a
specific flag may be queried.

The UI consumes the API to know whether multiple zones are supported.
If they are not supported, the "Create Zone" button is disabled (or 
greyed out).

The Zone create/delete APIs also check the feature flags in the API 
server. These methods will be denied if the flag has multi-zone
support off.  Unit tests exercise the zone create/delete methods with
support on and off.

The dev deployment runs with multi-zone support on by default, but the 
prod deployment has it off.

Closes #213
Closes #214

### DIFF
--- a/deploy/apex/overlays/dev/kustomization.yaml
+++ b/deploy/apex/overlays/dev/kustomization.yaml
@@ -13,6 +13,10 @@ configMapGenerator:
   files:
   - files/config.yaml
   name: dex
+- behavior: merge
+  literals:
+   - APEX_FFLAG_MULTI_ZONE=true
+  name: apiserver
 patchesJson6902:
 - target:
     kind: Ingress
--- a/hack/kind/kind.sh
+++ b/hack/kind/kind.sh
@@ -48,6 +48,12 @@ down() {
     kind delete cluster --name apex-dev
 }
 
+cacert() {
+    mkdir -p .certs
+    kubectl get secret -n apex apex-ca-key-pair -o json | jq -r '.data."ca.crt"' | base64 -d > .certs/rootCA.pem
+    CAROOT=$(pwd)/.certs mkcert -install
+}
+
 case $1 in
     "up")
         up
@@ -55,8 +61,11 @@ case $1 in
     "down")
         down
         ;;
+    "cacert")
+        cacert
+        ;;
     *)
-        echo "command required. up or down"
+        echo "command required. up, down, or cacert"
         exit 1
         ;;
 esac
--- a/internal/handlers/fflags.go
+++ b/internal/handlers/fflags.go
@@ -1,0 +1,48 @@
+package handlers
+
+import (
+	"net/http"
+
+	"github.com/gin-gonic/gin"
+	"github.com/redhat-et/apex/internal/models"
+)
+
+// ListFeatureFlags lists all feature flags
+// @Summary      List Feature Flags
+// @Description  Lists all feature flags
+// @Tags         FFlag
+// @Accepts		 json
+// @Produce      json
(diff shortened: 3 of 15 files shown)


deploy/apex/overlays/dev/kustomization.yaml
```diff
@@ -109,6 +109,13 @@ patchesJson6902:
     - op: add
       path: /spec/template/spec/volumes
       value: [{"name":"trust","secret":{"secretName":"apex-ca-key-pair","optional":false,"items":[{"key":"ca.crt","path":"tls-ca-bundle.pem"}]}}]
+- target:
```
dave-tucker, line 112: Could you add this to the configmap at `deploy/apex/base/apiserver/kustomization.yaml` instead with value of `false`... The deployment.yaml would set that environment variable to the valueFrom configmap... and in this kustomization.yaml you simply override the configmap value like we do for hostnames already...

Yeah it's a little obscure, but there is method in the madness...
When I'm back from PTO, we'll be adding a new `configMapGenerator` to these `overlays` which has a `strategy: merge`... that will merge values in our configMap with the one in base... and we won't need to use patches to set these values 😄

hack/kind/kind.sh
```diff
@@ -48,15 +48,24 @@ down() {
     kind delete cluster --name apex-dev
 }
 
+cacert() {
```
dave-tucker, line 51: YAS! Thank you!
Adding a note about this to the README would be handy

internal/handlers/fflags.go
```diff
@@ -0,0 +1,48 @@
...
+// @Summary      Get Feature Flag
+// @Description  Gets a Feature Flag by name
+// @Tags         FFlag
+// @Accepts		 json
+// @Produce      json
+// @Param		 name path      string feature flag name
+// @Success      200  {object}
+// @Failure      400  {object}  models.ApiError
+// @Failure      404  {object}  models.ApiError
+// @Router       /fflags/{id} [get]
```
dave-tucker, line 33: nit: Param is `{name}`

dave-tucker (changes requested):
This look awesome! And has tests!
Couple of minor comments then this is good to go.

deploy/apex/overlays/dev/kustomization.yaml
russellb: Sounds good, I'll move it

hack/kind/kind.sh
russellb: good call

deploy/apex/overlays/dev/kustomization.yaml
russellb: I changed it to use the configmap merge strategy

dave-tucker (approved):

PR #266, opened 2022-12-21
apiserver: Fix stashing of usernames

### LINKED ISSUE #263: Usernames no longer visible in the UI
We previously stashed usernames in our database so we could show them in our UI. That's no longer working. The username field is currently blank for all users.  This is the code that should be grabbing the username:

https://github.com/redhat-et/apex/blob/main/internal/routers/middleware.go#L66

It appears that `claims.UserName` isn't set anymore.

russellb:
The username field is no longer showing up in the claims. That field is blank, but the FullName field has the username in it. Adjust the code to look at both, preferring username first if set.

Fixes #263


dave-tucker (approved):
LGTM

PR #268, opened 2022-12-22
ci: Speed up

dave-tucker:
1. Build docker images for apiserver, frontend and in a seperate stage
2. Use the github actions cache to store image layers, hopefully speeding up image builds in future
3. Have the e2e job grab images from artifacts instead of building them and cache the `test` image layers
4. Have the e2e job call go test ./... directly instead of via the makefile in order to avoid calling other make targets

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,44 +2,97 @@ name: build
 
 on:
   push:
-    branches: [ "main" ]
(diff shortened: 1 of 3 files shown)


dave-tucker:
That's 19mins vs.... 19 mins... 😆 
But once that caching kicks in I think this will start moving faster.

PR #272, opened 2022-12-22
ui: Implement deletes

### LINKED ISSUE #155: ui: Implement Deletes
Once we have backend support in #153 we will need support in the UI too.

russellb:
This PR adds delete support for Users, Devices, Zones, and Peers. It depends on PR #265 because the Zone delete button is disabled if the feature is turned off by its feature flag.

Closes #155

### DIFF
--- a/ui/src/pages/Devices.tsx
+++ b/ui/src/pages/Devices.tsx
@@ -1,4 +1,4 @@
-
+import { Fragment } from "react";
 import {
     Datagrid,
     List,
@@ -10,7 +10,9 @@ import {
     ReferenceField,
     useRecordContext,
     useGetOne,
-    Loading
+    Loading,
+    BulkExportButton,
+    BulkDeleteButton,
 } from 'react-admin';
 
 const ZoneNameFromPeer = () => {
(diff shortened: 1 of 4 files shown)


nerdalert (approved):
Gave it a spin with #280 on the agent side and deletes are plumbed from the UX down to the driver and work.

At some point, would be nice to get error msgs back to the UX. For example, if you try and delete a zone that has peers, it shows it succeeds at first but then says error and flips back to the zone. Im for merging this and then enhancing rather than slowing down. 

We should have a UX session to walk through a user experience soon and think about the obvious scenarios and how we want to express them to the user. In the zone example, we could pre-check and block even attempting the transaction, we could let them try and throw back an error, we could add a force option, etc, etc. Same applies to the user table once we get sso plumbed through. Also how we want to express DB column CRUD, child-prefix or policy for example.

Ty!

russellb:
Thanks for the review, Brent! I'd rather track those follow-ups as a new issue.

I also just rebased this on `main` to resolve a couple of minor conflicts.

russellb:
the last push is just another rebase.

PR #275, opened 2022-12-22
Add check for dependencies to ensure required dependencies are installed.

vishnoianil:

### DIFF
--- a/hack/kind/kind.sh
+++ b/hack/kind/kind.sh
@@ -1,18 +1,48 @@
 #!/bin/sh
 
-set -e
+set -ex
+info_message() {
+    if [ -z "${1}" ]; then
+        echo "info_message() requires a message"
+        exit 1
+    fi
+    echo -e "[\033[0;34m ACTION \033[0m] ${1}"
+}
+
+pass_message() {
+    if [ -z "${1}" ]; then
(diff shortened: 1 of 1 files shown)


hack/kind/kind.sh
```diff
@@ -1,18 +1,48 @@
...
+    if [ -n "$1" ]; then
+        echo -e "[\033[0;31m FAILED \033[0m] ] ${1}"
+    fi
+}
 
 up() {
+    if ! [ -x "$(command -v kubectl)" ]; then
+        error_message "kubectl is required. Please install kubectl"
+        exit 1
+    fi
```
russellb, line 34: Might as well check for `kind` while you're at it

vishnoianil: I added that initially, but removed it, because kind itself does that. It throws `kind not found`.

vishnoianil: My above comment is hilarious 🗡️ , `Kind` throws `'kind not found` :). I meant it's the first command for up and down both and shell throws kind not found if it doesn't exist, so i thought it's redundant. But no harm in  explicit check.

vishnoianil: @russellb added explicit check for kind.

PR #280, opened 2022-12-23
Add peer deletes to the agent

nerdalert:
- When a peer or device is deleted from the database, the cached peers are diffed against the latest peer listing. If the cache does not contain the peer, it is removed from all nodes in the mesh. Both the peer and routes are removed from the endpoints.

- This does not do anything on the device being deleted. That is still a tbd on how we handle that. The deleted node has no connectivity to the mesh/peers since it is removed from all of the tunnel endpoints unless it is added back to the mesh.

### DIFF
--- a/internal/apex/apex.go
+++ b/internal/apex/apex.go
@@ -331,6 +331,7 @@ func (ax *Apex) Reconcile(zoneID uuid.UUID, firstTime bool) error {
 			newPeers = append(newPeers, p)
 		}
 	}
+
 	if changed {
 		ax.logger.Debugf("Peers listing has changed, recalculating configuration")
 		ax.buildPeersConfig()
@@ -338,6 +339,12 @@ func (ax *Apex) Reconcile(zoneID uuid.UUID, firstTime bool) error {
 			return err
 		}
 	}
+
+	// check for any peer deletions
+	if err := ax.handlePeerDelete(peerListing); err != nil {
+		ax.logger.Error(err)
+	}
+
 	return nil
 }
 
--- a/internal/apex/route.go
+++ b/internal/apex/route.go
@@ -2,6 +2,8 @@ package apex
 
 import (
 	"net"
+
+	"github.com/redhat-et/apex/internal/models"
 )
 
 // handlePeerRoute when a new configuration is deployed, delete/add the peer allowedIPs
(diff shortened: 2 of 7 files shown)


internal/apex/route.go
```diff
@@ -62,3 +64,37 @@ func (ax *Apex) addChildPrefixRoute(childPrefix string) {
...
+			if err := DeleteRoute(allowedIP, dev); err != nil {
+				ax.logger.Debug(err)
+			}
+		}
+
+	case Linux.String():
+		for _, allowedIP := range wgPeerConfig.AllowedIPs {
+			routeExists, err := RouteExists(allowedIP)
+			if err != nil {
+				ax.logger.Info(err)
```
russellb, line 83: I think you should add a `continue` here. You don't want to act on the result if an error occurs.

Do you really want `Info` here? Check the other logging in this function. It's a mix of Info and Debug, but all for error conditions.

internal/apex/utils_linux.go
```diff
@@ -162,3 +162,26 @@ func delLink(ifaceName string) error {
...
+	routeSpec := netlink.Route{
+		Dst:       ipNet,
+		LinkIndex: link.Attrs().Index,
+	}
+
+	if err := netlink.RouteDel(&routeSpec); err != nil {
+		return err
+	}
+
+	return nil
```
russellb, line 186: You can collapse these 5 lines into:

```go
    return netlink.RouteDel(&routeSpec)
```

internal/apex/wg.go
```diff
@@ -99,6 +101,66 @@ func (ax *Apex) addPeer(wgPeerConfig wgPeerConfig) error {
 	return wgClient.ConfigureDevice(ax.tunnelIface, cfg)
 }
 
+func (ax *Apex) handlePeerDelete(peerListing []models.Peer) error {
+	// if the canonical peer listing does not contain a peer from cache, delete the peer
+	for _, p := range ax.peerCache {
+		if !inPeerListing(peerListing, p) {
```
russellb, line 107: I suggest:

```go
    if inPeerListing(peerListing, p) {
        continue
    }
    ...
```

That will cut down on some nesting in this loop.

russellb (commented):
In your commit message: s/container/contain/

internal/apex/wg.go
```diff
@@ -99,6 +101,66 @@ func (ax *Apex) addPeer(wgPeerConfig wgPeerConfig) error {
...
+	return nil
+}
+
+func inPeerListing(peers []models.Peer, p models.Peer) bool {
+	for _, peer := range peers {
+		if peer.ID == p.ID {
+			return true
+		}
+	}
+	return false
```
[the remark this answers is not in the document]
nerdalert: For sure! An unordered list makes much more sense.

internal/apex/apex.go
```diff
@@ -331,13 +331,23 @@ func (ax *Apex) Reconcile(zoneID uuid.UUID, firstTime bool) error {
...
+
 	if changed {
 		ax.logger.Debugf("Peers listing has changed, recalculating configuration")
 		ax.buildPeersConfig()
 		if err := ax.DeployWireguardConfig(newPeers, false); err != nil {
 			return err
 		}
 	}
+
+	if len(ax.peerCache) > len(peerListing) {
```
russellb, line 343: I think it's possible to miss a delete here. IIRC, this is currently based on polling, right? A peer could have been deleted no matter what the lengths of these lists are compared to each other. Imagine 1 deleted and 1 added, or 1 deleted and 2 added.

The logic you want is:  does the peer cache have any peers that are not in the peer listing?

The quickest way to fix this would be to *always* call `handlePeerDelete()`, since it looks harmless to call even if nothing was deleted. We'll need to optimize the data structures and how the check is done later on. Right now, it looks like O(n^2). but that's not very important right now.

russellb (changes requested):

dave-tucker (approved):
LGTM

PR #281, opened 2023-01-03
Remove debug flag from kind.sh

vishnoianil:


vishnoianil:
Merging it. Minor change.

PR #286, opened 2023-01-04
Adds QA e2e in ec2 for operate first QA deployment

nerdalert:
- Adds vars to repo secrets.
- Ansible vars are encrypted and stored in GH secrets. These vars contain the deployment details (node count, VPCs, etc)
- Once nodes are deployed, connectivity is verified from the relay node to all of the spoke nodes. Output is saved and displayed in the action output. This will be tweeked and further tests will be added once this is up and functional via webhooks from operate first.

### DIFF
--- a/.github/workflows/qa-ec2.yml
+++ b/.github/workflows/qa-ec2.yml
@@ -1,0 +1,76 @@
+name: qa-ec2-e2e
+
+on: workflow_dispatch
+
+jobs:
+  deploy-ec2:
+    name: deploy-ec2-e2e
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    strategy:
+      fail-fast: false
+    env:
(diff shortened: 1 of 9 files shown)


nerdalert:
Need to add the webhooks from operate first QA deployment and replace the triggers in `qa-ec2.yml`

.github/workflows/qa-ec2.yml
```diff
@@ -0,0 +1,81 @@
+name: qa-ec2-e2e
+
+on:
+  # Replace with a webhook from the QA deployment in operate first
```
dave-tucker, line 4: You can replace this with `on: workflow_dispatch`.
That will allow us to trigger it manually or via the API etc...

dave-tucker:
Assuming this works and the secrets are set up, then this should be good to go. @nerdalert can you confirm?

PR #290, opened 2023-01-10
ui: Update dependencies

russellb:
Bump babel/core to the latest release, which also pulls in the latest version of the json5 dependency. json5 had a high severity CVE that is fixed in this version.

https://github.com/redhat-et/apex/security/dependabot/10

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
-    "@babel/core": "^7.0.0",
+    "@babel/core": "^7.20.12",
     "@emotion/babel-plugin": "11.10.2",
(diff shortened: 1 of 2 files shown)


russellb:
I don't remember doing a dependency update for a react app before. I wasn't sure if I should add json5 to package.json directly, or get it updated indirectly like this.

russellb:
I got an email about the `json5` issue, which prompted this PR. Now that I look at github's list, there are several more. I should probably just update all dependencies to the latest.

dave-tucker:
@russellb I've set up the dependabot bot to take care of these as that's the easiest approach moving forwards.
Will trawl through the other deps and see if we need to make any manual changes where the bot has failed.

PR #292, opened 2023-01-10
fix: allow kind.sh script work with osx's slightly different sed command

chirino:
Signed-off-by: Hiram Chirino <hiram@hiramchirino.com>


russellb:
lgtm, i'll merge once CI finishes

PR #299, opened 2023-01-13
chore: add a `make use-cockroach` task

chirino:
that changes the kind deployment to use cockroachdb.

The apiserver still needs some work to actually make it work.  But at least this gives developers a way to easily start testing against cockroach.

### DIFF
--- a/deploy/apex/base/apiserver/deployment.yaml
+++ b/deploy/apex/base/apiserver/deployment.yaml
@@ -23,6 +23,11 @@ spec:
                 secretKeyRef:
                   name: database-pguser-apiserver
                   key: host
+            - name: APEX_DB_PORT
+              valueFrom:
+                secretKeyRef:
+                  name: database-pguser-apiserver
+                  key: port
             - name: APEX_DB_NAME
               valueFrom:
                 secretKeyRef:
@@ -39,7 +44,10 @@ spec:
                   name: database-pguser-apiserver
                   key: password
             - name: APEX_DB_SSLMODE
-              value: require
+              valueFrom:
+                configMapKeyRef:
+                  name: apiserver
+                  key: APEX_DB_SSLMODE
             - name: APEX_IPAM_URL
               valueFrom:
                 configMapKeyRef:
--- a/deploy/apex/overlays/cockroach/kustomization.yaml
+++ b/deploy/apex/overlays/cockroach/kustomization.yaml
@@ -1,0 +1,42 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - ../dev
+  - cockroach.yaml
+
+namespace: apex
+
+secretGenerator:
+  - name: database-pguser-ipam
+    literals:
+      - dbname=ipam
+      - host=cockroachdb-public
+      - password=password
+      - port=26257
+      - user=ipam
+
+  - name: database-pguser-apiserver
+    literals:
+      - dbname=apiserver
+      - host=cockroachdb-public
+      - password=password
+      - port=26257
+      - user=apiserver
+
+configMapGenerator:
+  - name: ipam
+    behavior: merge
+    literals:
+      - GOIPAM_PG_SSLMODE=disable
+  - name: apiserver
+    behavior: merge
+    literals:
+      - APEX_DB_SSLMODE=disable
+
+patchesStrategicMerge:
+  - |-
+    $patch: delete
+    apiVersion: postgres-operator.crunchydata.com/v1beta1
+    kind: PostgresCluster
+    metadata:
+      name: database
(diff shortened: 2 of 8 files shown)


deploy/apex/base/apiserver/deployment.yaml
```diff
@@ -40,6 +40,11 @@ spec:
                   key: password
             - name: APEX_DB_SSLMODE
               value: require
+            - name: APEX_DB_PORT
+              valueFrom:
+                secretKeyRef:
+                  name: database-pguser-apiserver
+                  key: port
```
russellb, line 47: This is incredibly picky, but it might be nice to put this next to `APEX_DB_HOST`

hack/kind/kind.sh
```diff
@@ -99,6 +99,39 @@ cacert() {
     CAROOT=$(pwd)/.certs mkcert -install
 }
 
+usecockroach() {
+
```
russellb, line 103: How about a big warning message here to say that this isn't expected to work fully at this point? I just want to avoid anyone trying it and then wondering why their env isn't working.

russellb:
I pushed changes to your branch for the lint job issue and for my trivial review comments.

deploy/apex/overlays/cockroach/secrets.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: v1
```
dave-tucker, line 1: I would use a `secretGenerator` in `kustomization.yaml` for these:
https://kubernetes.io/docs/tasks/configmap-secret/managing-secret-using-kustomize/

deploy/apex/overlays/cockroach/kustomization.yaml
```diff
@@ -0,0 +1,34 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - ../dev
+  - cockroach.yaml
+  - secrets.yaml
+
+namespace: apex
+patchesJson6902:
+  - target:
```
dave-tucker, line 10: If it's possible, I would set that environment variable from a configmap (or secret) and override the configmap value here using:
```yaml
configMapGenerator:
  - name: ipam
     behavior: replace
     literals:
       - enable_ssl=disable
```

This is less likely to fail if we change the ordering or environment variables (i.e someone adds a new one to the top of the list.

dave-tucker (changes requested):

dave-tucker:
Another small nit, but it would be good to add this new overlay to github actions so it gets linted.
Do do that you'll need to run [this script](https://github.com/datreeio/CRDs-catalog#usage) against a kube cluster with the cockroack operator installed and copy the CRD from ~/.datree/crdSchemas to `./deploy/.crdSchemas` so we can validate the  YAML against the CRD

russellb:
@chirino fyi, Dave moved the kind.sh to the Makefile earlier today

chirino:
current status of the e2e tests when running with cockroach db are: 

```
--- FAIL: TestApexIntegrationSuite (111.50s)
    --- FAIL: TestApexIntegrationSuite/TestApexCtl (60.74s)
    --- FAIL: TestApexIntegrationSuite/TestBasicConnectivity (7.25s)
    --- PASS: TestApexIntegrationSuite/TestChildPrefix (8.35s)
    --- FAIL: TestApexIntegrationSuite/TestHubZone (13.43s)
    --- SKIP: TestApexIntegrationSuite/TestRelayNAT (0.00s)
    --- PASS: TestApexIntegrationSuite/TestRequestIPDefaultZone (5.44s)
    --- PASS: TestApexIntegrationSuite/TestRequestIPZone (16.30s)
```

chirino:
> @chirino fyi, Dave moved the kind.sh to the Makefile earlier today

just noticed.. rebased into the makefile.

deploy/apex/overlays/cockroach/kustomization.yaml
chirino: honestly we also need to be able to pass the DB certs like this too.  The cockroach db operator is not doing the automatic cert magic that the crunchy operator is doing and for prod use we will need to configure client and ca certs the the apiserver and ipam pods.

deploy/apex/overlays/cockroach/secrets.yaml
chirino: there are actually no secrets here.  this dev deployment of cockroach accepts ANY password.

russellb:
@chirino looks like this needs another rebase

@dave-tucker I'm assuming you're doing the reviews on this btw

dave-tucker:
>  I'm assuming you're doing the reviews on this btw

I am not. 

But I think to move this forward we should limit the changes in this PR to only the Makefile + adding an "overlay" for cockroachdb.
We can then properly review any code-changes required i.e the introduction of `transaction.go` and enabling of debug mode for the database, which are a little more contentious. By which I mean it would be good do document the "why" in the git history since we don't forget.

chirino:
I'll split it up in a bit.

chirino:
@dave-tucker just applied your suggestions and rebased.

dave-tucker:
LGTM

PR #319, opened 2023-01-17
Use gomigrate to manage the DB migrations.

chirino:
This allows us to apply or rollback migrations.  This change also manually creates the unique indexes so that migrations work against cockroach.

### DIFF
--- a/internal/database/migrations/20230113-0000.go
+++ b/internal/database/migrations/20230113-0000.go
@@ -1,0 +1,97 @@
+package migrations
+
+import (
+	"github.com/go-gormigrate/gormigrate/v2"
+	"github.com/google/uuid"
+	"github.com/lib/pq"
+	"time"
+)
+
+func addApexTables() *gormigrate.Migration {
+
+	// Migrations rules:
+	//
+	//  1. IDs are numerical timestamps that must sort ascending.
+	//     Use YYYYMMDD-HHMM w/ 24 hour time for format
+	//     Example: August 21 2018 at 2:54pm would be 20180821-1454.
+	//
+	//  2. Include models inline with migrations to see the evolution of the object over time.
+	//     Using our internal type models directly in the first migration would fail in future clean
+	//     installations.
+	//
+	//  3. Migrations must be backwards compatible. There are no new required fields allowed.
+	//
+	// 4. Create one function in a separate file that returns your Migration.
+	migrationId := "20230113-0000"
+
+	type Base struct {
+		ID        uuid.UUID  `gorm:"type:uuid;primary_key;" json:"id" example:"aa22666c-0f57-45cb-a449-16efecc04f2e"`
+		CreatedAt time.Time  `json:"-"`
+		UpdatedAt time.Time  `json:"-"`
+		DeletedAt *time.Time `sql:"index" json:"-"`
+	}
+
+	type Peer struct {
+		Base
+		DeviceID                 uuid.UUID      `json:"device_id" example:"fde38e78-a4af-4f44-8f5a-d84ef1846a85"`
+		ZoneID                   uuid.UUID      `json:"zone_id" example:"2b655c5b-cfdd-4550-b7f0-a36a590fc97a"`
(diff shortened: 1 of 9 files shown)


internal/database/migrations/migrations.go
```diff
@@ -0,0 +1,361 @@
+package migrations
```
dave-tucker, line 1: There is a lot of code here, but it's unclear what its actual purpose is...

internal/database/migrations/20230113-0000.go
```diff
@@ -0,0 +1,97 @@
...
+		CreatedAt  time.Time   `json:"-"`
+		UpdatedAt  time.Time   `json:"-"`
+		DeletedAt  *time.Time  `sql:"index" json:"-"`
+		Devices    []*Device   `json:"-"`
+		DeviceList []uuid.UUID `gorm:"-" json:"devices" example:"4902c991-3dd1-49a6-9f26-d82496c80aff"`
+		ZoneID     uuid.UUID   `json:"zone_id" example:"94deb404-c4eb-4097-b59d-76b024ff7867"`
+		UserName   string      `json:"username" example:"admin"`
+	}
+
+	return CreateMigrationFromActions(migrationId,
```
dave-tucker, line 81: As per the example in the `gormmigrate` repo, what's wrong with simply embedding the structs in this file... calling `tx.AutoMigrate` on those that can be created and using `tx.Exec` to manually run SQL if needed.
It's not clear what the indirection via `CreateMigrationFromActions` and `CreateTableAction` is acheiving.

internal/database/migrations/20230113-0000.go
```diff
@@ -0,0 +1,97 @@
...
+
+func addApexTables() *gormigrate.Migration {
+
+	// Migrations rules:
+	//
+	//  1. IDs are numerical timestamps that must sort ascending.
+	//     Use YYYYMMDD-HHMM w/ 24 hour time for format
+	//     Example: August 21 2018 at 2:54pm would be 20180821-1454.
+	//
+	//  2. Include models inline with migrations to see the evolution of the object over time.
```
dave-tucker, line 18: That seems contrary to what `gomigrate` says, where when adding a column we include a partial struct:
```go
                {
			ID: "201608301415",
			Migrate: func(tx *gorm.DB) error {
				// when table already exists, it just adds fields as columns
				type Person struct {
					Age int
				}
				return tx.AutoMigrate(&Person{})
			},
			Rollback: func(tx *gorm.DB) error {
				return tx.Migrator().DropColumn("people", "age")
			},
		},
```

dave-tucker (commented):
generally this looks good... but I have reservations, mainly because `AutoMigrate` is developer headache free - but we knew this wouldn't scale forever.

a lot of the comments are really just trying to understand the benefit of the abstractions that are being introduced to make this work and trying to figure out what it will mean when one of us needs to add a new field or table into the database. some docs could be added in that regard.

https://github.com/golang-migrate/migrate seems to be more mature, and at least seems to have some dev tooling in place to ease creating migrations.

ultimately though anything would be workable given sufficient docs.

internal/database/migrations/migrations.go
chirino: mostly copied over from kafka fleet manager..  it there in case we want the CLI to finer control over rolling forward and back.  It also has some helpers to make it easier to implement migrations.

internal/database/migrations/20230113-0000.go
chirino: Things get more complicated when you start rolling back.  AutoMigrate works fine for the first time you create a table.  Less so once you start modifying an existing table.

internal/database/migrations/migrations.go
chirino: for example: you can use 
```go
type Person struct {
	Age int
}
AddTableColumnsAction(&Person{})
```
to handle creating the Migrate and Rollback that adds/removes the `age` col.

dave-tucker:
LGTM. Rebased and will merge on green. Thanks for the contribution @chirino

PR #326, opened 2023-01-20
Prevent the node from trying to delete itself as a peer

nerdalert:

### DIFF
--- a/internal/apex/wg.go
+++ b/internal/apex/wg.go
@@ -107,6 +107,10 @@ func (ax *Apex) handlePeerDelete(peerListing []models.Peer) error {
 		if inPeerListing(peerListing, p) {
 			continue
 		}
+		// skip if the key is the same as the node itself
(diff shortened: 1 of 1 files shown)


internal/apex/wg.go
```diff
@@ -107,6 +107,10 @@ func (ax *Apex) handlePeerDelete(peerListing []models.Peer) error {
 		if inPeerListing(peerListing, p) {
 			continue
 		}
+		// skip if the key is the same as the node itself
+		if ax.keyCache[p.DeviceID] == ax.wireguardPubKey {
+			continue
+		}
```
russellb, line 113: We would only hit this if our own peer disappeared from the peer listing, right?  This should never happen. I would rather hold off and fix whatever the root issue is that you're hitting.

If you do want to merge this, I'd at least log an `ERROR` about it since it should never happen.

russellb (changes requested):

PR #338, opened 2023-01-23
Rename apexctl to apex-api-ctl

russellb:
While working on turning the apex agent into a daemon, I am adding a client utilitly to interacting with the daemon over a local socket. I think "apex-ctl" makes the most sense as the utility that controls "apex".

Since this utility is about talking back to the control plane API, I have renamed it to "apex-api-ctl".


dave-tucker (approved):
LGTM

PR #341, opened 2023-01-25
apex: Add unix socket control interface

russellb:
Add a unix socket control interface on linux. This interface uses
    jsonrpc and supports checking the status apexd. It will report back
    the info needed to register the device if it's waiting for auth. This
    would be one way to get info when running apex as a daemon. The other
    would be from the daemon logs.
    
    apexctl now knows how to retrieve this status information by running:
    
        apexctl apexd status
    
    We now embed a version at build time into each binary. This is done
    using ldflags to set the value of main.Version.
    
    apexctl can report its own version.
    
        apexctl version
    
    apexctl can also report the version of a running apexd.
    
        apexctl apexd version
    
    When running subcommands of "apexctl apexd" other than "version",
    apexctl will first validate that the apexctl and apexd versions match.
    
### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,16 +28,25 @@ endif
 ##@ All
 
 .PHONY: all
-all: go-lint yaml-lint markdown-lint ui-lint apexd  ## Run linters and build apexd
+all: go-lint yaml-lint markdown-lint ui-lint apexd apexctl ## Run linters and build apexd
 
 ##@ Binaries
 
 .PHONY: apexd
 apexd: dist/apexd dist/apexd-linux-arm dist/apexd-linux-amd64 dist/apexd-darwin-amd64 dist/apexd-darwin-arm64 dist/apexd-windows-amd64 ## Build the apexd binary for all architectures
 
+.PHONY: apexctl
+apexctl: dist/apexctl dist/apexctl-linux-arm dist/apexctl-linux-amd64 dist/apexctl-darwin-amd64 dist/apexctl-darwin-arm64 dist/apexctl-windows-amd64 ## Build the apex binary for all architectures
+
+APEX_VERSION?=$(shell date +%Y.%m.%d)
+APEX_RELEASE?=$(shell git describe --always)
+APEX_LDFLAGS?=-X main.Version=$(APEX_VERSION)-$(APEX_RELEASE)
+
 COMMON_DEPS=$(wildcard ./internal/**/*.go) go.sum go.mod
 
 APEXD_DEPS=$(COMMON_DEPS) $(wildcard cmd/apexd/*.go)
+
+APEXCTL_DEPS=$(COMMON_DEPS) $(wildcard cmd/apexctl/*.go)
 
 APISERVER_DEPS=$(COMMON_DEPS) $(wildcard cmd/apiserver/*.go)
 
@@ -58,6 +67,11 @@ dist/apexd-%: $(APEXD_DEPS) | dist
 	$(ECHO_PREFIX) printf "  %-12s $@\n" "[GO BUILD]"
 	$(CMD_PREFIX) CGO_ENABLED=0 GOOS=$(word 2,$(subst -, ,$(basename $@))) GOARCH=$(word 3,$(subst -, ,$(basename $@))) \
 		go build -ldflags="$(APEX_LDFLAGS)" -o $@ ./cmd/apexd
+
+dist/apexctl-%: $(APEXCTL_DEPS) | dist
+	$(ECHO_PREFIX) printf "  %-12s $@\n" "[GO BUILD]"
+	$(CMD_PREFIX) CGO_ENABLED=0 GOOS=$(word 2,$(subst -, ,$(basename $@))) GOARCH=$(word 3,$(subst -, ,$(basename $@))) \
+		go build -ldflags="$(APEX_LDFLAGS)" -o $@ ./cmd/apexctl
 
 .PHONY: clean
 clean: ## clean built binaries
--- a/cmd/apexctl/main.go
+++ b/cmd/apexctl/main.go
@@ -19,6 +19,9 @@ const (
 	encodeColumn     = "column"
 )
 
+// This variable is set using ldflags at build time. See Makefile for details.
+var Version = "dev"
+
 func main() {
 	app := &cli.App{
 		Name: "apexctl",
@@ -35,14 +38,12 @@ func main() {
 				Usage: "api server",
 			},
 			&cli.StringFlag{
-				Name:     "username",
-				Required: true,
-				Usage:    "username",
+				Name:  "username",
+				Usage: "username",
 			},
 			&cli.StringFlag{
-				Name:     "password",
-				Required: true,
-				Usage:    "password",
+				Name:  "password",
+				Usage: "password",
 			},
 			&cli.StringFlag{
 				Name:     "output",
@@ -53,6 +54,30 @@ func main() {
 		},
 		Commands: []*cli.Command{
 			{
+				Name:  "version",
+				Usage: "Get the version of apexctl",
+				Action: func(cCtx *cli.Context) error {
+					fmt.Printf("version: %s\n", Version)
+					return nil
+				},
+			},
+			{
+				Name:  "apexd",
+				Usage: "commands for interacting with the local instance of apexd",
+				Subcommands: []*cli.Command{
+					{
+						Name:   "version",
+						Usage:  "apexd version",
+						Action: cmdLocalVersion,
+					},
+					{
+						Name:   "status",
+						Usage:  "apexd status",
+						Action: cmdLocalStatus,
+					},
+				},
+			},
+			{
 				Name:  "organization",
 				Usage: "commands relating to organizations",
 				Subcommands: []*cli.Command{
@@ -61,7 +86,7 @@ func main() {
 						Usage: "list organizations",
 						Action: func(cCtx *cli.Context) error {
 							c, err := client.NewClient(cCtx.Context,
-								cCtx.String("host"),
+								cCtx.String("host"), nil,
 								client.WithPasswordGrant(
 									cCtx.String("username"),
 									cCtx.String("password"),
@@ -97,7 +122,7 @@ func main() {
 						},
 						Action: func(cCtx *cli.Context) error {
 							c, err := client.NewClient(cCtx.Context,
-								cCtx.String("host"),
+								cCtx.String("host"), nil,
 								client.WithPasswordGrant(
 									cCtx.String("username"),
 									cCtx.String("password"),
@@ -125,7 +150,7 @@ func main() {
 						},
 						Action: func(cCtx *cli.Context) error {
 							c, err := client.NewClient(cCtx.Context,
-								cCtx.String("host"),
+								cCtx.String("host"), nil,
 								client.WithPasswordGrant(
 									cCtx.String("username"),
 									cCtx.String("password"),
@@ -157,7 +182,7 @@ func main() {
 						},
 						Action: func(cCtx *cli.Context) error {
 							c, err := client.NewClient(cCtx.Context,
-								cCtx.String("host"),
+								cCtx.String("host"), nil,
 								client.WithPasswordGrant(
 									cCtx.String("username"),
 									cCtx.String("password"),
@@ -190,7 +215,7 @@ func main() {
 						},
 						Action: func(cCtx *cli.Context) error {
 							c, err := client.NewClient(cCtx.Context,
-								cCtx.String("host"),
+								cCtx.String("host"), nil,
 								client.WithPasswordGrant(
 									cCtx.String("username"),
 									cCtx.String("password"),
@@ -215,7 +240,7 @@ func main() {
 						Usage: "list all users",
 						Action: func(cCtx *cli.Context) error {
 							c, err := client.NewClient(cCtx.Context,
-								cCtx.String("host"),
+								cCtx.String("host"), nil,
 								client.WithPasswordGrant(
 									cCtx.String("username"),
 									cCtx.String("password"),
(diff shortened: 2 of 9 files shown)


cmd/apex/main.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"fmt"
+	"net"
+	"net/rpc/jsonrpc"
+)
+
+func main() {
+	// Connect to the UNIX domain socket
+	conn, err := net.Dial("unix", "/var/run/apex.sock")
```
dave-tucker, line 11: This should be `/run/apex.sock` as we found out in bpfd https://github.com/redhat-et/bpfd/commit/ad4c199db395e5a89a1cc243875e18dc1bf9198a

internal/apex/ctlserver.go
```diff
@@ -0,0 +1,84 @@
...
+import (
+	"fmt"
+	"net"
+	"net/rpc"
+	"net/rpc/jsonrpc"
+	"os"
+)
+
+// TODO make this path configurable
+const UnixSocketPath = "/var/run/apex.sock"
```
dave-tucker, line 12: As above

internal/apex/ctlserver.go
```diff
@@ -0,0 +1,84 @@
...
+
+// Assumptions made in here:
+//  - ax.logger is thread-safe
```
dave-tucker, line 15: That is a fine assumption. Yes Zap is thread-safe.

internal/apex/ctlserver.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	return nil
+}
+
+type ApexCtl struct {
+	ax *Apex
+}
+
+func (ac *ApexCtl) Status(_ string, result *string) error {
```
dave-tucker, line 36: @russellb we also need a `func (ac *ApexCtl) Version() {}` RPC since there is a possibility of client-server drift.
i.e we `make dist/apex` but forget to `make dist/apexd`

The version information should be added at compile time with ldflags https://www.digitalocean.com/community/tutorials/using-ldflags-to-set-version-information-for-go-applications

We can roll with just making sure that the client/daemon are the exact same version for now.

internal/apex/ctlserver.go
```diff
@@ -0,0 +1,84 @@
...
+		defer l.Close()
+
+		ac := new(ApexCtl)
+		ac.ax = ax
+		err := rpc.Register(ac)
+		if err != nil {
+			ax.logger.Errorln("Error on rpc.Register(): ", err)
+		}
+
+		for {
```
dave-tucker, line 73: This is bothering me a little bit 😆 

If possible could we make this:
`func (ax *Apex) CtlServerRunLinux(stop <-chan struct{}) error {}`

That way:
1. The caller can pass in a channel to signal that the inner loop should stop
2. The caller will have to use `go CtlServerRunLinux()` to avoid blocking... which means that they could use some thing like `sync.WaitGroup` to ensure that async code has finished before exiting.

internal/apex/ctlserver.go
```diff
@@ -0,0 +1,84 @@
...
+			conn, err := l.Accept()
+			if err != nil {
+				ax.logger.Error("Error on Accept(): ", err)
+				continue
+			}
+			go jsonrpc.ServeConn(conn)
```
dave-tucker, line 79: We should probably use a `sync.WaitGroup` ourselves here.
So:
```go
var wg sync.WaitGroup

select {
    case <-stop:
      // handle shutdown
      // wait for existing rpcs to finish
      wg.Wait()
      return
    default:
      conn, err := l.Accept()
      if err != nil {
	ax.logger.Error("Error on Accept(): ", err)
	continue
     }
     wg.Add(1)
     go jsonrpc.ServeConn(conn)

  }
}
```

dave-tucker (changes requested):
Looking great! A couple of comments but this is great work. Thank you 🙏

Makefile
```diff
@@ -18,24 +24,18 @@ dist:
...
-dist/apexd-windows-amd64: $(APEXD_DEPS) | dist
-	CGO_ENABLED=0 GOOS=windows GOARCH=arm64 go build -o $@ ./cmd/apexd
+dist/apex: $(APEX_CTL_DEPS) | dist
+	CGO_ENABLED=0 go build -o $@ ./cmd/apex
 
 dist/apexctl: $(APEXD_DEPS) | dist
 	CGO_ENABLED=0 go build -o $@ ./cmd/apexctl
 
+dist/apex-%: $(APEX_DEPS) | dist
+	CGO_ENABLED=0 GOOS=$(word 3,$(subst -, ,$(basename $@))) GOARCH=$(word 4,$(subst -, ,$(basename $@))) go build -o $@ ./cmd/apex
```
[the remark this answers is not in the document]
russellb: Good catch! This is wrong. It got lost in the renaming shuffle. This was from when the names were `apex` and `apex-ctl`. The extra `-` in `apex-ctl` meant I had to use different tokens.

cmd/apex/main.go
russellb: Got it, thanks!

internal/apex/ctlserver.go
russellb: done

internal/apex/ctlserver.go
russellb: Thanks :)

internal/apex/ctlserver.go
russellb: Good call. I'll add that. I can also make a `version` subcommand to `apex` while I'm at it.

internal/apex/ctlserver.go
russellb: done

internal/apex/ctlserver.go
russellb: done

internal/apex/ctlserver.go
```diff
@@ -0,0 +1,119 @@
...
+		// We accept connections and send them back over this channel
+		// so we can select waiting on new connections and a message
+		// to stop the CtlServer at the same time.
+		connChan := make(chan net.Conn)
+		go func() {
+			for {
+				conn, err := l.Accept()
+				if err != nil {
+					// TODO - if Accept() returns an error because the listener
+					// has been closed, we should just silently end this goroutine.
```
russellb, line 87: @dave-tucker I didn't see a clean way to do this. Any ideas?

russellb:
@dave-tucker I think this is ready for another look, though I left one inline comment on something I didn't find a clean way to handle yet.

internal/apex/ctlserver.go
```diff
@@ -0,0 +1,119 @@
...
+		err := rpc.Register(ac)
+		if err != nil {
+			ax.logger.Errorln("Error on rpc.Register(): ", err)
+		}
+
+		// We accept connections and send them back over this channel
+		// so we can select waiting on new connections and a message
+		// to stop the CtlServer at the same time.
+		connChan := make(chan net.Conn)
+		go func() {
```
dave-tucker, line 82: You've spawned another goroutine here without adding to the waitgroup.

internal/apex/ctlserver.go
dave-tucker: Yeah from the previous round of comments I think I'd mentioned keeping this in a `RunCtlServer` function which was a blocking function. i.e
```go
func runCtlServer(wg *sync.Waitgroup, stopCh <-chan struct{}) error {}
```

The logic would be roughly:

1. Have a `for { select {} }` that has 2 cases: `case <-ax.ctlServerStop:` which waits will run when we're shutting down and `default` which is run when no other cases are available i.e we're not shutting down.
1. Any error in `conn, err := l.Accept()` can be returned directly i.e `return err`
2. If no error from `Accept()`, increment the waitgroup by 1 and spawn a goroutine to service the connection - that goroutine has a `defer wg.Done()`


That way, we can handle any errors that happen properly... We might want to try and run the server again for example.

internal/apex/ctlserver.go
```diff
@@ -0,0 +1,119 @@
...
+	ax.wg.Add(1)
+	go func() {
+		defer l.Close()
+		defer ax.wg.Done()
+
+		ac := new(ApexCtl)
+		ac.ax = ax
+		err := rpc.Register(ac)
+		if err != nil {
+			ax.logger.Errorln("Error on rpc.Register(): ", err)
```
dave-tucker, line 75: This seems like a fatal error right? We should be returning here and not just logging

russellb: Oops, yes.

internal/apex/ctlserver.go
russellb: I pushed another iteration, but it's not quite what you described here. I do have it properly catching errors and restarting the control interface routine if an error occurs. Take a look at your leisure and then let's jump on video chat to talk through feedback.

Thanks again for the detailed reviews!

russellb: I also left the rudimentary state tracking. I figured we can replace it in a follow up PR.

russellb:
just rebasing for conflicts with `main`

internal/apex/apex.go
```diff
@@ -59,6 +69,14 @@ type Apex struct {
...
 	symmetricNat         bool
 	logger               *zap.SugaredLogger
+	// See the ApexStatus* constants
+	status    int
+	statusMsg string
+	version   string
+	// Used to tell the CtlServer to stop for shutdown
+	ctlServerStop chan struct{}
+	// Used during shutdown to wait for CtlServer to stop
+	wg sync.WaitGroup
```
chirino, line 79: You don't need ctlServerStop or wg fields here.  You can pass a ctx and wg as arguments to the CtlServerStart function.

internal/apex/apex.go
```diff
@@ -404,6 +436,8 @@ func (ax *Apex) relayStateReconcile(zoneID uuid.UUID) error {
 }
 
 func (ax *Apex) Shutdown(ctx context.Context) error {
+	ax.ctlServerStop <- struct{}{} // Tell CtlServer to stop cleanly
+	ax.wg.Wait()                   // Wait for CtlServer to stop
 	return nil
```
chirino, line 441: you won't need this when you rebase.

internal/apex/ctlserver.go
```diff
@@ -0,0 +1,147 @@
...
+				err = <-errChan
+				ax.logger.Error("Error on Accept(): ", err)
+				break
+			}
+			wg.Add(1)
+			go func() {
+				defer wg.Done()
+				jsonrpc.ServeConn(conn)
+			}()
+		case <-ax.ctlServerStop:
```
chirino, line 132: this will become `case <-ctx.Done():`

chirino (commented):
added some comments on how to update if you rebase on #391

russellb:
> added some comments on how to update if you rebase on #391

Thank you for the review, @chirino ! Il'l rebase this on #391 after I merge it.

russellb:
rebased after merging #391 which simplified things a bit

russellb:
rebased to resolve conflicts with main

cmd/apex/main.go
```diff
@@ -0,0 +1,94 @@
...
+				Usage:   "Get the version of apex and apexd",
+				Action:  cmdVersion,
+			},
+			{
+				Name:    "status",
+				Aliases: []string{"s"},
+				Usage:   "Get the status of apexd",
+				Action:  cmdStatus,
+			},
+		},
```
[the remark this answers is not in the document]
russellb: Yeah, that's an option, too. I split it out to have a clear separation between the tool that works with the control plane API and the one that talks to a local daemon. I thought mixing them could be confusing.

internal/apex/ctlserver.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+func (ax *Apex) CtlServerLinuxRun(ctx context.Context, ctlWg *sync.WaitGroup) error {
+	os.Remove(UnixSocketPath)
+	l, err := net.ListenUnix("unix", &net.UnixAddr{Name: UnixSocketPath, Net: "unix"})
+	if err != nil {
+		ax.logger.Error("Error creating unix socket: ", err)
+		return err
+	}
+
```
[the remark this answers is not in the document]
russellb: I thought I had a good reason for doing it this way. If I did, it doesn't seem to apply anymore!  I'll change this.

russellb:
I will rebase this again after #429 goes in.

cmd/apex/main.go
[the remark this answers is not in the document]
chirino: As a user I would rather have 1 cli tool that I use to interact with the whole system.  It can have lots of sub commands that interact with different things.

chirino (approved):
LGTM

russellb:
@chirino can you take another look? I took your suggestion to merge the client functionality into `apexctl` instead of making a separate `apex`.

cmd/apexctl/main.go
```diff
@@ -52,6 +53,30 @@ func main() {
...
+				Usage: "commands for interacting with the local instance of apexd",
+				Subcommands: []*cli.Command{
+					{
+						Name:   "version",
+						Usage:  "apexd version",
+						Action: cmdLocalVersion,
+					},
+					{
+						Name:   "status",
+						Usage:  "apexd status",
```
chirino, line 75: love it.

chirino (approved):
Still LGTM.

We can probably use this to more reliably wait for hub routers to go online instead of doing this sleeps like this: https://github.com/redhat-et/apex/blob/main/integration-tests/integration_test.go#L339

russellb:
> Still LGTM.
> 
> We can probably use this to more reliably wait for hub routers to go online instead of doing this sleeps like this: https://github.com/redhat-et/apex/blob/main/integration-tests/integration_test.go#L339

Yeah, that's a good idea!

PR #345, opened 2023-01-26
SaaS-ify the Database Schema

dave-tucker:
Based on discussions last week we wanted to:
1. Rename Zone to Organization
2. Remove Peers and collapse into Device
3. For every User we see, create a dedicated Organization for them
4. (If feature-flag enabled) Allow creation of New Organizations
5. (Not implemented) Allow user to join another Organization
6. (Not implemented) If a User is a member of multiple organizations, they must configure which org they want their devices onboarded to. Default will be their own Org.

Issue #348 


russellb (changes requested):
To recap our conversation via chat, I'd like to get a design doc in `docs/design/` that covers the plan for this schema change. If this is phase 1 of a multi-phase effort, it would be great to get that mapped out to ensure everyone is on the same page. It would be easiest to do that in another PR so we can iterate and merge it separately and ahead of this one. Thanks!

dave-tucker:
Design doc is in #396

internal/handlers/organization.go
```diff
@@ -0,0 +1,293 @@
...
+	// Create the organization
+	if err := api.ipam.AssignPrefix(ctx, request.IpCidr); err != nil {
+		c.JSON(http.StatusInternalServerError, models.ApiError{Error: err.Error()})
+		return
+	}
+	newOrganization := models.Organization{
+		Name:        request.Name,
+		Description: request.Description,
+		IpCidr:      request.IpCidr,
+		HubZone:     request.HubZone,
```
nerdalert, line 81: Default to hub zone true. :exploding_head:
```
HubZone:     true,
```

nerdalert (approved):
Im ok with the yolo breakage. Flipping the default org to a hub-zone should funnel the agent into requiring a discovery/relay node.

PR #346, opened 2023-01-27
Makefile: Replace kind.sh functionality

dave-tucker:
Since we're now exclusively using kind, bring that functionality into the Makefile. This allows for:

1. Dependencies: So a dev cluster will always build images before deploy
2. Re-use: We can re-use our make targets once the cluster is up for deploying, clearing the database etc...


dave-tucker:
YOLO'ing this one. Will deal with the fallout later 😆

PR #359, opened 2023-01-30
Handle http errors in peer listings

nerdalert:
- Prevents peer deletion if the peer lookup fails and throws an error instead of fataling out.

### DIFF
--- a/internal/apex/apex.go
+++ b/internal/apex/apex.go
@@ -265,7 +265,7 @@ func (ax *Apex) Run() error {
 		if err := ax.Reconcile(user.ZoneID, false); err != nil {
 			// TODO: Add smarter reconciliation logic
 			// to handle disconnects and/or timeouts etc...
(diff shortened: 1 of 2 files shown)


internal/apex/apex.go
```diff
@@ -265,7 +265,7 @@ func (ax *Apex) Run() error {
 		if err := ax.Reconcile(user.ZoneID, false); err != nil {
 			// TODO: Add smarter reconciliation logic
 			// to handle disconnects and/or timeouts etc...
-			return err
+			ax.logger.Error(err)
```
russellb, line 268: How about adding a prefix to the log message that makes it a bit more clear what the context of the error is?

```suggestion
			ax.logger.Errorf("Failed to reconcile state with the apex API server: ", err)
```

russellb (changes requested):

PR #360, opened 2023-01-30
Redploy DB via makefile and makefile docs updates

nerdalert:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -160,7 +160,7 @@ redeploy: load-images # Redploy apex after images changes
 	@kubectl rollout restart deploy/frontend -n apex
 
 .PHONY: recreate-db
-recreate-db: recreate-db # Delete and bring up a new apex database
+recreate-db: recreate-db ## Delete and bring up a new apex database
 	@kubectl delete -n apex deploy/apiserver postgrescluster/database deploy/ipam
 	@kubectl apply -k ./deploy/apex/overlays/dev
 
--- a/Makefile
+++ b/Makefile
@@ -159,6 +159,11 @@ redeploy: load-images # Redploy apex after images changes
 	@kubectl rollout restart deploy/apiserver -n apex
 	@kubectl rollout restart deploy/frontend -n apex
 
+.PHONY: recreate-db
+recreate-db: recreate-db # Delete and bring up a new apex database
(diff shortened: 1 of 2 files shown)


docs/README.md
```diff
@@ -46,11 +46,12 @@ echo "127.0.0.1 auth.apex.local api.apex.local apex.local" | sudo tee -a /etc/ho
 
 #### Deploy using KIND
 
-You should first ensure that you have `kind` and `kubectl` installed.
+This section is only if you want to build the controller stack, if you want to attach to a running controller see [The Apex Agent](#the-apex-agent).
+You should first ensure that you have `kind`, `kubectl` and [`mkcert`](https://github.com/FiloSottile/mkcert) installed.
```
russellb, line 50:
```suggestion
> **Note**
> This section is only if you want to build the controller stack. If you want to attach to a running controller, see [The Apex Agent](#the-apex-agent).

You should first ensure that you have `kind`, `kubectl` and [`mkcert`](https://github.com/FiloSottile/mkcert) installed.
```

For nice formatting: https://github.com/community/community/discussions/16925

Makefile
```diff
@@ -159,6 +159,12 @@ redeploy: load-images # Redploy apex after images changes
 	@kubectl rollout restart deploy/apiserver -n apex
 	@kubectl rollout restart deploy/frontend -n apex
 
+.PHONY: redeploy-db
+redeploy-db: redeploy-db # Delete and bring up a new apex database
```
russellb, line 163: I know the docs include "Delete" here, but the `redeploy` target isn't destructive to data, so someone might mistakenly think `redeploy-db` restarts the DB, but with the same existing data. Maybe we could call it something like `recreate-db` ?

dave-tucker (approved):
LGTM

Makefile
```diff
@@ -159,6 +159,11 @@ redeploy: load-images # Redploy apex after images changes
 	@kubectl rollout restart deploy/apiserver -n apex
 	@kubectl rollout restart deploy/frontend -n apex
 
+.PHONY: recreate-db
+recreate-db: recreate-db # Delete and bring up a new apex database
```
russellb, line 163:
```suggestion
recreate-db: recreate-db ## Delete and bring up a new apex database
```

It won't show up in help text with a single `#`

russellb (changes requested):

PR #362, opened 2023-01-31
Makefile: Fix help text for some targets

russellb:
A number of targets had docs with a single `#`, but the help text output generation looks for `##`.


dave-tucker (approved):
LGTM

PR #364, opened 2023-01-31
Publish apexctl binary to AWS s3 bucket and

vishnoianil:
update docs with installation instructions.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -189,18 +189,27 @@ jobs:
         run: |
           go test -v ./internal/apex/...
 
-      - name: Build
-        id: build-apex
+      - name: Build Apexd
+        id: build-apexd
         run: |
           CGO_ENABLED=0 go build -o apexd-$(go env GOOS)-$(go env GOARCH) -v ./cmd/apexd
           echo "artifact-name=apexd-$(go env GOOS)-$(go env GOARCH)" >> $GITHUB_OUTPUT
+        shell: bash
+
+      - name: Build Apexctl
+        id: build-apexctl
+        run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -189,18 +189,27 @@ jobs:
...
+        id: build-apexd
         run: |
           CGO_ENABLED=0 go build -o apexd-$(go env GOOS)-$(go env GOARCH) -v ./cmd/apexd
           echo "artifact-name=apexd-$(go env GOOS)-$(go env GOARCH)" >> $GITHUB_OUTPUT
         shell: bash
 
+      - name: Build Apexctl
+        id: build-apexctl
+        run: |
+          CGO_ENABLED=0 go build -o apexctl-$(go env GOOS)-$(go env GOARCH) -v ./cmd/apexctl
```
russellb, line 202: This isn't a problem for this PR, but I wish we weren't duplicating `go` commands in `build.yaml` and the `Makefile`. I'd love to change these to use `make`, but we can figure that out later ...

docs/README.md
```diff
@@ -88,6 +91,29 @@ CAROOT=$(pwd)/.certs mkcert -install
 
 For windows, we recommend installing the root certificate via the [MMC snap-in](https://learn.microsoft.com/en-us/troubleshoot/windows-server/windows-security/install-imported-certificates#import-the-certificate-into-the-local-computer-store).
 
+## Deploying the Apexctl Utility
```
russellb, line 94:
```suggestion
## Using the Apexctl Utility
```

How about "Using" instead of "Deploying."  I suggest that because I don't think of CLI utilities as something deployed.

russellb (approved):

PR #365, opened 2023-02-01
Properly update the hub-router status of the device

### LINKED ISSUE #183: [Bug] Joining as a regular node and then rejoining as a relay node does not update the peer table
Peer table needs to be updated on the re-join.

vishnoianil:
If a device joins as a normal device and after a disconnect it joins again as a hub-router, the peer list doesn't update the hub-router status in the GUI. Same happens if the device joins as a hub router and then as a normal device after disconnect.

Resolves #183 


russellb:
This is a good candidate for unit test coverage, if you feel like looking into that as a follow-up

PR #369, opened 2023-02-01
Add offline_access option for OIDC token refresh

nerdalert:
- To fetch the original token to refresh the token grant after the default 10hrs we need to enable the oidc offline_access option


vishnoianil (approved):

russellb:
For reference, I had to look up what this was about ...

https://openid.net/specs/openid-connect-core-1_0.html#OfflineAccess

PR #391, opened 2023-02-06
chore: apexd should shut down cleaner on an os signal.

chirino:
Make the api a little more go idiomatic by using a Context cancellation to signal to *Apex that it should shut down and wait for the cancellation to occur using a WaitGroup.

Signed-off-by: Hiram Chirino <hiram@hiramchirino.com>

### DIFF
--- a/internal/apex/apex.go
+++ b/internal/apex/apex.go
@@ -3,11 +3,13 @@ package apex
 import (
 	"context"
 	"fmt"
+	"github.com/redhat-et/apex/internal/util"
 	"net"
 	"net/url"
 	"os"
 	"reflect"
 	"runtime"
+	"sync"
 	"time"
 
 	"github.com/google/uuid"
@@ -168,7 +170,7 @@ func NewApex(ctx context.Context, logger *zap.SugaredLogger,
 	return ax, nil
 }
 
-func (ax *Apex) Run() error {
+func (ax *Apex) Start(ctx context.Context, wg *sync.WaitGroup) error {
 	var err error
 
 	if err := ax.handleKeys(); err != nil {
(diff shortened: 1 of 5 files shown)


internal/apex/apex.go
```diff
@@ -403,10 +405,6 @@ func (ax *Apex) relayStateReconcile(zoneID uuid.UUID) error {
 	return nil
 }
 
-func (ax *Apex) Shutdown(ctx context.Context) error {
-	return nil
-}
```
russellb, line 408: Could you take a look at this PR?  https://github.com/redhat-et/apex/pull/341

I was using Apex.Shutdown() in there. I want to make sure I understand how to adjust that PR to work with this. It looks like I could select on `<-ctx.Done()` instead of the custom channel I had created to write to from Apex.Shutdown() ?

russellb: I'm pretty sure I've answered my own question, haha. Feel free to comment here or on my PR if I've missed something. I'll review this again later today.

chirino: If you want I can rebase on top of your PR

russellb: No, I think I'd rather merge this first. I like what you've done and it'll simplify mine a little bit.

internal/util/go_with_wait_group_test.go
```diff
@@ -0,0 +1,36 @@
...
+
+	counter1 := atomic.Int32{}
+	for i := 0; i < 10; i++ {
+		// run a few functions in parallel
+		util.GoWithWaitGroup(nil, func() {
+			time.Sleep(200 * time.Millisecond)
+			counter1.Add(1)
+		})
+	}
+	assert.Equal(t, int32(0), counter1.Load())
```
russellb, line 22: Is there some way we can know this would never fail? It seems like it could fail if for some reason one of the goroutines was scheduled after 200 milliseconds and before getting to this assertion. I know it seems unlikely, but who knows, maybe under really heavy load?

A couple of alternatives:

- Just rely on the test not crashing on a nil waitgroup.

- Right now there's no check that verifies that the goroutine ran. You could drop the sleep inside the routines and do a different type of assertion.  In psuedo-code:

```
for {
    break if the counter has reached 10
    assert a test failure if we've waited too long (10 seconds or something absurdly long)
    otherwise sleep a bit and check again
}
```

chirino: good idea.. I just took a crack it doing all of that.  should be better.

internal/util/go_with_wait_group_test.go
```diff
@@ -0,0 +1,40 @@
...
+	}
+
+	done := atomic.Bool{}
+	go func() {
+		wg.Wait()
+		done.Store(true)
+	}()
+
+	// should not be done yet
+	assert.False(t, done.Load())
```
russellb, line 34: It seems *possible* for this to fail (unlikely, I know), and the assert.Eventually() you have below still checks to make sure that the waitgroup successfully finished. I'd just drop this assert.

russellb (changes requested):

russellb:
lgtm, will merge once CI finishes

PR #392, opened 2023-02-06
Added architecture summary document

srampal:
Helps provide some overall architecture context which wasnt in the docs yet.


docs/README.md
```diff
@@ -45,6 +45,10 @@
 - **Controller** - The Controller is the hosted service that handles authentication, authorization, management of zones, enrollment of nodes, and coordination among nodes to allow them to peer with other nodes.
 - **Agent** - The Agent runs on any node which wants to join an Apex Zone.
 
+## High Level Solution Description
```
russellb, line 48: The table of contents should be updated when adding new sections to this doc.

docs/design/solution-architecture.md
```diff
@@ -0,0 +1,14 @@
+# High Level Solution Architecture
+
+The following figure illustrates briefly how the overall solution works. ![Apex Solution](./apex-architecture.png).
```
russellb, line 3: Is there a source for the diagram? We'll need something that others can modify and update if needed.

Something to check out is https://github.blog/2022-02-14-include-diagrams-markdown-files-mermaid/ which would have the diagram automatically generated. That would be ideal if you can get it to display what you want.

docs/design/solution-architecture.md
```diff
@@ -0,0 +1,14 @@
...
+
+The figure illustrates two edge nodes N1 and N2 that are each running in their own respective 
+private networks in which private IP addressesing is used (based on RC 1918 for instance), and hence these nodes/ devices are not directly addressable from outside their own private networks.  Access to the public internet is via NAT gateways in each private network that perform SNAT (Source Network Address Translation), translating the private source IP address and source TCP and UDP ports of packets originated by devices inside the private network when sending traffic to external destinations on the public network or Internet. In the example shown, NAT-gw-1 performs SNAT and swaps in its public facing IP address (3.100.10.10) and an available TCP/ UDP port on this IP address for traffic originated within the private network and destined for the public Internet.
```
russellb, line 6: s/RC/RFC/

docs/design/solution-architecture.md
```diff
@@ -0,0 +1,14 @@
...
+
+This was a brief summary of the overall architecture and operation of the Apex solution. There are additional details related to how end devices register with the the controller and secure cryptographic identification of end points using X.509 certificates whicih are not covered here for the sake of brevity. Also note that at the time of writing, the Apex architecture is still evolvign and some concepts described here may change eventually. For instance in the current desig, the use of the relay node is optional within an Apex zone and STUN servers are queried to determine public IP mapping information. A future version of this design may be altered to mandate the use of the relay nodes and also use these for public IP information discovery without depending on public STUN servers. The purpose of this summary is to illustrate the core concepts while recognizing that some of the details may change as the architecture evolves.
```
russellb, line 13: s/desig/design/

s/evolvign/evolving/

I'd run a spell check on the whole thing in case I missed something else when reading it.

russellb (changes requested):
I left a few minor comments inline, but my overall thought is there's a lot of overlap in this document with `docs/design/design-discovery.md`. There's a heavy emphasis on how the tunnel setup works, which that doc also covers differently.

I also feel that this feels more like something to go in `docs/design/` than linked from the user-oriented `docs/README.md`.

Providing an architecture overview at this spot in the user-oriented docs makes sense, but this seems to get into too many networking details for what I would imagine there. Instead, I would envision something introducing the components at a high level and how they relate to each other.

docs/design/solution-architecture.md
srampal: I agree editable diagrams are preferable to images where possible. That mermaid utility looks interesting, thanks for the pointer. However it seems more suited for simpler flowcharts than arbitrary diagrams like what we have in the PR currently with clouds and arbitrary end to end curved lines like the wireguard tunnel.

russellb: I looked around a bit. I see some example that people have made that look pretty good. Here's one that's networking oriented.

https://github.com/mermaid-js/mermaid/issues/1227#issuecomment-1192060275

srampal:
> I left a few minor comments inline, but my overall thought is there's a lot of overlap in this document with `docs/design/design-discovery.md`. There's a heavy emphasis on how the tunnel setup works, which that doc also covers differently.
> 
> I also feel that this feels more like something to go in `docs/design/` than linked from the user-oriented `docs/README.md`.
> 
> Providing an architecture overview at this spot in the user-oriented docs makes sense, but this seems to get into too many networking details for what I would imagine there. Instead, I would envision something introducing the components at a high level and how they relate to each other.

1) The main motivation of the PR is that there was no document so far that would allow someone approaching the repo the first couple of times could read to get the overall end to end picture of the solution with a few minutes of reading. Thats the issue this PR is trying to address. Hence a pointer is needed early on but instead of putting the text right there and adding content to the already busy README, it is better off as a link to a separate page right after some of the concepts are introduced in the README.

> I left a few minor comments inline, but my overall thought is there's a lot of overlap in this document with `docs/design/design-discovery.md`. There's a heavy emphasis on how the tunnel setup works, which that doc also covers differently.
> 

There may be some overlap. However I dont believe the discovery document gives the end to end picture that one needs to  start with. 

> I also feel that this feels more like something to go in `docs/design/` than linked from the user-oriented `docs/README.md`.
> 
> Providing an architecture overview at this spot in the user-oriented docs makes sense, but this seems to get into too many networking details for what I would imagine there. Instead, I would envision something introducing the components at a high level and how they relate to each other.

This does attempt to represent pictorially and in text, all the components at a high level and where they sit in an end to end solution, as well as the core problem being solved. I tried to balance between adding some detail relevant to the end to end view and yet leaving out other details like the cryptographic identities. Will take another pass at modifying the flow of the content a bit.

russellb:
For this part of the docs, I would suggest a focus on "what" and very little about "how". For example, I would consider any mention of STUN to be "how".  The "what" for apexd is just that it uses various networking techniques to establish connectivity to other devices in the same Apex zone. I hope that helps.

docs/design/solution-architecture.md
```diff
@@ -0,0 +1,14 @@
...
...abling every node to determine the public reachability information of every other node within the Apex zone (subject to any additional administered policies that may filter distribution of reachability information within a zone).  The apex agents can then attempt to initiate Wireguard  tunnel sessions with each other's public IP information. In most common cases, this attempt shoudl succeed since the NAT gateways at each location have already opened up these public facing ports for the outbound connections initiated previously by the Apex agents when connecting to the Apex controller and relay nodes. In the case where the NAT gateways execute a stricter form of Source NATting known as Symmetric NAT, it may not be possible for apex nodes to initiate wireguard tunnels directly with each other, in which case they have logic to revert to sending traffic to the relay node and having it perform an IP routing operation towards the target Apex node. As far as end user application traffic on each node is concerned, it is routed to use the Wireguard virtual interface and get encrypted and tunneled towards the desired destination wireguard interface IP by the wireguard interface driver logic.
```
vishnoianil, line 11: 2nd technique need a bit of change. The hub router needs to join as a first device. Than it become the peer to all the other nodes in that zone. And ituses the wireguard state (wgctl dump) to find out the endpoints and publish it to controller (api server).

vishnoianil (commented):
Overall it looks good to me, and it can be a precursor document for the connectivity documented that i wrote. Once that document is ready to go in, we can refer that document from this doc.

@russellb I will converge my connectivity doc with discovery doc and we can update this document to refer user to that detailed document if they want to dig in details. wdyt?

vishnoianil:
@srampal Just a follow-up to address the comments in this PR, so that we can move this PR.

PR #394, opened 2023-02-07
fix: `make run-test-container` now works on osx

chirino:
Added info message when you start the test container so that beginners know what command they should run to get apexd up.

Added net-tools to the image incase you need to do some network troubleshooting. 

Fixed a Makefile message that referred to hack/kind/kind.sh

### DIFF
--- a/hack/resolve-ip/main.go
+++ b/hack/resolve-ip/main.go
@@ -8,7 +8,7 @@ import (
 
 func main() {
 
-	if len(os.Args) < 2 {
+	if len(os.Args) != 2 {
 		fmt.Fprintln(os.Stderr, "host argument missing")
 		os.Exit(1)
 	}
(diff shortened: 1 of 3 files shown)


hack/resolve-ip/main.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	"fmt"
+	"net"
+	"os"
+)
+
+func main() {
+
+	if len(os.Args) < 2 {
```
russellb, line 11:
```suggestion
	if len(os.Args) != 2 {
```

hack/resolve-ip/main.go
```diff
@@ -0,0 +1,26 @@
...
+		fmt.Fprintln(os.Stderr, "host argument missing")
+		os.Exit(1)
+	}
+	host := os.Args[1]
+	addrs, err := net.LookupHost(host)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "error: %+v\n", err)
```
russellb, line 18:
```suggestion
		fmt.Fprintf(os.Stderr, "error: %w\n", err)
```

related: https://github.com/redhat-et/apex/pull/389

russellb (changes requested):

PR #395, opened 2023-02-09
Minor fixes related to comments, debug logs and some minor code optim…

vishnoianil:
…izations


internal/apex/apex.go
```diff
@@ -543,13 +536,13 @@ func (ax *Apex) findLocalEndpointIp() (string, error) {
...
 func binaryChecks() error {
-	// Windows userspace binary
+	// Windows user space binary
 	if GetOS() == Windows.String() {
 		if !IsCommandAvailable(wgWinBinary) {
 			return fmt.Errorf("%s command not found, is wireguard installed?", wgWinBinary)
 		}
 	}
-	// Darwin wireguard-go userspace binary
+	// Darwin wireguard-go user space binary
```
russellb, line 545:
```suggestion
	// Darwin wireguard-go userspace binary
```

I would add userspace to the dictionary.

russellb (changes requested):
One minor change suggested.

This will conflict with #389, as well.

internal/apex/apex.go
```diff
@@ -550,7 +543,7 @@ func (ax *Apex) findLocalEndpointIp() (string, error) {
 
 // binaryChecks validate the required binaries are available
 func binaryChecks() error {
-	// Windows userspace binary
+	// Windows user space binary
```
russellb, line 546:
```suggestion
	// Windows userspace binary
```

russellb (approved):
I'm good with this.

In the future, it would be ideal to separate text-only changes (formatting, spelling) from any real code changes (where you simplified duplication). It's important that the code changes get a closer look and sometimes those can get overlooked if they're in the middle of a bunch of text changes. Hopefully that makes sense!

vishnoianil:
> 

Yup, makes sense.

Issue #348, opened 2023-01-27
Make the database SaaS-ier


dave-tucker:
Based on discussions last week we wanted to:

1. Rename Zone to Organization
1. Remove Peers and collapse into Device
1. For every User we see, create a dedicated Organization for them
1. Allow creation of New Organizations
1. Allow user to join another Organization
1. If a User is a member of multiple organizations, they must configure which org they want their devices onboarded to. Default will be their own Org.

This makes it easy for us to operate a single instance that is multi-tenanted

PR #396, opened 2023-02-09
docs: Add database design

dave-tucker:
Updates: #348 

### DIFF
--- a/docs/design/database.md
+++ b/docs/design/database.md
@@ -67,6 +67,8 @@ This model was built under the assumption that multi-tenancy would be achieved b
 
 ## v2: SaaS Preview Milestone
 
+See [issue #348](https://github.com/redhat-et/apex/issues/348).
+
 In order to create a more SaaS-ready schema, we've made the following changes.
 
 1. Replace the `Zone` construct with `Organization`. This is a GitHub-like construct, where each unique `User` is also its own `Organization`.
--- a/docs/design/database.md
+++ b/docs/design/database.md
@@ -66,11 +66,11 @@ On top of this flow, we were somewhat abusing `POST /zones/$id/peers` to also up
 
 This model was built under the assumption that multi-tenancy would be achieved by deploying an Apex stack per-tenant.
 
-# v2: SaaS (Milestone-1)
+# v2: SaaS Preview Milestone
 
 In order to create a more SaaS-ready schema, we've made the following changes.
 
-1. Replace the `Zone` construct with `Organization`. This is a GitHub like construct, where each unique `User` is also it's own `Organization`.
+1. Replace the `Zone` construct with `Organization`. This is a GitHub-like construct, where each unique `User` is also its own `Organization`.
 1. Deprecate `Peer` and combine it's functionality with the `Device` table.
 
 ```mermaid
@@ -117,7 +117,7 @@ Since a `User` is also an `Organization`, once someone has registered to Apex we
 
 The device-onboarding flow is as follows:
 
-1. A User registers for Apex using an sign-up form on https://apex-hosted.cloud, with the option to use SSO from Google, Github, Facebook etc..
+1. A User registers for Apex using a sign-up form on <https://apex-hosted.cloud>, with the option to use SSO from Google, Github, Facebook etc..
 1. User downloads and runs the Apex agent
 1. Apex agent does a `POST /device` to register the unique device with Apex for the user - it's onboarded into their personal organization.
 1. `GET /organization/$id/devices` is polled to retrieve the peer listing for a zone.
--- a/docs/design/database.md
+++ b/docs/design/database.md
@@ -1,0 +1,133 @@
+Database Schema
+===============
+
+# v1: PoC
+
+In v1, the schema looks as follows:
+
+```mermaid
+erDiagram
+    USER ||--o{ DEVICE : has
+    USER ||--|| ZONE : belongs_to
+    DEVICE ||--o{ PEER: has
+    ZONE ||--o{ PEER: contains
+    USER{
+        string id
+        string username
+    }
+    DEVICE{
+        string id
+        string user_id
+        string public_key
+        string hostname
+    }
+    ZONE{
+        string id
+        string name
+        string description
+        string cidr
+        string hub_zone
+    }
+    PEER{
+        string id
+        string device_id
+        string zone_id
+        string endpoint_ip
+        string allowed_ips
+        string node_address
+        string child_prefix
+        string hub_router
+        string hub_zone
+        string zone_prefix
+        string reflexive_ipv4
+        string endpoint_local_address_ipv4
+        string symmetric_nat
+    }
+```
+
+We have 4 concepts:
+
+- Users: Unique users of the service
+- Devices: A unique end-user device
+- Zone: A way of describing which devices may communicate with eachother
+- Peer: An association between a Device and a Zone. This is used by the apex agent to identify who it may peer with.
+
+Since connectivity was Zone-based, we would create a "Default Zone" where all devices would be onboarded into. Adding > 1 Zone was disabled, since there were technical challenges around moving a device from one Zone to another, or for a device to have peers in > 1 Zone.
+
+The device-onboarding flow was as follows:
+
+1. Operator/Customer sets up OIDC to provide users with access to Apex
+1. User runs the Apex agent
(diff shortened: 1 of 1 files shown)


docs/design/database.md
```diff
@@ -0,0 +1,133 @@
+Database Schema
+===============
```
russellb, line 2: I think markdown lint will complain about this. It doesn't like mixing heading styles (underlining vs. the `#` prefixes)

docs/design/database.md
```diff
@@ -0,0 +1,133 @@
...
+1. User runs the Apex agent
+1. Apex agent does a `POST /device` to register the unique device with Apex for the user. It is assigned into the single zone that a user is in (either the default zone, or a new zone where the user was moved to by an admin).
+1. Once the device is registered a `POST /zones/$id/peers` was issued to create an association between this device and a zone.
+1. `GET /zones/$id/peers` was polled to retrieve the peer listing for a zone.
+
+On top of this flow, we were somewhat abusing `POST /zones/$id/peers` to also update peers after NAT-T.
+
+This model was built under the assumption that multi-tenancy would be achieved by deploying an Apex stack per-tenant.
+
+# v2: SaaS (Milestone-1)
```
russellb, line 69:
```suggestion
# v2: SaaS Preview Milestone
```

docs/design/database.md
```diff
@@ -0,0 +1,133 @@
...
+
+In order to create a more SaaS-ready schema, we've made the following changes.
+
+1. Replace the `Zone` construct with `Organization`. This is a GitHub like construct, where each unique `User` is also it's own `Organization`.
```
russellb, line 73:
```suggestion
1. Replace the `Zone` construct with `Organization`. This is a GitHub-like construct, where each unique `User` is also its own `Organization`.
```

docs/design/database.md
```diff
@@ -0,0 +1,133 @@
...
+
+- Users: Unique users of the service
+- Devices: A unique end-user device
+- Organization: A collection of users who have permissions to onboard one or more devices into this organization
+
+Since a `User` is also an `Organization`, once someone has registered to Apex we onboard devices into their personal organization.
+
+The device-onboarding flow is as follows:
+
+1. A User registers for Apex using an sign-up form on https://apex-hosted.cloud, with the option to use SSO from Google, Github, Facebook etc..
```
russellb, line 120:
```suggestion
1. A User registers for Apex using a sign-up form on <https://apex-hosted.cloud>, with the option to use SSO from Google, Github, Facebook etc..
```

markdownlint will complain about raw URLs.

docs/design/database.md
```diff
@@ -0,0 +1,133 @@
...
+1. User downloads and runs the Apex agent
+1. Apex agent does a `POST /device` to register the unique device with Apex for the user - it's onboarded into their personal organization.
+1. `GET /organization/$id/devices` is polled to retrieve the peer listing for a zone.
+1. An dedicated API endpoint was added to update device information if the public-facing address changes etc...
+
+A key point here is that there is no admin intervention required to onboard a new user as anyone can sign-up and register.
+
+Supporting a user belonging to multiple organizations is easy in this model since a `Device` may only be in one `Organization` at a time. A `User` may chose to move their device at any time.
```
russellb, line 128: ... but actually supporting this isn't expected at first, right?  You could call the "multi-organization" support v2.5 or something?

docs/design/database.md
```diff
@@ -0,0 +1,133 @@
...
+
+
+# v3: (TBD) Milestone-2
+
+In our v3 schema we seek to add a few constructs like `Group` and `Network` to allow for more fine-grained connection policies.
```
russellb, line 133: I don't want to bikeshed `Group` and `Network` name choices. How about just listing objectives we expect to go after next in future iterations?

```
In future versions of the schema, we seek to evolve the schema to support some additional key features:
* Allowing finer-grained connectivity grouping for devices within the same organization. We may want to allow subsets of devices with an organization to have their own mesh network.
* Allowing for security-policy for traffic within an Apex managed network.
```

If we have existing issues in the backlog for those features, it'd be good to cross-reference them here, too.

russellb (changes requested):
Thanks for writing this up. I am generally supportive of the v2 iteration as described here. My comments are relatively minor.

docs/design/database.md
dave-tucker: I've added a note about this in the text that the functionality won't be implemented in v1->v2 but that the schema supports such a feature. It wouldn't require changes so we wouldn't be at v2.5.

docs/design/database.md
dave-tucker: Ack. Fixing now.

docs/design/database.md
dave-tucker: I removed this entirely. We can open an issue and another PR to track/discuss v3 changes.

docs/design/database.md
```diff
@@ -0,0 +1,127 @@
...
+1. User runs the Apex agent
+1. Apex agent does a `POST /device` to register the unique device with Apex for the user. It is assigned into the single zone that a user is in (either the default zone, or a new zone where the user was moved to by an admin).
+1. Once the device is registered a `POST /zones/$id/peers` was issued to create an association between this device and a zone.
+1. `GET /zones/$id/peers` was polled to retrieve the peer listing for a zone.
+
+On top of this flow, we were somewhat abusing `POST /zones/$id/peers` to also update peers after NAT-T.
+
+This model was built under the assumption that multi-tenancy would be achieved by deploying an Apex stack per-tenant.
+
+## v2: SaaS Preview Milestone
```
russellb, line 68:
```suggestion
## v2: SaaS Preview Milestone

See [issue #348](https://github.com/redhat-et/apex/issues/348).
```

russellb (approved):
lgtm.

My only other suggestion is to include a link to the issue tracking the v2 work so someone could go find the code associated with it if they were coming from this doc.

russellb:
I'll hold off on merging in case anyone else wants to weigh in on this.

nerdalert:
LGTM, next step to brainstorm is the slicing of the org. VPCification. 🎉

nerdalert (approved):
LGTM, next step to brainstorm is the slicing of the org. VPCification. 🎉

docs/design/database.md
```diff
@@ -0,0 +1,129 @@
...
+        string cidr
+        string hub_zone
+    }
+```
+
+In this simplified model we only have 3 concepts:
+
+- Users: Unique users of the service
+- Devices: A unique end-user device
+- Organization: A collection of users who have permissions to onboard one or more devices into this organization
```
vishnoianil, line 115: I think we need some clarity around users <-> organization relationship. Looks like we support both the models N:1 and 1:N for user and organization. 1:N relation is clear. But needs some clarification around N:1 model. 

- Every user will have it's own personal organization, and given that each organization can have multiple users, I assume that the original owner of the organization need to invite other users to his organization? Will that invited user have similar kind of ownership on the invited organization (e.g can this invited user invite other users)? or we will have a concept of original owner and invitees, where invitee only have "read" only access? I assume it's the later, but want to clarify.
- In v2 we are not allowing users to join multiple organization, but do we allow user to join any other organization (1:1) ?

dave-tucker: The diagram shows that a user may belong to 1 or more organizations - the first organization they belong to is their personal org.

> Every user will have it's own personal organization, and given that each organization can have multiple users, I assume that the original owner of the organization need to invite other users to his organization? Will that invited user have similar kind of ownership on the invited organization (e.g can this invited user invite other users)? or we will have a concept of original owner and invitees, where invitee only have "read" only access? I assume it's the later, but want to clarify.

The invite mechanism hasn't been designed yet - and is somewhat tied in with the ACL work we need to do across the API... but your assumption is correct. We'd have different permission levels in an org and the default would be read-only for everyone other than the owner.

> In v2 we are not allowing users to join multiple organization, but do we allow user to join any other organization (1:1) ?

I think this is backwards. V1 was 1:1 User:Zone. V2 is 1:Many(1 or more) User:Organization.
V2 can support joining multiple orgs, even if it takes a while for us to implement it. The main thing is the database layout wouldn't need to change

vishnoianil: Sounds good.

vishnoianil (approved):

PR #397, opened 2023-02-10
feat: change GetPubIPv4 to give the IP:port endpoint address of the wg port.

chirino:


nerdalert (approved):
LGTM! figuring out how to be driver independent for NAT-T will be great. We essentially ignore the initial STUN once nodes start reconciling from the WG hub-router/discovery node. We can STUN initially but once the port is bound we can't source from the WG port anymore for reconciling middle box/state changes. @vishnoianil is labbing some stuff here as well. Pushing a proposal doc with a skeleton for disco in a few.

chirino:
@nerdalert you might notice that the PR using github.com/libp2p/go-reuseport to create the UDP connection, instead of the net package.  I was hoping that a some point we could test out see IF that allows WG to also open up the UDP port AND what happens if both WG and STUN packets are sent on the same port.  I figure if WG opens it's port with right right reuse options, it should just drop all the STUN packets since they are not WG packets.  Not too sure how well the stun client impl would deal with receiving WG packets.

vishnoianil:
This could have been helpful for the symmetric NAT hole punching scenario (if we can use the wireguard port). Unfortunately wireguard doesn't open the socket with SO_REUSEPORT (https://git.zx2c4.com/wireguard-linux/tree/drivers/net/wireguard/socket.c), so once wireguard tunnel is setup, we won't be able to send the stun packet through the same port. We can go otherway around and write a stun client that bind to wireguard port with SO_REUSEPORT, and assuming wireguard will be able to bind to that as well, but I think bombardment of wireguard packet to stun server can create a challenge for it and make it useless. We can use ebpf magic to set SO_REUSEPORT for the wireguard socket, that option is open :). 

The other alternative is to use stun client to open a random port and then configure the wireguard to listen on that port. 

Irrespective of what approach we use, the port mapping at NAT (wireguard port : external port) can change due to various reasons. There is no decent approach to detect it, apart from running explicit ping from the apex agent to external node (stun, bootnode), and detecting the failure. If we can determine a decent approach to detect the mapping change, we can use any of the above approach to work (without involving the hub discovery).

PR #398, opened 2023-02-10
fix a CI warning: The `set-output` command is deprecated 

chirino:
and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -28,9 +28,10 @@ jobs:
           go-version: 1.19
 
       - id: go-cache-paths
+        shell: bash
         run: |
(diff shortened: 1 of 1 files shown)


vishnoianil:
@chirino can you please signoff the pr, DCO failing.

vishnoianil:
Looks like windows powershell (default shell for windows runner in github) is not happy with the >>$GITHUB_OUTPUT, I think we need to either enforce `shell: bash` or write separate step for windows to use powershell friendly command `$env:GITHUB_OUTPUT`

russellb:
windows builds passed now

PR #419, opened 2023-02-14
Fix paths for the docs workflow job

vishnoianil:

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -4,11 +4,11 @@ on:
   push:
     branches: ["main"]
     paths:
-      - "**/*.md"
+      - '**/*.md'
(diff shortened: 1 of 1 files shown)


.github/workflows/docs.yml
```diff
@@ -4,11 +4,11 @@ on:
   push:
     branches: ["main"]
     paths:
-      - "**/*.md"
+      - '**/*.md'
```
russellb, line 7: Double quotes are used in build.yml and that seems to work

vishnoianil: yeah, that's what i was thinking, but I am not sure why it's not triggering the markdown-lint job for https://github.com/redhat-et/apex/pull/416

PR #422, opened 2023-02-15
clean up: reduce code duplication in stun.go, use a dynamic port the wg interface that we can open.

chirino:
Using a random high port for the wg is not guaranteed to not be in use by another process, so lets ask the OS for a dynamic port instead.

This now avoids an additional stun request once we have determined that the device is behind a symmetric nat.

Wait for apex to be deployed when running `make recreate-db`

### DIFF
--- a/internal/apex/apex.go
+++ b/internal/apex/apex.go
@@ -199,13 +199,16 @@ func (ax *Apex) Start(ctx context.Context, wg *sync.WaitGroup) error {
 		localEndpointIP = ax.userProvidedEndpointIP
 		localEndpointPort = ax.listenPort
 	}
-	if ax.stun && localEndpointIP == "" {
-		ipPort, err := GetPublicUDPAddr(ax.logger, ax.listenPort)
+
+	// If we are behind a symmetricNat, the endpoint ip discovered by a stun server is useless
+	if !ax.symmetricNat && ax.stun && localEndpointIP == "" {
+		ipPort, err := StunRequest(ax.logger, stunServer, ax.listenPort)
 		if err != nil {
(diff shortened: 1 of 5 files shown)


internal/apex/apex.go
```diff
@@ -487,27 +490,19 @@ func (ax *Apex) nodePrep() {
...
+		ax.nodeReflexiveAddress = stunAddr.IP.String()
 	}
 
-	isSymmetric, err := IsSymmetricNAT(ax.logger)
+	isSymmetric := false
+	stunAddr2, err := StunRequest(ax.logger, stunServer2, ax.listenPort)
 	if err != nil {
 		log.Error(err)
+	} else {
+		isSymmetric = stunAddr.String() == stunAddr2.String()
```
nerdalert, line 505:
```suggestion
		isSymmetric = stunAddr.String() != stunAddr2.String()
```

nerdalert (changes requested):
Looks good @chirino, ty! Just one issue is flipping the stun results evaluation. If the results don't match we infer it is behind a symmetric NAT cone somewhere in the path.

internal/apex/stun.go
```diff
@@ -3,114 +3,29 @@ package apex
...
-	"strconv"
-	"time"
-
 	"github.com/pion/stun"
 	"go.uber.org/zap"
+	"net"
 )
 
 const (
+	stunServer  = "stun.l.google.com:19302"
```
vishnoianil, line 12: if we use stunServer1 in utils.go, we can probably get rid of one of the stun server here?

internal/apex/apex.go
chirino: doh.. good catch.  thx.

PR #426, opened 2023-02-16
Remove redundant stun server

vishnoianil:


russellb:
What makes them redundant? They all resolve to different things for me.

vishnoianil:
> What makes them redundant? They all resolve to different things for me.

At max we need two stun server addresses to determine whether device is behind symmetric NAT or not. For rest we can rely on any of these two stun server, so i think we don't need third stun address, until and unless i am missing something here .

russellb:
Ok. “Redundant” sounded like one was a duplicate. 

I’d have to review the logic. Having extras to try in case one is down would be helpful. It’s probably not implemented that way yet though.

russellb (approved):

PR #427, opened 2023-02-16
Add .vscode directory to paths-ignore for build and doc job

vishnoianil:
So that *.md only patches with any change in .vscode directory won't trigger build jobs.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,11 +6,13 @@ on:
     paths-ignore:
       - '**/*.md'
       - '**/*.gitignore'
+      - '.vscode/**'
   pull_request:
     branches: ["main"]
     paths-ignore:
       - '**/*.md'
       - '**/*.gitignore'
+      - '.vscode/**'
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -5,10 +5,12 @@ on:
     branches: ["main"]
     paths:
       - '**/*.md'
+      - '.vscode/**'
   pull_request:
     branches: ["main"]
     paths:
       - '**/*.md'
+      - '.vscode/**'
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}


russellb (changes requested):
If you add it to gitignore, you should remove the file from the tree, too.

vishnoianil:
> If you add it to gitignore, you should remove the file from the tree, too.

Yeah, I was thinking of removing the file as well, but it might remove the local .vscode/settings.json from the local dev environment as well, whenever anyone will pull the main? so just out of caution i kept it, but happy to toss it if that's okay.

russellb:
I don’t think it will remove it if you have any local changes to the file. 

What’s the harm in keeping it, though? On slack you said you didn’t want it triggering build jobs, right? You could just ignore it in that config?

vishnoianil:
> I don’t think it will remove it if you have any local changes to the file.
> 
> What’s the harm in keeping it, though? On slack you said you didn’t want it triggering build jobs, right? You could just ignore it in that config?

Yeah, we can do that as well. (Also i prefer setting.json to be in the repo, so you don't have to deal with those annoying dictionary warning).

russellb:
> > I don’t think it will remove it if you have any local changes to the file.
> > What’s the harm in keeping it, though? On slack you said you didn’t want it triggering build jobs, right? You could just ignore it in that config?
> 
> Yeah, we can do that as well. (Also i prefer setting.json to be in the repo, so you don't have to deal with those annoying dictionary warning).

I use vscode, too. I appreciate your updates. Let's leave it and just fix CI.

vishnoianil:
> > > I don’t think it will remove it if you have any local changes to the file.
> > > What’s the harm in keeping it, though? On slack you said you didn’t want it triggering build jobs, right? You could just ignore it in that config?
> > 
> > 
> > Yeah, we can do that as well. (Also i prefer setting.json to be in the repo, so you don't have to deal with those annoying dictionary warning).
> 
> I use vscode, too. I appreciate your updates. Let's leave it and just fix CI.

Done

vishnoianil:
@russellb Integration test is failing, but rest of the jobs are passing, so probably for this PR, we don't have to wait for e2e to pass?

russellb (approved):

PR #429, opened 2023-02-16
Add rpm and systemd integration for apex

russellb:
Add rpm packaging and systemd integration for the apex daemon.

With this change, you can run `make rpm`, and then install the resulting package on a Fedora host. You can then edit `/etc/sysconfig/apex` to set the service URL and then start the daemon with `systemctl start apex`.

Update the agent docs to include how to build and install the rpm, as well as install and use the systemd integration.

Split out from #341


dave-tucker (approved):
LGTM!

PR #433, opened 2023-02-17
lint: Ignore ui/node_modules for markdown-lint

### LINKED ISSUE #432: Markdown lint is too greedy!
### Describe the bug

I ran `make all` and was overrun with markdownlint warnings.

### How to Reproduce

Do some development on the UI, which populates `./ui/node_modules` and then the lint will never pass.

### Expected Behavior

The lint should only check docs that are under our control (i.e not paths that are .gitignored).

### Deployment Environment

_No response_

### Additional Context

_No response_

russellb:
Closes #432


dave-tucker (approved):
LGTM

PR #440, opened 2023-02-18
design: Add a project name proposal.

russellb:
The project's github repo should move. We will need a new github organization to host the project. The `apex` org is taken, so one option is to create a github org with some variable of `apex` like `apex-mesh` or `apex-hosted`. The other option is to rename the whole project to something more unique.

This PR proposes a full project rename.

Issue #428

PR #442, opened 2023-02-18
Apex component naming proposal

### LINKED ISSUE #441: Consistent naming for apex stack components
### Describe the Problem Statement

Currently we are are using multiple names for different components across our documentation and code, such as

* Apex Stack is referred as Apex Service, Apex ApiServer, Apex Controller, Apex Control Plane, Apex SaaS
* Apex Agent is referred as hub-router, relay, agent, discovery node

It would be really helpful to write a coherent documentation and code if we can zero down on the naming of these components.

vishnoianil:
Initial proposal to start the discussion around finding consistent naming of the apex components, so that we can write coherent document/code

Resolves #441

### DIFF
--- a/docs/proposals/naming.md
+++ b/docs/proposals/naming.md
@@ -1,0 +1,44 @@
+# Enhancement Title
+
+> [Issue #441](https://github.com/redhat-et/issues/441)
+
+## Summary
+
+Proposal for using consistent naming for the major components of the Apex stack
+
+## Proposal
+
+Currently we are are using multiple names for different components across our documentation and code, such as
+
+* Apex Stack is referred as Apex Service, Apex ApiServer, Apex Controller, Apex Control Plane, Apex SaaS
+* Apex Agent is referred as hub-router, relay, agent, discovery node
+
+It would be really helpful to write a coherent documentation and code if we can zero down on the naming of these components.
+
+Here is my initial proposal to start the discussion
+
+* Apex Service: It's an umbrella term to refer to the entire stack, that contains following individual components (Named as is).
+  * Apex ApiServer
+  * Apex Frontend
+  * Apex Database
+  * Apex Ipam
+  * Apex ApiProxy
+  * Apex backend-cli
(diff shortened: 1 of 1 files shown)


dave-tucker:
Can we combine this with #440 and perhaps have a `docs/naming.md`

docs/proposals/naming.md
```diff
@@ -0,0 +1,44 @@
...
+  * Apex backend-cli
+  * Apex backend-web
+* Apex Agent: Agent running on each node, just doing Agent's job.
+  * Binary Naming: We need to go one more level down to clarify the binary naming as we are planning to separate the relay functionality out of agent.
+    * apexd: Agent binary (running as a daemon) that does the magic of connectivity.
+    * apex: A command line utility to do read/write operation through Apexd
+* Apex ICE: A node running the discovery and relay function.
+  * Binary Naming:
+    * apexiced: Binary (running as a daemon) that does the magic of discovery and relay.
+    * apexice: A command line utility to do read/write operation through Apexiced
```
chirino, line 35: ICE is the protocol the client uses to discover, prioritize, and test the endpoint candidate addresses of peers.  I don't really like it to describe a node that would be running stun and relay functionality.  I would rather call it something like relay.

russellb:
> Can we combine this with #440 and perhaps have a `docs/naming.md`

@vishnoianil I will incorporate this into my PR if you're good with it.

vishnoianil:
> > Can we combine this with #440 and perhaps have a `docs/naming.md`
> 
> @vishnoianil I will incorporate this into my PR if you're good with it.

Yup, sounds good.

docs/proposals/naming.md
vishnoianil: yeah makes sense. Relay was an option i was also thinking, but again it's kind of restrict to one functionality, and as of now it's doing more than that. Let's continue discussion on #440

vishnoianil:
Proposal merged with #440

PR #457, opened 2023-02-20
fix: better support arm64 developer machines

chirino:
The crunchydb operator does not work well on arm64, so this implements a simpler deployment of the postgres server.


russellb (approved):
I like the overlay approach here. I'm good with this going on once CI is happy (other than known e2e failures).

PR #464, opened 2023-02-20
Publish RPMs to a public repo after commits to main

### LINKED ISSUE #436: Check rpm builds in CI
PR #429 added rpm builds, but they're not checked in CI to avoid regression. We need a github action that runs the rpm build.

russellb:
```
commit 2d944741186134f7b77be3eebecab4f30804df37
Date:   Mon Feb 20 20:49:36 2023 -0500

    rpm: Package the manpages
    
commit 629050c550e1809b8df20cfcf3ec678e4d88d69a
Date:   Mon Feb 20 18:02:41 2023 -0500

    Makefile: Add manpage generation
    
commit cf644df5a14467d920128948c1274803e29c6964
Date:   Mon Feb 20 17:11:44 2023 -0500

    docs: Document how to use the copr repo for rpms
    
commit 33e8f3dbf5b1f9b78a1f765e0f59575565ff2faf
Date:   Mon Feb 20 14:20:33 2023 -0500

    ci: Publish rpms to a copr repo
    
    This PR sets up a job to publish new RPMs to a copr repo for each new
    commit to the main branch
    
commit 96e59dd36dcb333d28a3d7d750e0f50eefd9c08f
Date:   Fri Feb 17 12:42:58 2023 -0500

    ci: Add an rpm build job
    
    Ensure the rpm build is functional via CI. --cap-add SYS_ADMIN didn't
    work in CI for some reason, so I switched to a privileged container. I
    also made the build a little less noisy.
    
    Closes #436
    
commit 2b3df2803fcf4736043eff2112f55623cc8ebed7
Date:   Mon Feb 20 11:26:05 2023 -0500

    ci: Split unit tests out into their own job
    
    I was surprised to see the unit tests running in the lint job. This
    patch splits them out into their own job that runs after lint passes.
    
    This change also renames `lint` to `go-lint` to make the job name more
    explicit about the type of linting it is doing.
    
commit 48ca3d6858087e1664f4d50aa413ed6d189dc957
Date:   Mon Feb 20 11:23:32 2023 -0500

    ci: Stop running go fmt
    
    This command silently fixes any formatting issues before running the
    linter. We should let the linter find the issues and then fail the CI
    job.

```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -369,3 +369,16 @@ jobs:
       - name: copy binaries to s3
         run: |
           aws s3 sync . s3://apex-net/
+
+  build-rpm:
+    needs: ["go-lint"]
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: docker/setup-buildx-action@v2
+      - name: Install make
+        run: |
+          sudo apt -y install make
+      - name: Build rpm
+        run: |
+          make rpm
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -375,7 +375,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - uses: docker/setup-buildx-action@v2
       - name: Install make
         run: |
           sudo apt -y install make
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,7 +19,45 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  lint:
+  go-lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Set up Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: 1.19
+
+      - id: go-cache-paths
+        shell: bash
+        run: |
+          echo "go-build=$(go env GOCACHE)" >> $GITHUB_OUTPUT
+          echo "go-mod=$(go env GOMODCACHE)" >> $GITHUB_OUTPUT
+
+      - name: Go Build Cache
+        uses: actions/cache@v3
+        with:
+          path: ${{ steps.go-cache-paths.outputs.go-build }}
+          key: ${{ runner.os }}-go-build-${{ hashFiles('**/go.sum') }}
+
+      - name: Go Mod Cache
+        uses: actions/cache@v3
+        with:
+          path: ${{ steps.go-cache-paths.outputs.go-mod }}
+          key: ${{ runner.os }}-go-mod-${{ hashFiles('**/go.sum') }}
+
+      - name: Build
+        run: |
+          go build ./...
+
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v3
+        with:
+          version: v1.50.0
+
+  go-unit:
+    needs: go-lint
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
@@ -54,11 +92,6 @@ jobs:
       - name: Unit tests
         run: |
           go test -v ./...
-
-      - name: golangci-lint
-        uses: golangci/golangci-lint-action@v3
-        with:
-          version: v1.50.0
 
   ui-lint:
     runs-on: ubuntu-latest
@@ -111,7 +144,7 @@ jobs:
           kubeconform -summary -output json -schema-location default -schema-location 'https://raw.githubusercontent.com/datreeio/CRDs-catalog/main/{{.Group}}/{{.ResourceKind}}_{{.ResourceAPIVersion}}.json' -schema-location 'deploy/.crdSchemas/{{.ResourceKind}}_{{.ResourceAPIVersion}}.json' kubeconfigs/
 
   build-images:
-    needs: lint
+    needs: go-lint
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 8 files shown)


.github/workflows/build.yml
```diff
@@ -340,3 +369,44 @@ jobs:
...
+      - uses: actions/checkout@v2
+      - uses: docker/setup-buildx-action@v2
+      - name: Install make
+        run: |
+          sudo apt -y install make
+      - name: Build rpm
+        run: |
+          make rpm
+
+  publish-rpms:
```
dave-tucker, line 386: This should live in the `deploy` workflow. You can then get rid of ` if: github.ref == 'refs/heads/main'`

.github/workflows/build.yml
```diff
@@ -340,3 +369,44 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: docker/setup-buildx-action@v2
+      - name: Install make
+        run: |
+          sudo apt -y install make
+      - name: Build rpm
+        run: |
+          make rpm
```
dave-tucker, line 384: Is it also worth adding the RPMs as artifacts to the CI job? That way if we could grab RPMs to test specific PRs which might be cool

.github/workflows/build.yml
```diff
@@ -340,3 +369,44 @@ jobs:
       - name: copy binaries to s3
         run: |
           aws s3 sync . s3://apex-net/
+
+  build-rpm:
+    needs: ["go-lint"]
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: docker/setup-buildx-action@v2
```
dave-tucker, line 378: Is buildx used in this job?

.github/workflows/build.yml
```diff
@@ -340,3 +369,44 @@ jobs:
...
+      matrix:
+        include:
+          - mock_root: fedora-37-x86_64
+            srpm_distro: fc37
+          - mock_root: fedora-38-x86_64
+            srpm_distro: fc38
+    if: github.ref == 'refs/heads/main'
+    steps:
+      - uses: actions/checkout@v2
+      - uses: docker/setup-buildx-action@v2
```
dave-tucker, line 400: Same comment here re: buildx

.github/workflows/build.yml
```diff
@@ -340,3 +369,44 @@ jobs:
...
+      - name: Install make
+        run: |
+          sudo apt -y install make
+      - name: Build srpm
+        run: |
+          MOCK_ROOT="${{ matrix.mock_root }}" make srpm
```
dave-tucker, line 406: It's been ages since I last build RPMs... but aren't there some directories that we might want/need to cache between builds to keep build times low.

dave-tucker (changes requested):

.github/workflows/build.yml
russellb: OK. We should probably move publishing binaries to S3 there, as well.  None of these are really `deploy`, though ... maybe a new `publish` workflow?

.github/workflows/build.yml
russellb: Yeah, that's a good idea. I'll add it.

For building the full rpm, I only test building it once (for Fedora 37), but I have copr do multiple builds for different distros/versions (Just F37 and F38 so far).

So we'd just have the F37 rpm in artifacts, but source rpms for every distro we publish for on copr.

.github/workflows/build.yml
russellb: docker build is used, at least. I copied it from other jobs that were doing some image builds. What I really want is "please set up docker so I can `docker build` and `docker run` in later steps."

.github/workflows/build.yml
russellb: Every source RPM is built in a new chroot, because each one is a different distro/version.

The only case I know of that could benefit is the `build-rpm` job because it's building the source rpm and then a binary rpm using the same chroot type. I could reuse it instead of having it rebuilt. It doesn't take that long, and doing everything in a clean chroot is probably better.  I'll take another close look at the times to see if it's worth it.

russellb:
Updated this to generate manpages and include them in the rpm.

.github/workflows/build.yml
russellb: ... ah, but uploading to s3 depends on artifacts saved off in a past job in the workflow, so I guess that one can't move.

.github/workflows/build.yml
dave-tucker: You don't need that step then since both of those are available on the runner itself.

dave-tucker:
@russellb looks like there's a bad target reference somewhere in the Makefile in the last push - failed to find target for `dist/apex`

russellb:
> @russellb looks like there's a bad target reference somewhere in the Makefile in the last push - failed to find target for `dist/apex`

yeah, my bad. I'm about to post a fix along with updates based on your review.

.github/workflows/build.yml
russellb: done

.github/workflows/build.yml
russellb: fixed

.github/workflows/build.yml
russellb: So far I sped things up about 10% by just telling it not to do any cleaning of the chroot before and after the build. Each run of `mock` happens in a new container anyway.

I know there's more I can do, so I'll keep poking at it over time.

.github/workflows/build.yml
russellb: rpms from `build-rpm` stored as artifacts now

russellb:
@dave-tucker I think this is ready other than doing more to cache things and speed up builds. No other jobs depend on the rpm jobs, though.

dave-tucker (approved):
LGTM! We can iterate on this later.

PR #467, opened 2023-02-21
Use golang-1.19 image, instead of golang-1.19-alpine

vishnoianil:
Apex agent container image build is failing for windows platform, because the docker container 
file is using golang-1.19-alpine image to build the  binary, and obviously it's doesn't support windows/amd64-arm64
arch.


russellb:
I went to review this, but I'm missing context. Could you update the commit message and PR summary to describe the problem it fixes?

vishnoianil:
> I went to review this, but I'm missing context. Could you update the commit message and PR summary to describe the problem it fixes?

Done, let me know if it's helpful.

PR #469, opened 2023-02-22
Revert "Apex agent container image build is failing for"

russellb:
This reverts commit 85a479cac8ff4562273950d6d7cedb7b967a2e73.

Switch back to the alpine-based image, as our jobs assume alpine tools for installing things. A later commit disabling windows builds on linux workers was the fix needed.


russellb:
See this for the latest failure: https://github.com/redhat-et/apex/actions/runs/4245432405/jobs/7380928724

russellb:
Anil wants to push a different fix

PR #471, opened 2023-02-22
Faster make deploy and recreate db

chirino:
This also pegs the build to a specific revision of crunchy db and makes the test container multi-arch.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -109,8 +109,9 @@ test: ## Run unit tests
 
 APEX_LOCAL_IP:=`go run ./hack/resolve-ip apex.local`
 .PHONY: run-test-container
-run-test-container: dist/apexd dist/apexctl test-images ## Run docker container that you can run apex in
-	docker run --rm -it --network bridge \
+run-test-container: ## Run docker container that you can run apex in
+	@docker build -f Containerfile.test -t quay.io/apex/test:ubuntu --target ubuntu .
+	@docker run --rm -it --network bridge \
 		--cap-add SYS_MODULE \
 		--cap-add NET_ADMIN \
 		--cap-add NET_RAW \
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -109,8 +109,9 @@ test: ## Run unit tests
 
 APEX_LOCAL_IP:=`go run ./hack/resolve-ip apex.local`
 .PHONY: run-test-container
-run-test-container: dist/apexd dist/apexctl test-images ## Run docker container that you can run apex in
-	docker run --rm -it --network bridge \
+run-test-container: ## Run docker container that you can run apex in
+	@docker build -f Containerfile.test -t quay.io/apex/test:ubuntu --target ubuntu .
```
vishnoianil, line 113: @chirino given that `run-test-container` target is using ubuntu image, I think it would be better to move ubuntu builder in Container.test before the alpine image builder, so it won't build alpine/fedora every time `run-test-container` runs?

chirino: I don't see it trying to build the other targets.  Seems docker is smart enough to skip stuff that's not needed to build the target

chirino:
@vishnoianil hey could your review and merge this?  I think I’m happy with it now.

Makefile
vishnoianil: I ran the following command locally (I tossed all the pre-built fedora/alpine/ubuntu) and I can see that all the images are built, but once you build the images, the second run is pretty quick because it start using the caches, so not a big issue as such.

```
#docker build -f Containerfile.test -t quay.io/apex/test:ubuntu --target ubuntu .
Sending build context to Docker daemon  143.7MB
Step 1/33 : FROM docker.io/library/golang:1.19-alpine as build
 ---> feb4bbda921c
Step 9/33 : FROM alpine:3.16 as alpine
 ---> dfd21b5a31f5
Step 17/33 : FROM fedora:36 as fedora
 ---> 0e68c6a5f825
Step 25/33 : FROM ubuntu:22.04 as ubuntu
 ---> 58db3edaf2be

Successfully built e3755b4724ea
Successfully tagged quay.io/apex/test:ubuntu
```

vishnoianil (approved):

PR #472, opened 2023-02-23
Update the UX with new schema

nerdalert:
- Fixes links and adds some fields to devices.


russellb:
a minor thing, but can you change "UX" to "UI" in the commit message?

ui/src/pages/Devices.tsx
```diff
@@ -43,6 +43,15 @@ export const DeviceList = () => (
...
       <TextField label="Hostname" source="hostname" />
+      <TextField label="Allowed IPs" source="allowed_ips" />
+      <TextField label="Tunnel IP" source="tunnel_ip" />
+      <TextField label="Local IP" source="local_ip" />
+      <TextField label="Organization Prefix" source="organization_prefix" />
+      <TextField label="Reflexive IPv4" source="reflexive_ipv4" />
+      <TextField
+        label="Endpoint Local IPv4"
+        source="endpoint_local_address_ipv4"
+      />
```
russellb, line 54: This looks like too much detail for the list of all devices. I would leave most of this for the detailed view for when you click on a device to see all of its details. That is `DeviceShow` down below.

ui/src/App.tsx
```diff
@@ -63,15 +61,6 @@ const App = () => {
...
         recordRepresentation={(record) => `${record.hostname}`}
       />
-      <Resource
-        name="zones"
-        list={ZoneList}
-        show={ZoneShow}
-        create={ZoneCreate}
-        icon={ZoneIcon}
-        recordRepresentation={(record) => `${record.name}`}
-      />
```
russellb, line 73: What about adding Organizations?

russellb (changes requested):

ui/src/pages/Devices.tsx
nerdalert: Going to push back on that a bit. I personally don't like having to drill down to get basic information. EC2 has 15 columns. I trimmed this down some. I took out allowed IPs but a routes column might make sense to go back in eventually.

ui/src/App.tsx
nerdalert: Was going to do in a separate PR but added to this.

russellb (changes requested):
While trying this out, I spotted "Zone" still on the `Users` list page and `User` detail page.

Otherwise, lgtm.

PR #473, opened 2023-02-24
feat: mount the .certs directory into the test container image

chirino:
instead of building the certs into it.

This will allow the same image usable with any install of the apex service.


chirino:
not yet ready...

PR #476, opened 2023-02-25
Proposed Changes to Apex Deployment manifests for Op1st to ROSA handover

Gregory-Pereira:
With the Deployment of APEX moving to ROSA, I have been messing around with the [Deployment](https://console-openshift-console.apps.open-svc-sts.k1wl.p1.openshiftapps.com/k8s/ns/apex/core~v1~Pod) to get it production ready. With these changes, every `Pod` , `Deployment`,  `Certificate`, `Ingress` and `ReplicaSet` are working to the best of my knowledge, although I would love someone to validate this at their earliest convenience.

List of changes:
- apiproxy `caddy` container did not haver permissions to run as it was running as a different user in Smaug compared to rosa. I am not certain this is the correct user (1001930000) to run the `/usr/bin/caddy` binary but it seems to work now.
- `Imagestream`s have been implemented to pull images into the openshift `image-registry` rather than using them directly, which resolved the`imagePullBackOff` issues for the `ipam` and `apiserver` pods.
- As discussed DNS records have been updated and moved into the ROSA `route53`. As such ingresses needed to be updated to reflect this (`apex-hosted.cloud`).
- After the certificates were correctly created, they each generate a secret. I created 3 sealed secrets as these should live in the `apex` Argocd `Application` on ROSA rather than be deployed as cluster-resources in our gitops repo.
    - These secrets belong in an overlay as they are not base resources and someone might be confused if the deployment ever moves again, as these sealed secrets contain data specific to the ROSA cluster because of how the sealed-secrets controller works. As such I created the ROSA directory in the production overlay and moved everything in there.

/cc @ryanthecook
/cc @dave-tucker 
/cc @russellb 
/cc @chirino


Gregory-Pereira:
Are openshift-specific manifests not allowed in this repository? Seems like the repo you are checking aginst for CRDs is not set to also check the `openshift` branch where the `kind: Imagestream` CRD is available ([see openshift standalone definitions: https://raw.githubusercontent.com/datreeio/CRDs-catalog/openshift/openshift/v4.8.2-standalone/_definitions.json)]. This seems like a mistake as overlays should contain resources that are specific to the deployment environment.

chirino:
Add the missing CRDs to `./deploy/.crdSchemas` so we can validate the  YAML against the CRD. 
You can generate those schemas  running [this script](https://github.com/datreeio/CRDs-catalog#usage) against your cluster and  copy them from `~/.datree/crdSchemas` to `./deploy/.crdSchemas`.

The dev overlay is run against a kind cluster, so don't include any openshift CRDs in the yamls that pulls in.

vishnoianil:
@Gregory-Pereira I pushed PR #498, that successfully deploys both Production and QA Apex deployment. Thanks for all the great work in migrating the overlays to ROSA. I used some of the changes from this PR and fixed all the remaining issues that was causing the failure in the deployment. I added you as a co-author to the PR and commit to #498.

Please take a look at that PR and if you can attempt to deploy it through your environment to test it that would be great.

russellb:
These changes were included in #498. Thank you for all of your help!

PR #493, opened 2023-02-27
deploy: Check for new commits to main

russellb:
The deploy workflow fails often if additional PRs have been merged by the time this job runs. This change will check to see if there are any commits in main that are not in the branch the job is running on. If so, it will not attempt to push a commit that would fail.


vishnoianil:
@russellb ui-lint job failing because of the npm errors.

russellb:
> @russellb ui-lint job failing because of the npm errors.

Working now

vishnoianil (approved):

PR #494, opened 2023-02-27
Change the default org prefix to CGNAT

nerdalert:
- Updated addressing in docs and removed some zone mentions as well. This is not intended to fix all of the zone removal docs. Once we have org user moves we can add back those intructions.


nerdalert:
Also threw in a log change. Just to get rid of the stack trace for every poll. It's relatively temporary but better than current.

vishnoianil (approved):

vishnoianil (approved):

PR #495, opened 2023-02-28
fix: close the stun client

chirino:


russellb (approved):
It'd be good to have some more detail in the commit message to explain what the impact of this bug was, but the fix LGTM. I'll merge once CI passes

PR #499, opened 2023-03-02
Add a passwd prompt option so credentials are not displayed

nerdalert:


nerdalert:
The backend is identical to the initial username/password onboard, just adds the option to obscure the passwd.

```
# 1 no prompts
apexd --username kitteh1 --password floofykittens https://apex.local

# 2 passwd prompt
apexd --username kitteh1  https://apex.local

# 3 one-time pass
apexd https://apex.local
```

PR #519, opened 2023-03-03
Remove remaining apex-net s3 bucket references from the code

vishnoianil:
Fix binaries artifacts name in build job and scripts 
Fix namespace name in makefile for run-sql-apiserver target
Fix apexd references in ansible vars/readme


vishnoianil:
@russellb @nerdalert YOLOing this to trigger the deploy job with latest code and debug the failures.

Issue #437, opened 2023-02-17
Outstanding e2e fixes


nerdalert:
### Describe the bug

TODOs to resolve the issues:

- [x]  Support for overlapping cidr tests once IPAM namespacing is resolved
- [x]  Re-enable support for `--request-ip` in orgs
- [x] child-prefixes are not being added to the DB on a rejoin (this is probably the case for any other modifications after a node has been joined.
- [x]  Re-enable device delete
- [x] Apexctl test

### How to Reproduce

make e2e

### Expected Behavior

passed tests

PR #523, opened 2023-03-03
Fix IPAM pool releasing and the related e2e

### LINKED ISSUE #522: IPAM not releasing addresses on device delete
### Describe the bug

IPAM not releasing addresses on device delete

### How to Reproduce

Join a node and delete it

### Expected Behavior

to delete the IPAM lease

### Deployment Environment

_No response_

### Additional Context

_No response_

nerdalert:
- Resolves #522
- Re-enables nexctl e2e #437

### DIFF
--- a/integration-tests/integration_test.go
+++ b/integration-tests/integration_test.go
@@ -827,7 +827,6 @@ func (suite *NexodusIntegrationSuite) TestRelayNAT() {
 }
 
 func (suite *NexodusIntegrationSuite) Testnexctl() {
-	suite.T().Skip("Skipping since deleting org fails with key constraint errors")
 	assert := suite.Assert()
 	require := suite.Require()
 	parentCtx := context.Background()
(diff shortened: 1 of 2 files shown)


nerdalert:
> Can you update the commit message with a bit more info on how this addresses the bug? I see in the bug that "IPAM not releasing addresses on device delete" but then the change just adds a new error case, so the connection from the change to the bug isn't obvious to me.

The change looks up the device in the DB to know what address to release. It was trying to release empty values before. Not sure what you mean on just changes the error return. This lookup is new.

https://github.com/nexodus-io/nexodus/pull/523/files#diff-604fecaefd8072119f139e076c9370b719c56a8395bc323cf4f391149052de78R334

```
if res := api.db.First(&device, "id = ?", deviceID); res.Error != nil {
		c.JSON(http.StatusBadRequest, models.NewApiInternalError(res.Error))
		return
	}
```

russellb:
Thanks for the explanantion. I see it now.

PR #524, opened 2023-03-03
ui: Make logo fit inside the card

russellb:
The wordmark logo was not properly resizing inside of the card it resides in. In my case, I was only seeing part of it.

Also replace the logo in the top app bar.

image


russellb:
image

vishnoianil (approved):

PR #525, opened 2023-03-03
Updated deploy yamls with following changes

vishnoianil:
> Use NEXAPI prefix for env var of apiserver
> User subdomain (try.nexodus.io) as a frontend url

Also,

> Fix integration tests with new urls
> Updated references  nexodus.local to try.nexodus.local


vishnoianil:
FYI..@Gregory-Pereira

.github/workflows/build.yml
```diff
@@ -324,7 +324,7 @@ jobs:
 
       - name: Add hosts to /etc/hosts
         run: |
-          echo "127.0.0.1 auth.nexodus.local api.nexodus.local nexodus.local" | sudo tee -a /etc/hosts
+          echo "127.0.0.1 auth.try.nexodus.local api.try.nexodus.local try.nexodus.local" | sudo tee -a /etc/hosts
```
russellb, line 327: I would also be fine changing this back. We only need the added subdomain for prod.

russellb (approved):
I’m good with this. You could have left the local domain as-is, but it’s easy to change later too.

.github/workflows/build.yml
vishnoianil: yeah I thought of that, but just for consistency sake, i changed it locally as well. Hopefully it won't create confusion when it comes to configuration of deployments (locally and prod/qa).

Issue #533, opened 2023-03-05
make recreate-db errors


nerdalert:
### Describe the bug

When running `make recreate-db` I am consistently observing the following:

```
ip-172-31-25-203:~/nexodus$ make recreate-db
postgrescluster.postgres-operator.crunchydata.com "database" deleted
certificate.cert-manager.io/nexodus-auth-cert configured
postgrescluster.postgres-operator.crunchydata.com/database created
make[1]: Entering directory '/home/ubuntu/nexodus'
error: status is not found
make[1]: *** [Makefile:312: init-db] Error 1
make[1]: Leaving directory '/home/ubuntu/nexodus'
make: *** [Makefile:355: recreate-db] Error 2
```

### How to Reproduce

```
make run-on-kind
make recreate-db
```

### Expected Behavior

:green_circle:

### Deployment Environment

I am observing the following both locally and on ec2

### Additional Context

Have not done any triaging yet though it appears a status is unexpected (may just need more time to converge). There is a lot more in the feature since I added it so I'm betting someone on the team will know from the log output. If not assign it to me and I'll dig in. The work around is:

```
kubectl delete -n nexodus deploy/apiserver postgrescluster/database deploy/ipam
kubectl apply -k ./deploy/nexodus/overlays/dev
```

Some logs from the api-server when recreate-db doesn't complete:

```
2023-03-05T06:19:32.464Z	ERROR	zap@v0.1.0/zap.go:87	failed to connect to `host=database-primary.nexodus.svc user=apiserver database=apiserver`: failed SASL auth (FATAL: password authentication failed for user "apiserver" (SQLSTATE 28P01))	{"status": 500, "method": "GET", "path": "/users/me", "query": "", "ip": "10.244.0.5", "user-agent": "Go-http-client/2.0", "latency": "12.246268ms", "time": "2023-03-05T06:19:32Z", "traceID": "a35d44fb4ff9cb7caecd40e9d46439f4"}
github.com/gin-contrib/zap.GinzapWithConfig.func1
	/go/pkg/mod/github.com/gin-contrib/zap@v0.1.0/zap.go:87
github.com/gin-gonic/gin.(*Context).Next
	/go/pkg/mod/github.com/gin-gonic/gin@v1.9.0/context.go:174
go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin.Middleware.func1
[...]

PR #550, opened 2023-03-06
fix: #533 remove attempt to wait for the DB to be up

chirino:
before restarting the depending pods.  It was not reliable.


nerdalert (approved):
That did it :tada:

PR #556, opened 2023-03-07
Keycload will throw error while parsing nexodus.json

vishnoianil:
file because it doesn't allow comments in the json file


dave-tucker (approved):

russellb:
sorry about that!

PR #559, opened 2023-03-08
feat: move the go-oidc-agent http handlers into the apiserver process.

chirino:
also add an --insecure-skip-tls-verify flag to nexd


deploy/nexodus/base/apiserver/kustomization.yaml
```diff
@@ -12,9 +12,13 @@ configMapGenerator:
       - NEXAPI_TRACE_INSECURE="1"
       - NEXAPI_FFLAG_MULTI_ORGANIZATION=false
       - NEXAPI_DB_SSLMODE=require
+      - NEXAPI_DOMAIN=api.try.nexodus.local
+      - NEXAPI_REDIRECT_URL=https://try.nexodus.local/#/login
+      - NEXAPI_ORIGINS=https://try.nexodus.local
```
dave-tucker, line 17:
```suggestion
      - NEXAPI_DOMAIN=api.nexodus.local
      - NEXAPI_REDIRECT_URL=https://nexodus.local/#/login
      - NEXAPI_ORIGINS=https://nexodus.local
```

go.mod
```diff
@@ -163,3 +163,5 @@ require (
 replace github.com/docker/docker => github.com/docker/docker v20.10.3-0.20221013203545-33ab36d6b304+incompatible // 22.06 branch
 
 replace github.com/metal-stack/go-ipam => github.com/dave-tucker/go-ipam v0.0.0-20230220173413-c25c56b2408b
+
+replace github.com/redhat-et/go-oidc-agent => github.com/chirino/go-oidc-agent v0.0.0-20230308174912-b4dc11f2e6eb // addroutes branch
```
dave-tucker, line 167: Can you open a PR please in go-oidc-agent please? I'd rather that we didn't use `replace` unless we really have to.

internal/nexodus/nexodus.go
```diff
@@ -159,6 +162,7 @@ func NewNexodus(ctx context.Context,
 		version:             version,
 		username:            username,
 		password:            password,
+		skipTlsVerify:       insecureSkipTlsVerify,
```
dave-tucker, line 165: That should be `skipAuthServerTlsVerify` and we should ONLY set that TLS option on the http client used in the go-oidc-agent.

internal/routers/routers.go
```diff
@@ -29,10 +29,22 @@ func NewAPIRouter(
...
 	gin.SetMode(gin.ReleaseMode)
 	r := gin.New()
 
+	loggerMiddleware := ginzap.GinzapWithConfig(logger.Desugar(), &ginzap.Config{TimeFormat: time.RFC3339, UTC: true, TraceID: true})
+	r.Use(otelgin.Middleware(name))
+	r.Use(ginzap.RecoveryWithZap(logger.Desugar(), true))
+
+	r.Use(browserFlow.CookieSessionMiddleware())
+	agent.AddCodeFlowRoutes(r.Group("/browser", loggerMiddleware), browserFlow)
+	agent.AddDeviceFlowRoutes(r.Group("/device", loggerMiddleware), deviceFlow)
```
dave-tucker, line 46: We should probably look at the API here.
It would make more sense to me to do something like this:
```go
webAuthProxy := r.Group("/web/", loggerMiddleware)
webAuthProxy.Use(agent.NewRouter( ...)
```

As it stands the API names are incorrect. One doesn't do code flow, and the other device flow.
They both do whatever flows are configured for the client-id/secret.

internal/routers/routers.go
```diff
@@ -47,14 +59,6 @@ func NewAPIRouter(
...
 	p.Use(r)
 
-	r.Use(otelgin.Middleware(name))
-	r.Use(ginzap.GinzapWithConfig(logger.Desugar(), &ginzap.Config{TimeFormat: time.RFC3339, UTC: true, TraceID: true}))
-	r.Use(ginzap.RecoveryWithZap(logger.Desugar(), true))
-
-	r.GET("/health", func(c *gin.Context) {
-		c.JSON(http.StatusOK, gin.H{"message": "ok"})
-	})
-
```
dave-tucker, line 57: Why?

internal/routers/routers.go
```diff
@@ -81,8 +85,10 @@ func NewAPIRouter(
 	}
 	verifier := provider.Verifier(config)
 
-	private := r.Group("/")
+	private := r.Group("/api", loggerMiddleware)
 	{
+		private.Use(browserFlow.CorsMiddleware())
```
dave-tucker, line 90: I think you'll have to actually copy/paste the middleware config, you can't just use the one from only one of the agents. Things like trustedOrigins are would be incorrect etc...

dave-tucker (changes requested):
A good start, but I think this has a way to go before I'm happy with it.
I've left some initial comments, but I think I'll quickly bring in go-oauth-agent here as a package and archive the old repo. That should allow us to review all the changes in one place.

internal/routers/routers.go
dave-tucker: Note: I blame myself for those API names 😆

russellb:
This PR and discussion about it highlight to me that we lack good documentation of architectural principles we want to follow for our control plane. One aspect of that architecture is around microservices and how we decide when something should be part of the apiserver or as a supporting service. I don't think I have the answers, nor could I write a document about this on my own. I do think we as a team should write something that helps inform changes like this and how we design future capabilities.

What do you think?

vishnoianil:
@chirino overlay/prod and overlay/qa yamls needs update as well for the changes in this PR.

internal/routers/routers.go
chirino: I don't mind the path rename or method refactors.  Could we do it in a follow up PR?

internal/routers/routers.go
chirino: the origin were setup correctly.. but I didi need to move the CorsMiddleware() the root router.

chirino:
@vishnoianil how do I test changes to overlay/prod and overlay/qa ?

chirino:
@dave-tucker rebased on main.

chirino:
got a better version of this in the works..

PR #565, opened 2023-03-08
pkg: Add go-oidc-agent

dave-tucker:
This brings in go-oidc-agent from
https://github.com/redhat-et/go-oidc-agent and its dependent pkgs. It was originally incubated out of tree, but since we're the only user it makes sense for it to co-exist here.


.markdownlint-cli2.yaml
```diff
@@ -2,3 +2,5 @@ config:
   line-length: false
   no-emphasis-as-header: false
   first-line-heading: false
+ignores:
+  - pkg
```
russellb, line 6: Going to fix this after importing?

pkg/oidcagent/pkg/cookie/README.md
```diff
@@ -0,0 +1,105 @@
+# Cookie store for [Session](https://github.com/go-session/session)
```
russellb, line 1: Looks like you imported this twice? once as `pkg/cookie` and once as `pkg/oidcagent/pkg/cookie` ?

pkg/oidcagent/pkg/ginsession/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
```
russellb, line 1: This one is imported twice too

pkg/oidcagent/pkg/cookie/README.md
dave-tucker: oops. fixed.

.markdownlint-cli2.yaml
dave-tucker: Maybe. Added more detail to the commit message.
Some of the packages that we have are forks of older, unmaintained packages.
Not sure if we should edit the README's there or not.

russellb:
> @russellb @chirino this now includes hiram's patch to go-oidc-agent, and therefore I did some git magic to preserve the git history. The last 4 commits are purely organizational/ci related, no go-oidc-agent code was changed.

Nice magic!

> I plan on adding another commit as a follow-up to rename a few APIs, then we can rebase #559

Adding another commit to this PR? Or another PR after this one? I wasn’t sure if you meant to hold off on merging.

dave-tucker:
>  Or another PR after this one? 

I meant that 😆

.markdownlint-cli2.yaml
```diff
@@ -2,3 +2,8 @@ config:
   line-length: false
   no-emphasis-as-header: false
   first-line-heading: false
+paths:
+ - "**/md"
```
russellb, line 6: This is different from what was in the command line. Did you verify it still works?

dave-tucker: well the lint is clear, but this should be `**/*.md`. Will fix it now.

dave-tucker: With fix applied:
```
markdownlint-cli2 v0.6.0 (markdownlint v0.27.0)
Finding: **/*.md !pkg !**/node_modules/*
Linting: 21 file(s)
Summary: 0 error(s)
```

PR #566, opened 2023-03-08
gen-docs: Add job pre-reqs and pretty Makefile output

russellb:
Add pretty output for running `make gen-docs`. Full output from the tool can be seen when you run with `NOISY_BUILD=y` like this:

    $ NOISY_BUILD=y make gen-docs

Also make some other jobs depend on the gen-docs job. It runs in 20% of the time of go-lint, so make jobs that depend on go-lint depend on this one too. If the docs are broken, don't waste time running all the more consuming jobs.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -82,7 +82,7 @@ jobs:
           fi
 
   go-unit:
-    needs: go-lint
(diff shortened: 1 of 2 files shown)


dave-tucker (approved):
LGTM

PR #567, opened 2023-03-09
Add nexd logging to e2e

nerdalert:
- Commit creates a file with the nexd run options for each node, copies them to the node for execution and pipes logs to a file. The file is then picked up in gather for a dump in the case of a test failure for easier debugging.


vishnoianil (approved):

russellb:
Anil already approved it, so I'm just merging

PR #574, opened 2023-03-10
feat: add --insecure-skip-tls-verify flag to the nexd cli tool

chirino:
add --insecure-skip-tls-verify flag to the nexd cli tool.  This can be handy if your testing against a nexodus server that is using self signed certs and your device has not installed those certs into the system certs.

chore: Uppercase first letter of all flag Usage to be consistent with how other tools like git and kubectl show help.

### DIFF
--- a/cmd/apiserver/main.go
+++ b/cmd/apiserver/main.go
@@ -53,6 +53,7 @@ func init() {
 
 // @BasePath  		/api
 func main() {
+	// Override to capitalize "Show"
 	cli.HelpFlag.(*cli.BoolFlag).Usage = "Show help"
 	app := &cli.App{
 		Name: "nexodus-controller",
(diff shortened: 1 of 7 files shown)


cmd/apiserver/main.go
```diff
@@ -53,31 +53,32 @@ func init() {
 
 // @BasePath  		/api
 func main() {
+	cli.HelpFlag.(*cli.BoolFlag).Usage = "Show help"
```
russellb, line 56: It took me a bit to figure out why this was needed. It's to get `Show` to be capitalized, right? I appreciate your dedication to getting the output consistent!

A comment would be helpful for the future so others know why this was needed.

russellb (approved):
One trivial comment, which I can just push a small commit to address

cmd/apiserver/main.go
russellb: pushed a commit to the branch for this

chirino:
Yep.  It’s to upper case the help too

PR #575, opened 2023-03-10
chore: use os specific build tags for the os specific code.

chirino:
Also run go-lint against each supported OS.

This should reduce binary sizes a bit.


internal/nexodus/keys_darwin.go
```diff
@@ -0,0 +1,25 @@
+//go:build darwin
+
+package nexodus
+
+import "fmt"
+
+// handleKeys will look for an existing key pair, if a pair is not found this method
+// will generate a new pair and write them to location on the disk depending on the OS
+func (ax *Nexodus) handleKeys() error {
```
nerdalert, line 9: We should add moving to https://pkg.go.dev/golang.zx2c4.com/wireguard/wgctrl/wgtypes#GenerateKey to the backlog

nerdalert (approved):
LGTM, 🎉

PR #577, opened 2023-03-11
Add e2e coverage for nexctl socket status

nerdalert:
- Replaces sleeps for discovery node joins with nexctl nexd status

### DIFF
--- a/cmd/nexctl/local.go
+++ b/cmd/nexctl/local.go
@@ -11,7 +11,7 @@ import (
 func callNexd(method string) (string, error) {
 	conn, err := net.Dial("unix", "/run/nexd.sock")
 	if err != nil {
-		fmt.Printf("Failed to connect to nexd: %+v", err)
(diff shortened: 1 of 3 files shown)


nerdalert:
@chirino refresh me if you added e2e support for OSX. If so this breaks it since socket status isn't supported on OSX yet. Can put this on the back burner if so.

chirino:
Not yet

nerdalert:
@chirino I deleted the comment after thinking through it and its being executed in the container it shouldn't break OSX. Ty!

russellb (approved):
Nice one, Brent! This lgtm. I have a couple little ideas, but I'll post them as a follow up.

russellb:
> Nice one, Brent! This lgtm. I have a couple little ideas, but I'll post them as a follow up.

This is what I was thinking about: https://github.com/nexodus-io/nexodus/pull/579

The `CheckPeriodically()` helper didn't reduce the size of the code as much as I thought, but it still seemed like a useful pattern to put in a helper.

PR #594, opened 2023-03-13
docs/design: Document Stack Decisions

dave-tucker:


docs/development/design/stack.md
```diff
@@ -0,0 +1,60 @@
+# Nexodus Stack
+
+## Summary
+
+This proposal documents the decisions made around the Nexodus software stack.
```
russellb, line 5: I don't know how much I care, but in the `project-naming.md` design doc, we call out "Nexodus Service" as the name that refers to the control plane side. Other docs still say "Nexodus stack" or "Nexodus controller" in places, but I've been trying to make myself stick to "Nexodus Service," unless we pick something else.

You could also add here something about "early key decisions made before we started capturing things in docs/design/".

docs/development/design/stack.md
```diff
@@ -0,0 +1,60 @@
...
+As the stack and complexity grew, we decided to exclusively support Kubernetes, requiring local development to use KIND.
+
+To support use of KIND for local development, and OpenShift for production deployment we settled on using Kustomize.
+
+#### Decisions
+
+- We only support deployment on K8s
+- Kustomize is our deployment configuration tool
+- All manifests in `deploy/base` MUST be useable in any K8s distribution.
+- We allow operators to be used, provided they may also be installed on any K8s distribution.
```
russellb, line 23: How about also:

- While we may use OpenShift for the project's production cluster, we will not make any changes that require the use of OpenShift to run a production instance of the service.

docs/development/design/stack.md
```diff
@@ -0,0 +1,60 @@
...
+
+### Microservices First
```
russellb, line 25: I think some of this could go into `docs/development/development.md` in the "guiding principles" section. The longer text in here is good, but I'd like to come up with a 1 or 2 sentence summary that captures the intent of it.  Here's a first try:

- **Modularity** - We embrace microservices and modularity in general, though we will compromise if being non-modular is simpler or offers a better developer or user experience. See `[Nexodus Stack](stack.md)` for additional comments on this.

docs/development/design/stack.md
```diff
@@ -0,0 +1,60 @@
...
+
+#### Background
+
+An early agreed architectural principal of the Nexodus stack was to embrace microservices. This would allow us to iterate on components seperately, give us maximum re-use if the project failed, and most importantly, allow us to scale each component seperately.
```
chirino, line 29: While there is nothing wrong microservices, I think it's important to note when they are most useful, and thats when you can delegate the operation of each service to a different group and thus decouple and scale the organization easier.  If 1 group has to run and maintain multiple services, sometimes a monolith makes more sense and has lower overheads.  So I think we need to be careful what we choose to make a microservice.

docs/development/design/stack.md
```diff
@@ -0,0 +1,60 @@
...
+
+This resulted in the following decisions:
+
+1. IPAM would be a separate service, consumed over gRPC, since it already has an [upstream project](https://github.com/metal-stack/go-ipam).
+1. Authentication would use Open ID Connect (OIDC), provided by default in our stack by Keycloak.
+1. The `go-oidc-agent` proxy would be deployed as a binary, in-front of existing services vs. embedding that logic as library code in the apiserver.
```
chirino, line 35: I think we are going to end up embedding it soon though right?

docs/development/design/stack.md
```diff
@@ -0,0 +1,60 @@
...
+1. We would use [gorm](https://gorm.io) as an ORM to give us choice to switch out the DB implementation if required.
+1. We would use PostgreSQL as our database of choice, and deploy it via the Crunchy Data Postgres Operator.
```
chirino, line 37: I have a follow up design PR to also support CockroachDB as it has several attributes that make it desirable for run highly available services.

PR #597, opened 2023-03-14
docs: document the need for rate limiting in the service.

chirino:


docs/development/design/rate-limiting.md
```diff
@@ -0,0 +1,37 @@
+# Used Envoy for Rate Limiting
```
dave-tucker, line 1:
```
# Use Envoy for Rate Limiting
```

docs/development/design/rate-limiting.md
```diff
@@ -0,0 +1,37 @@
...
+
+<!-- [Issue #600](https://github.com/redhat-et/issues/600) -->
+
+## Summary
+
+To protect the apiserver service from a misbehaving tenant, the rest APIs should be rate limited. This will allow us to better reason about how to scale the service to a large number of tenants as there will be limit to the ammout of apiserver resources each tenant can consume.
```
dave-tucker, line 7: `s/rest/REST/g`

docs/development/design/rate-limiting.md
```diff
@@ -0,0 +1,37 @@
...
+
+This also translates to a limit of number of devices that a tenant can enroll before he starts to hit rate limits.  There should be a way to increase the limits per tenant.
```
dave-tucker, line 9: >  There should be a way to increase the limits per tenant.

How does your proposal allow this? Limitador config is file based, and can't be adjusted at runtime as far as I can tell. Adjusting files - which are part of the deployment - isn't a great solution. Ideally the rate-limit config could be read from the database.

docs/development/design/rate-limiting.md
```diff
@@ -0,0 +1,37 @@
...
+Using an Envoy proxy has the following benefits:
+
+* Fast proxy used by many projects and supported by many organizations
+* We could possibly move more functionality to it in the future (like oauth processing)
+* Some K8s platforms are moving to envoy as the Ingress gateway so in the future we may be able to move this functionality into the ingress gateway (avoiding a proxy hop).
+* Aligned with service mesh deployment models.
+
+## Alternatives Considered
+
+Live without rate limiting.  If you have a custom Nexodus service deployment, and don't share it with multiple tenants, you may not need rate limiting.
```
dave-tucker, line 30: This isn't the only alternative though.

The concern that I have is how we communicate the rate-limiting information to the proxy.
It would be ideal if this could be read from the database to support different limits per-tenant.

We could actually embed this logic in the apiserver: https://pkg.go.dev/github.com/gin-gonic/examples/ratelimiter

Or perhaps Limitador has an API that we could drive directly.

dave-tucker (changes requested):

docs/development/design/rate-limiting.md
chirino: Maybe we could some info into the JWT and use that to pick the right rate limiting config for the tenant.

docs/development/design/rate-limiting.md
chirino: Maybe we could put some info into the JWT and use that to pick the right rate limiting config for the tenant.

docs/development/design/rate-limiting.md
chirino: Given rate limiting is such a common service need, we should try to use something that can be commonly supported across services.

And maybe even more importantly, rate limited requests should be dropped as early as possible in the proxy chain to avoid adding load to the downstream servers.  Like once Gateway API becomes a commonly supported API on k8s, it should be possible to have the ingress gateway use limitador to enforce the rate limits.  Allowing us to drop the request even earlier in the proxy chain.

docs/development/design/rate-limiting.md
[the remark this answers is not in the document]
chirino: done.

docs/development/design/rate-limiting.md
dave-tucker: That would require moving user/organization membership into the ID Provider though 🤔 
It's something I pondered on when writing the AuthZ design, but not sure that it's viable for us.

chirino: Don't think it has to be the user/org in the IDP.  it could be a tier level attribute assigned to the user like: free, bronze, gold, etc.

dave-tucker: Even so, someone needs to assign that attribute in the IdP and that can't be done by OIDC.

docs/development/design/rate-limiting.md
```diff
@@ -0,0 +1,35 @@
+# Used Envoy for Rate Limiting
+
+<!-- [Issue #600](https://github.com/redhat-et/issues/600) -->
+
+## Summary
+
+To protect the apiserver service from a misbehaving tenant, the REST APIs should be rate limited. This will allow us to better reason about how to scale the service to a large number of tenants as there will be limit to the amount of apiserver resources each tenant can consume.
```
dave-tucker, line 7: Can we clarify here that "tenant" is User and not Organization?

chirino: +1

docs/development/design/rate-limiting.md
chirino: Passing jwt claims to limitador is possible

chirino:
@dave-tucker take another peek, I think it should resolve all your comments.

Issue #598, opened 2023-03-14
Support Cockroach DB


chirino:
### Describe the Problem Statement

Support a DB technology that can reduce DB downtime of the service and allow the DB to horizontally scale.  

### Describe the Enhancement

We should support Cockroach DB in addition to Postgres.  Cockroach shards and replicates over several work nodes allowing it to be horizontally scalable and have higher availability than traditional single node databases.  It also support doing zero downtime upgrades of the database.

Cockroach is already highly compatible with Postgres, it uses the same client drivers and supports most of the same SQL dialect.

### Alternate Solutions

* Don't support it - this would allow us to stay more nimble by only having to support one database type, but keeping the sevice up with a high level of 9's might be harder.

### Additional context

* [How to choose between PostgreSQL and CockroachDB](https://docs.rackspace.com/blog/how-to-choose-between-postgresql-and-cockroachdb/)
* [Demo of CockroachDB's in-place rolling upgrades](https://dantheengineer.com/demo-of-cockroachdbs-in-place-rolling-upgrades/)

russellb:
Completed in #599

PR #599, opened 2023-03-14
docs: document the cockroach db enhancement proposal. #598

chirino:


docs/development/design/cockroachdb.md
```diff
@@ -0,0 +1,22 @@
...
+
+> [Issue #598](https://github.com/redhat-et/issues/598)
+
+## Summary
+
+Support a DB technology that can reduce DB downtime of the service and allow the DB to horizontally scale.  
+
+## Proposal
+
+We should support Cockroach DB in addition to Postgres.  Cockroach shards and replicates over several work nodes allowing it to be horizontally scalable and have higher availability than traditional single node databases.  It also support doing zero downtime upgrades of the database.
```
dave-tucker, line 11: Could we support Cockroach in place of Postgres?

docs/development/design/cockroachdb.md
```diff
@@ -0,0 +1,22 @@
...
+
+Cockroach is already highly compatible with Postgres, it uses the same client drivers and supports most of the same SQL dialect.
```
dave-tucker, line 13: There are some issues with `gorm` and cockroachdb right? We should document those here and we plan to work around them OR fix them.

dave-tucker (changes requested):

docs/development/design/cockroachdb.md
chirino: If we could support both, it would be ideal.  The downside to Cockroach is that it's more resource intensive.  You need more hardware to get the similar performance to Postgres.  So for smaller deployments or on prem deployments, I think Postgres makes alot of sense.

docs/development/design/cockroachdb.md
chirino: doen.

dave-tucker (approved):
LGTM

russellb (approved):

nerdalert (approved):

PR #601, opened 2023-03-14
Remove go routines from nexd launch in e2e

nerdalert:
- Now that we launch from a file for easier log piping we can remove the groutines.
- Fixed a teardown issue of a container in the TestRelay test.
- Bumped context timers for tshooting parallel tests.


chirino (approved):
LGTM

PR #607, opened 2023-03-16
Add a check for a relay node in peer setup for local peering

### LINKED ISSUE #606: Brief bug summary
### Describe the bug

IPAM assigned addresses get a /32 prefix appended to their assigned IP. Relays/Hub Routers do not get the /32 appended since the route/allowed-ip that peer nodes will use is the organization prefix as a catch all, e.g. 100.100.0.0/16. 

Locally peered addresses are still trying to add that prefix if `peer.Relay == true`. Reflexive addresses check for a peer as a relay and ignore trying to add them as it is taken care of earlier in `buildPeersConfig()`. This just needs to be checked for local peering as well.

### How to Reproduce

Peer local addresses to a relay. E2e replicates this scenario since reflexive addresses are not tested there.

### Expected Behavior

-

### Deployment Environment

make e2e and attach nodes that are locally and not reflexively peered.

### Additional Context

_No response_

nerdalert:
- Resolves #606 by applying the same relay check logic that currently exists for reflexive address peering to local address peering.

### DIFF
--- a/internal/nexodus/wg-peers.go
+++ b/internal/nexodus/wg-peers.go
@@ -113,7 +113,7 @@ func (ax *Nexodus) buildPeersConfig() {
 
 		// If both nodes are local, peer them directly to one another via their local addresses (includes symmetric nat nodes)
 		// The exception is if the peer is a relay node since that will get a peering with the org prefix supernet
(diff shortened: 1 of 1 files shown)


dave-tucker (approved):
LGTM

PR #608, opened 2023-03-17
Add a check for existing relay in the nexd agent

nerdalert:
- This adds a check prior to a node joining an organization as a relay node to verify there isn't an existing one already in the database. If there is an existing relay, the agent exits with instructions to delete the device ID in order to join as a relay for the organization. Currently if two nodes join as a relay it is not handled and becomes a free for all as to which relay a node will add a tunnel with the org prefix.
- Also adds a device relay field to nexctl and the UI.


russellb (approved):
Thanks for the response! This is a good improvement. As usual, we’ll keep iterating to improve!

PR #614, opened 2023-03-19
feat: add initial support for writing cucumber based e2e tests.

chirino:


dave-tucker:
Generally in favour of more tests, but I'm not familiar with the library.
I'm much more familiar with seen https://onsi.github.io/ginkgo/ which I believe is more widely adopted for BDD style tests in Go.
Curious as to whether you gave that a look, or whether the "human readable" cucumber style tests were a hard requirement.

chirino:
Yeah. I tend to like that they are more readable, and that they compare api responses against expected json, without an intermediate client doing any magic.

chirino:
It’s also easier to write test that send bad requests to the api servers.

nerdalert:
Same as Dave, only used Gingkgo but have seen cucumber in the wild. I'm good learning a new framework as it looks like you've laid out a nice groundwork. I'm all for easier and more readable.

nerdalert (approved):

PR #630, opened 2023-03-22
nexd: Rename env var with APEX_ prefix

russellb:
Rename this env var to use a NEXD_ prefix like all other env vars.


chirino (approved):
LGTM

PR #631, opened 2023-03-22
nexd: Fix mistakes in ctlserver exit handling

russellb:
This commit fixes a couple of problems in exit handling from the unix socket control server.

First, the code mistakenly did a Done() on a waitgroup instead of Wait(). This caused a panic in a case where no subroutines were ever started due to lack of permissions to open the socket.

With this fixed, it exposed that the server was not exiting properly when the context was canceled. The listener routine needed to skip trying to return its error when everything was shutting down.

Finally, if the control server does exit with an error, such as having insufficient permissions to open the socket. It was trying to restart it in a busy loop. This adds a sleep to rate limit it. When running with insufficient permissions, you'd see this retry many, many times before the rest of nexd figured out that it couldn't create a wireguard interface and exit.

### DIFF
--- a/internal/nexodus/ctlserver_linux.go
+++ b/internal/nexodus/ctlserver_linux.go
@@ -4,12 +4,14 @@ package nexodus
 
 import (
 	"context"
-	"github.com/nexodus-io/nexodus/internal/util"
 	"net"
 	"net/rpc"
 	"net/rpc/jsonrpc"
 	"os"
 	"sync"
+	"time"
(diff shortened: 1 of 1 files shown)


russellb:
Manually tested with
- run nexd normally and exit with ctrl-c
- Run nexd without root permissions and observe that it exits without a panic or the control server restarting countless times before exit

vishnoianil (approved):

PR #634, opened 2023-03-23
Makefile: drop go vet

russellb:
This isn't needed since golangci-lint already runs this by default.

You can check which linters are used if you run the tool manually with the verbose flag, `-v`.

```
$ golangci-lint run -v ./...
...
INFO [lintersdb] Active 9 linters: [errcheck errorlint gosec gosimple govet ineffassign staticcheck typecheck unused]
...
```


chirino (approved):
lgtm

PR #636, opened 2023-03-23
apiserver: Add rego store for user/org mappings

dave-tucker:
Testing that this doesn't break anything in CI...
Need to add another commit to update the OPA policy to see if I can:
1. Get `org_id` for requests that affect organizations
2. Check `org_id` is in `data.user_org_map[user_id]`


chirino:
Keeping a copy of the user/org mappings in memory does not seems like a scalable solution.  
Perhaps we should do the resource filtering in the db layer.

dave-tucker:
> Keeping a copy of the user/org mappings in memory does not seems like a scalable solution.
> Perhaps we should do the resource filtering in the db layer

It's a trade off.

Keeping it in-memory for now allows us to add more complex policy at a lower overhead than SQL queries.
For example, The policy for `DELETE /organizations/$org_id/devices/$device_id/`  becomes quite complex even now:
"Allow device to be off-boarded if the user is in the org, and the user owns the device OR the user is the org owner".
That's at least 3 tables that need to be hit - Users, Organizations (via user_organizations), Devices.

In-memory means that this will simply hit the Devices table.

OPA has different models for providing data. We might chose to make this pull at some point (i.e exposing a gRPC service to expose user_org mappings and user_device mappings for policy), which would scale much better across many replicas of apiserver.

chirino:
Good enough for now.  But we should work to remove the in memory store later.

PR #639, opened 2023-03-24
Add error handling to symmetric NAT checks

nerdalert:
- If the first STUN lookup fails, return an error rather than moving on to the second.
- Changed logging to debug and added a more descriptive output to the error.

### DIFF
--- a/internal/nexodus/nexodus.go
+++ b/internal/nexodus/nexodus.go
@@ -178,7 +178,12 @@ func NewNexodus(ctx context.Context,
 		return nil, err
 	}
 
-	ax.nodePrep()
+	// remove orphaned wg interfaces from previous node joins
+	ax.removeExistingInterface()
+
+	if err := ax.symmetricNatDisco(); err != nil {
+		ax.logger.Warn(err)
+	}
 
 	return ax, nil
 }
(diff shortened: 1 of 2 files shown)


nerdalert:
If the STUN server is unreachable, debug output looks like the following and the wait time is only a couple of seconds.

```
2023-03-23T23:12:18.429-0400	INFO	nexd/main.go:32	Debug logging enabled
2023-03-23T23:12:18.430-0400	DEBUG	nexodus/wg-iface.go:16	existing link not found: utun8
2023-03-23T23:12:18.430-0400	DEBUG	nexodus/stun.go:18	dialing stun server stun1.l.google.com:19302
2023-03-23T23:12:18.571-0400	DEBUG	nexodus/nexodus.go:185	STUN binding request failed: a firewall may be blocking UDP connections to stun1.l.google.com:19302
2023-03-23T23:12:18.571-0400	DEBUG	nexodus/ctlserver_darwin.go:11	Ctl interface not yet supported on OSX
```

Since two lookups are required for inferring symmetric NAT, there is no point in doing the second if the first fails. This will change as the ICE implementation progresses and is being used for discovery but this resolves the immediate issue of a long delay waiting for lookups when STUN ports are blocked. One note, the `reuseport.Dial` won't return an error since it is UDP without writing/reading from that socket.

internal/nexodus/nexodus.go
```diff
@@ -178,7 +178,12 @@ func NewNexodus(ctx context.Context,
 		return nil, err
 	}
 
-	ax.nodePrep()
+	// remove orphaned wg interfaces from previous node joins
+	ax.removeExistingInterface()
+
+	if err := ax.symmetricNatDisco(); err != nil {
+		ax.logger.Debug(err)
```
vishnoianil, line 185: shell if we should throw this as a warning? Because this can eventually lead to wrong peer configuration.

vishnoianil (approved):
@nerdalert Minor comments, but looks good to me.

internal/nexodus/nexodus.go
nerdalert: @vishnoianil I was on the fence, this helped! The other thing I was thinking is this is the best early warning for outbound UDP connectivity issues. Odds are if this random non-rfc google UDP STUN port is blocked, so is WG.

PR #642, opened 2023-03-24
Ensure stun requests are IPv4

russellb:
Fix stun requests to ensure they are always IPv4 and fix one only spot that assumed IPv4 but didn't force it.

```
commit ee2fa95a4a499831544f97be79d060a9d4e58df3 (HEAD -> v6fixes)
Date:   Fri Mar 24 14:19:00 2023 -0400

    utils: Ensure discoverGenericIPv4 is actually IPv4
    
    Force this method to use IPv4 since that is the intent.
    
    In passing, using the net.SplitHostPort() method instead of manually
    splitting the string, as this is best practice to avoid accidental
    parsing mistakes (primarily with IPv6, but we should use the helper
    everywhere).
    
commit 5f18f43b2dae76a1ec98b14465b64c4416d36c25
Date:   Fri Mar 24 14:13:28 2023 -0400

    stun: Ensure we issue stun requests via IPv4
    
    Nexodus is still an IPv4-only system, so we must ensure that we're
    issuing stun requests via IPv4. This will need to be revisited as part
    of proper IPv6 support.
    
```

### DIFF
--- a/internal/nexodus/stun.go
+++ b/internal/nexodus/stun.go
@@ -2,10 +2,11 @@ package nexodus
 
 import (
 	"fmt"
+	"net"
+
 	"github.com/libp2p/go-reuseport"
 	"github.com/pion/stun"
 	"go.uber.org/zap"
-	"net"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


nerdalert (approved):
Good catch!

Issue #187, opened 2022-11-14
Add a userspace networking mode for the agent


russellb:
Running the `apex` agent today requires root access, or at least `CAP_NET_ADMIN` capabilities to be able to set up a wireguard network interface. This limits `apex` from being used in many container-based environments where this permissions level is not normally available.

To make `apex` the most useful in those environments, the agent needs a way to provide connectivity into an Apex Zone using only userspace networking.

As some inspiration, Tailscale supports a mode like this: https://tailscale.com/kb/1112/userspace-networking/

dave-tucker:
Soft blocked on #168

nerdalert:
### userspace wireguard-go 
I dug into this a bit, we already run the user space `wireguard-go` for Darwin, it's basically the same for Linux. Create the dev with wireguard-go and then everything else is the same, e.g. using the `wg` binary to configure and read from the interface.

Since it's using `/dev/net/tun` it still requires privileges. Here is an example:

```
# Create the user space interface:
sudo ./wireguard-go wg0
$ sudo wg show

interface: wg0
  listening port: 59502

# Configure the userspace interface with the normal wg binary
sudo wg set wg0 listen-port 12345 private-key /etc/wireguard/private.key

$ sudo wg show
interface: wg0
  public key: VchcCdEgTREfX9CzTgmjdvUdeDQ6xv/wmaftA4TQdXs=
  private key: (hidden)
  listening port: 12345

# Killing the wireguard-go process destroys the interface (as you may have noticed in our OSX implementation)
```

It would only be a few lines of code to add a `--userspace` option for Linux but it's probably not nessecary until we know what the use case is. One would be the same implementation across platforms but there are even some disclaimers around Linux and WIndows on the [wireguard-go](https://github.com/WireGuard/wireguard-go) project site I remember seeing when doing the windows implementation.

If anyone wants to kick the tires on it in a container here is a quick paste to get you going, tldr, I dont see anyway around cap net admin.


```
### Note: this will complain about prvieldges. To get it working add net --cap-add=NET_ADMIN
docker run -itd  -v /dev/net/tun:/dev/net/tun --name wg-go  ubuntu
docker exec -i -t $(docker ps -l -q) /bin/bash

# Inside the container
apt-get update 
apt-get install -y wget git make 

# Build
wget https://go.dev/dl/go1.19.3.linux-amd64.tar.gz
rm -rf /usr/local/go 
[...]

PR #651, opened 2023-03-26
design: Propose a userspace-only mode for nexd

russellb:
This document includes a proposal for adding a new proxy mode to `nexd` that allows interfacing with a Nexodus network with a userspace-only approach. This allows connectivity in environments where privileges to create new network interfaces are not available.

Issue #187

### DIFF
--- a/docs/development/design/userspace-mode.md
+++ b/docs/development/design/userspace-mode.md
@@ -1,0 +1,108 @@
+# Enhancement Title
+
+> [Issue #187](https://github.com/nexodus-io/nexodus/issues/187)
+
+## Summary
+
+The traditional operation of `nexd` requires elevated privileges to create a network interface. This prevents usage in certain environments where providing these privileges is not possible. The primary example is a developer using containers. A development may have application tunneling needs to reach a service in some remote location, but is unable to run with elevated privileges. Nexodus should provide a mode that works for this use case.
+
+## Proposal
+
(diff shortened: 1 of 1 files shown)


docs/development/design/userspace-mode.md
```diff
@@ -0,0 +1,108 @@
...
+### Proxy Configuration
+
+As a first step, `nexd` will take proxy configuration via two configuration flags: `--ingress` and `--egress`.
+
+`--ingress` is used to proxy for connections coming from the Nexodus network to a server accessible by `nexd` via a local network.
+
+`--egress` is used to proxy for connectoins coming from a network local to `nexd` to a server reachable over the Nexodus network.
+
+In both cases, the format for the value is `[protocol]:[port]:[desination_ip]:[destination_port]`. `protocol` can be `tcp` or `udp` to start.
+
```
chirino, line 57: At some point it might be nice to have `nextctl` commands to manage the ingress/egress list.  Kinda like:
* `nexctl list [ingress|egress]`
* `nexctl [create|delete] [ingress|egress] <port-mapping>`

This would be super handy if your on a system and and the services you want to access or or are exposing a bit dynamic and you don't want to have interrupt existing connections.  

If we do that, would those persist between nexd restarts?   If they do, does it kind of ignore options on subsequent restarts of the `nexd` cli?

chirino (approved):
LGTM -

docs/development/design/userspace-mode.md
russellb: Yeah, I agree that runtime management would be useful. See a few paragraphs down for some comments on this. I proposed leaving for the future because I didn’t want to deal with it yet. 😀

nerdalert (approved):
LGTM 🥇

PR #652, opened 2023-03-26
design: Add Nexodus network to our project terminology

russellb:
Standardize on how we refer to a network managed by Nexodus in documentation. Earlier in the project we referred to this as a "zone," but that construct is gone for now. At some point a network slicing construct may be reintroduced within organizations, but for now just use the more generic "network" terminology.


vishnoianil (approved):

chirino (approved):
LGTM.

PR #654, opened 2023-03-27
fix: filter down devices to what the current user should have access to see

chirino:
Adds some cucumber tests to verify list/get don’t show other user’s devices.


internal/handlers/device.go
```diff
@@ -43,14 +44,31 @@ func (api *API) ListDevices(c *gin.Context) {
...
 }
 
+func (api *API) DeviceIsVisibleToCurrentUser(c *gin.Context) func(db *gorm.DB) *gorm.DB {
+	return func(db *gorm.DB) *gorm.DB {
+		userId := c.Value(gin.AuthUserKey).(string)
+		// this could potentially be driven by rego output
+		if api.dialect == database.DialectSqlLite {
+			return db.Where("user_id = ? OR organization_id in (SELECT organization_id FROM user_organizations where user_id=?)", userId, userId)
+		} else {
+			return db.Where("user_id = ? OR organization_id::text in (SELECT organization_id::text FROM user_organizations where user_id=?)", userId, userId)
```
chirino, line 67: I assume a user should be able to see all the devices he owns and all the devices owned by orgs that he is member of.  If this should be jus the ones he owns.. then this list query gets much simpler.

nerdalert (approved):

PR #668, opened 2023-03-27
nexd: Improve help text for flags

russellb:
Update the help text for flags so that a more descriptive word is used to represent the value passed along with a string flag. For example, instead of the text saying:

```
    --child-prefix value
```

It now says:

```
    --child-prefix CIDR
```

This is the full updated help text for the flags that were updated:

```
   --public-key key                             Base64 encoded public key for the local host - agent generates keys by default (default: "key") [$NEXD_PUB_KEY]
   --private-key key                            Base64 encoded private key for the local host (dev purposes only - soon to be removed) [$NEXD_PRIVATE_KEY]
   --listen-port port                           Wireguard port to listen on for incoming peers (default: 0) [$NEXD_LISTEN_PORT]
   --request-ip IP                              Request a specific IP address from Ipam if available (optional) [$NEXD_REQUESTED_IP]
   --local-endpoint-ip IP                       Specify the endpoint IP address of this node instead of being discovered (optional) [$NEXD_LOCAL_ENDPOINT_IP]
   --child-prefix CIDR [ --child-prefix CIDR ]  Request a CIDR range of addresses that will be advertised from this node (optional) [$NEXD_REQUESTED_CHILD_PREFIX]
...
   --username string                            Username string for accessing the nexodus service [$NEXD_USERNAME]
   --password string                            Password string for accessing the nexodus service [$NEXD_PASSWORD]
```


russellb:
> This isn't quite right. It set the default value for each flag as a side effect.
> 
> ... `(default: "key")`

nevermind, this was just a mistake in the changes for one flag

nerdalert (approved):

PR #670, opened 2023-03-27
perf: use the wgctrl go module to dump the wg sessions.

chirino:
Reduces our dependency on the wg binary.

### DIFF
--- a/internal/nexodus/dump.go
+++ b/internal/nexodus/dump.go
@@ -1,13 +1,8 @@
 package nexodus
 
 import (
-	"bufio"
-	"errors"
-	"fmt"
-	"io"
-	"strconv"
-	"strings"
-	"time"
+	"github.com/nexodus-io/nexodus/internal/util"
+	"golang.zx2c4.com/wireguard/wgctrl"
 )
 
 // WgSessions wireguard peer session information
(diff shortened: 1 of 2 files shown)


internal/nexodus/dump.go
```diff
@@ -33,54 +29,25 @@ func ShowDump(iface string) (string, error) {
...
-			PublicKey:       publicKey,
-			PreSharedKey:    psk,
-			Endpoint:        endpoints,
-			AllowedIPs:      allowedIPs,
-			LatestHandshake: latestHandshakeTime.String(),
-			Tx:              tx,
-			Rx:              tr,
+			PublicKey:       peer.PublicKey.String(),
+			PreSharedKey:    peer.PresharedKey.String(),
+			Endpoint:        peer.Endpoint.String(),
```
chirino, line 45: notice all the .String() conversions here... should we change the WgSessions to preserve the types more?

nerdalert: Probably, there is a scenario where the peer.Endpoint() can be `none` but it just throws a debug error. Going to merge to get some burn in on this and we can followup in reconcileState with better error handling of types and how we want to act if at all for the (none) return scenario for Endpoint. Ty!

nerdalert (approved):

PR #675, opened 2023-03-28
Resolve the bug blocking suite.T().Parallel() on all tests in e2e

nerdalert:
- This was a residual bug from zone removal where CIDRs collapsed into prefix that overlapped between orgs.
- Instead of TunnelIP we now use the CID for deletions resolving the bug that prevented concurrent testing.
- Also removed the Cleanup() as I think that was an issue as well. The containers get torn down regardless when the parent context exits. I've noticed a flake in GH with Cleanup that had to have been from a premature teardown. Either way, removing them doesn't add overhead since it's all concurrent anyways.

### DIFF
--- a/integration-tests/integration_test.go
+++ b/integration-tests/integration_test.go
@@ -8,10 +8,6 @@ import (
 	"crypto/tls"
 	"encoding/json"
 	"fmt"
-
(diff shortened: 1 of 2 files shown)


chirino (approved):
LGTM

PR #678, opened 2023-03-30
feat: you no longer need to modify your /etc/hosts to run the nexodus…

chirino:
… service on your local machine.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -269,10 +269,6 @@ jobs:
           sudo apt update
           sudo apt-get -qy install libnss3-tools
 
-      - name: Add hosts to /etc/hosts
-        run: |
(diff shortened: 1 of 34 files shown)


russellb:
A couple of questions:

1) I'm pretty sure this will break development on a machine not connected to the internet. I suppose it's still possible by adding these entries to /etc/hosts, though?

2) Does the development environment still work properly when accessed remotely? At a quick glance, it seems to assume the only clients are local. Does the UI work from another host? Can nexd connect from a remote host?

chirino:
> 1. I'm pretty sure this will break development on a machine not connected to the internet. I suppose it's still possible by adding these entries to /etc/hosts, though?

Yep.  Stil possible.

> 2. Does the development environment still work properly when accessed remotely? At a quick glance, it seems to assume the only clients are local. Does the UI work from another host? Can nexd connect from a remote host?

The remote hosts would need to setup /etc/hosts entries.    But yes, they would.

chirino:
this is part of #684

PR #680, opened 2023-03-30
Add a nexd socket check for an existing process running

nerdalert:
- Add a check to prevent more than one nexd being initialized
- Added checks for OS on nexctl for unsupported OSs

### DIFF
--- a/cmd/nexctl/local.go
+++ b/cmd/nexctl/local.go
@@ -4,15 +4,16 @@ import (
 	"fmt"
 	"net"
 	"net/rpc/jsonrpc"
+	"runtime"
 
+	"github.com/nexodus-io/nexodus/internal/nexodus"
 	"github.com/urfave/cli/v2"
 )
 
 func callNexd(method string) (string, error) {
 	conn, err := net.Dial("unix", "/run/nexd.sock")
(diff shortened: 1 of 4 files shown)


nerdalert:
Leveraging Russell's excellent socket ctl to check if an existing nexd service is running on Linux. @russellb I ended up duplicating the callNexd function in nexodus instead of moving local.go out of nexctl to preserve `Version` numbers set at build. Ty. 

```
### Linux:

# nexctl nexd ... commands
$ sudo ./nexctl nexd version
nexctl version: dev
nexd version: dev

$ sudo ./nexctl nexd status
Status: Running

# Starting on Linux with an existing nexd process running

$ sudo ./nexd-dev-linux-amd64 --username kitteh1 --password floofykittens  https://try.nexodus.local
{"level":"fatal","ts":1680137512.865208,"caller":"nexd/main.go:205","msg":"existing nexd service already running"}

### Non-Linux Hosts:

$ sudo ./nexctl nexd version
nexctl version: dev
nexd ctl interface not currently supported on darwin

$ sudo ./nexctl nexd status
nexd ctl interface not currently supported on darwin
```

PR #682, opened 2023-03-30
fix: nexodus would fail on startup if it could not fetch the jwks certs

chirino:
We now fetch the jwk’s and the first API request, and retry again very 5 seconds if the fetch fails.


nerdalert (approved):
LGTM, ty!

PR #683, opened 2023-03-30
docs: Split up the user guide

russellb:
The `user-guide.md` file has become a bit large. It's already under its own sub-folder in the docs, `user-guide`. Split it apart into more focused topics as a more scalable way to organize this information as we continue to grow.

In this change, we now have separate files for these topics which were previously combined in `user-guide.md`.

- The nexd agent
- Discovery and Relay modes
- nexctl
- Subnet Routers


nerdalert (approved):
LGTM ty!

PR #685, opened 2023-03-31
Plumbing for userspace mode

russellb:
This commit introduces some necessary plumbing for the userspace-only proxy mode of nexd. There are several places in the code that call out to OS-specific methods. In all of these cases, we need to conditionally use a different approach if nexd is running in this userspace mode. All of these conditionals are including in this patch, but without the implementations, to make it an easier chunk to review.

The code in route_userspace.go is complete. There's nothing to do, at least for now. A comment in the code explains further.

The keepalive code is not OS specific, but it also doesn't work with userspace mode, so it's skipped for now. The full implementation will enable native wireguard keepalives as a fallback, at least for now.

The other piece of implementation left in this commit is the path for the generated public and private keys. The default paths don't work when running in an unprivileged mode. Instead, we just use the current working directory. There's probably a better place to store them, but this works for the moment.

Follow-up commits will introduce the implementations of all of these userspace-only methods, a simple L4 proxy implementation, and hooking it all together by exposing the functionality through nexd.

Partial implementation of https://docs.nexodus.io/development/design/userspace-mode/ for issue #187.


nerdalert (approved):
LGTM

PR #691, opened 2023-03-31
nexd: Add core handling for userspace mode

russellb:
This is the next step for userspace-only mode support after PR #685.

This commit introduces the core networking and wireguard code for the 
userspace-only mode of operation. There is no real user-visible
behavior introduced yet. It is not exposed via `nexd` yet, but you can 
test out the core functionality with an env var.

You can observe with the following invocation of `nexd`. Note that
`root` or `sudo` are not required here. It is intended to be used
without any elevated permissions.

```
NEXD_LOGLEVEL=debug NEXD_USERSPACE_MODE_TEST=true dist/nexd --username kitteh1 --password floofykittens https://try.nexodus.local
```

If you watch the output, you'll see the userspace stack get 
initiatlized. All of the same discovery and peering that `nexd`
normally does will occur. One exception is that I have not yet 
implemented keepalives for this mode in `nexd`. Instead, I have
wireguard's native keepalives turned on for every 5 seconds.

For a peer it has successfully completed a handshake with, you'll see 
this debug message when the keepalives occur:

```
2023-03-31T12:05:39.485-0400DEBUG   device/send.go:82   peer(JuO2…16GM) - Sending keepalive packet
```

If you start `nexd` normally on another device with DEBUG logging
enabled, you should be able to observe that the keepalives sent by
`nexd` are successful to this new peer.

```
2023-03-31T16:09:05.081ZDEBUG   nexodus/keepalive.go:43 peer [ 100.100.0.2 ] is reachable
```

A follow-up commit will introduce features that make use of this,
namely the L4 proxy support proposed in
`docs/development/design/userspace-mode.md`.

### DIFF
--- a/cmd/nexd/main.go
+++ b/cmd/nexd/main.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"os/signal"
+	"strconv"
 	"sync"
 	"syscall"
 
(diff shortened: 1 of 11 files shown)


russellb:
marked as a draft pending the merge of #685

nerdalert (approved):
LGTM!

PR #700, opened 2023-03-31
nexd: Add an L4 proxy mode

### LINKED ISSUE #187: Add a userspace networking mode for the agent

russellb:
This commit makes use of the ability to run entirely in userspace mode to implement basic L4 proxy functionality that requires no elevated privileges.

This mode can be used in a Kubernetes Pod where `nexd` is not able to create a new network device. It's possible to grant a pod the access it needs, but in many cases an application developer does not have this level of access available.

Multiple ingress or egress proxy rules may be specified at the same time, but they all must be specified as command line arguments. Runtime management of rules will come later.

Documentation on how to use this mode can be found in `docs/user-guide/nexd-proxy.md`. A preview of these docs can be found here: https://deploy-preview-700--nexodus-docs.netlify.app/user-guide/nexd-proxy/

Closes #187

### DIFF
--- a/.markdownlint-cli2.yaml
+++ b/.markdownlint-cli2.yaml
@@ -2,6 +2,7 @@ config:
   line-length: false
   no-emphasis-as-header: false
   first-line-heading: false
+  code-block-style: false
 globs:
  - "**/*.md"
 ignores:
--- a/cmd/nexd/main.go
+++ b/cmd/nexd/main.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"os"
 	"os/signal"
-	"strconv"
 	"sync"
 	"syscall"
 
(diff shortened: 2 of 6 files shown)


russellb:
The most glaring thing missing here is functional tests. I just ran out of time to get those written this week.  I will file issues for that, among other to-do items I have in mind.

vishnoianil (approved):

vishnoianil:
@nerdalert PR looks good to me, if you find any issue during your testing ping me, I can take a look.

PR #707, opened 2023-04-01
Fixup tests post-schema change and validate user deletes

nerdalert:
- Cleanup the commented e2e block with a user delete and validation


chirino (approved):
lgtm

PR #712, opened 2023-04-01
nexd: Support unix socket control interface on mac

### LINKED ISSUE #710: nexd: Add nexctl socket interface support on Mac
`nexctl nexd ...` doesn't work on Mac.

russellb:
The only issue with working on mac was using a different default path for the unix socket. Otherwise, it shares the same implementation as Linux.

Closes #710


internal/nexodus/ctlserver_unix.go
```diff
@@ -1,4 +1,4 @@
-//go:build linux
+//go:build unix
```
chirino, line 1: I think this can be simplified to:

    //+build linux darwin 

And then you don't need to makefile tricks.

internal/nexodus/ctlserver_unix.go
```diff
@@ -41,9 +43,17 @@ func (ax *Nexodus) CtlServerLinuxStart(ctx context.Context, wg *sync.WaitGroup)
 	})
 }
 
+func UnixSocketPath() string {
+	if runtime.GOOS == Darwin.String() {
+		return unixSocketPathDarwin
```
chirino, line 48: this could be a platform specific function.

internal/nexodus/ctlserver_unix.go
russellb: Ha. I tried to figure that out and tried lots of incorrect ways — comma separated, colon, semicolon …

internal/nexodus/ctlserver_unix.go
russellb: Yeah … but it doesn’t look like it’d make a meaningful difference to the binary size. I find this easier to follow than having to jump between files for this simple case. 

It’s probably worth a short design doc that lays out guidelines for the platform specific builds. I wouldn’t think it’s “make platform specific functions any time there is platform specific behavior,” is it? How about “split them when the implementation differs between platforms in non trivial ways and especially if platforms bring in different dependencies as this may have a meaningful  impact on binary size.”

PR #714, opened 2023-04-01
feat: add unix socket support for windows build

chirino:
add .gitattribute file to handle checking out code on Windows better. 
give .exe suffix to windows binaries.


nerdalert (approved):
LGTM!

PR #715, opened 2023-04-02
Remove explicit child-prefix route adds

nerdalert:
- Child prefixes are no longer handled separately from the allowed-ip host route additions. All allowed-ip peers are in the same slice passed to route add now.
- Build peers is for putting together the configuration, and deploy is for executing the peer config. This fixes that as well by removing any executions which was creating a race condition by trying to add a route in build() before the interface was created in deploy().


chirino:
@nerdalert you might want to review my attempt at conflict resolution.

nerdalert:
Thanks! This got resolved in the V6 merge.

PR #716, opened 2023-04-02
Remove the suite.fail from renNexd() in e2e

nerdalert:
- There isn't any output from the containerexec method so if there is a failure you simply get nil from err and a blank from output since that is a script getting piped to logs. The gather method collects logs for output and debugging. This changes it to just require the script is executed via containerexec.
- Also Im seeing a false positive on a test sourcing from this.


chirino (approved):
LGTM

PR #717, opened 2023-04-02
Flatten API responses and add more tests

chirino:
This adds more cucumber tests and it also 'flattens' the api responses.  Getting resource no longer joins the associated resources in the response.  This will allow us to have a more consistent data access cost per API request and make rate limiting work better.  This also simplifies the overall API.


nerdalert (approved):
LGTM!

PR #718, opened 2023-04-02
bump up to Go 1.20

chirino:


chirino:
@russellb think I'm going to need your help figuring out why this breaks the RPM build https://github.com/nexodus-io/nexodus/actions/runs/4591215284/jobs/8107361785?pr=718

chirino:
took a peek at `dist/rpm/mock/build.log` and it seems like it's using `go1.19.7` to compile.  How do you update the compiler for the rpm?

chirino:
Changed it up so the build is now go 1.19 source compatible, but we compile with 1.20 on CI except for the rpms.

nerdalert (approved):
LGTM

PR #719, opened 2023-04-02
Makefile: Fix generate to only run if necessary

russellb:
Don't run the `generate` target if no dependent code has changed since the last time it ran.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -81,7 +81,7 @@ dist/nexctl-%: $(NEXCTL_DEPS) | dist
 .PHONY: clean
 clean: ## clean built binaries
 	$(CMD_PREFIX) rm -rf dist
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -157,12 +157,15 @@ opa-fmt: ## Lint the OPA policies
 	$(CMD_PREFIX) docker run --platform linux/x86_64 --rm -v $(CURDIR):/workdir -w /workdir docker.io/openpolicyagent/opa:latest fmt --write $(policies)
 
 .PHONY: generate
-generate: gen-docs ## Run all code generators and formatters
+generate: .generate ## Run all code generators and formatters
+
+.generate: .gen-docs
```
chirino, line 162: don't you need the source file added as deps here so that it does kick off if you have any source modifications?

russellb: The source files are dependencies of `.gen-docs`. So, changed source causes `.gen-docs` to get rebuilt, which causes `.generate` to get rebui.d

PR #720, opened 2023-04-02
nexd: Error on flags incompatible with proxy

russellb:
Exit with an error if nexd is started with a top-level flag that is incompatible with `nexd proxy`.


chirino (approved):
lgtm

nerdalert (approved):

PR #721, opened 2023-04-02
Add nexrc.env for easy sourcing of nexodus URLs

russellb:
I didn't want to type `try.nexodus.127.0.0.1.nip.io` all the time, so I made this to keep in my environment. I figured others may want the same so I put it in the repo.


chirino:
This looks like an .env file, maybe .env is a better extension?

russellb:
> This looks like an .env file, maybe .env is a better extension?

Yeah, that's fine.  I'll change it

russellb:
updated to `nexrc.env`

PR #734, opened 2023-04-03
Separate agent and relay functions 

vishnoianil:
Current agent implementation takes care of the following four major function
*.* Kernel space wireguard agent to establish the peering 
*.* User space wireguard agent to support containers
*.* Relay agent to route traffic between nodes behind symmetric NAT
*.* Discovery node to discover the latest wiregaurd state and update the peers

Agent, Relay and Discovery are three different functions with their own complex scenarios to manage and we need to evolve these functions for future scenarios (better discovery, different driver support etc etc). This PR does the following high level changes (commits has more details about the changes).

1) Separate wireguard dataplane management related code in it's own internal package.
2) Refactor the apiserver communication code a bit so it can be used by both agent and relay
3) Separate the agent and relay struct and related methods cleanup
4) Add router and relay command and move the related flags.
5) Adopt the integration tests to new commands.

Current nexd commands looks like 

```
./dist/nexd -h
NAME:
   nexd - Node agent to configure encrypted mesh networking with nexodus.

USAGE:
   nexd [global options] command [command options] controller-url

COMMANDS:
   version  Get the version of nexd
   proxy    Run nexd as an L4 proxy instead of creating a network interface
   router   Enable child-prefix function of the node agent to enable prefix forwarding.
   relay    Enable relay and discovery support function for the node agent.
   help, h  Shows a list of commands or help for one command

GLOBAL OPTIONS:
   --help, -h  Show help

   Agent Options

   --relay-only         Set if this node is unable to NAT hole punch in a hub zone (Nexodus will set this automatically if symmetric NAT is detected) (default: false) [$NEXD_RELAY_ONLY]
   --stun               Discover the public address for this host using STUN (default: false) [$NEXD_STUN]
   --unix-socket value  Path to the unix socket nexd is listening against (default: "/run/nexd.sock")

   ApiServer Options

   --insecure-skip-tls-verify  If true, server certificates will not be checked for validity. This will make your HTTPS connections insecure (default: false) [$NEXD_INSECURE_SKIP_TLS_VERIFY]
   --password string           Password string for accessing the nexodus service [$NEXD_PASSWORD]
   --username string           Username string for accessing the nexodus service [$NEXD_USERNAME]

   Wireguard Options

   --listen-port port      Wireguard port to listen on for incoming peers (default: 0) [$NEXD_LISTEN_PORT]
   --local-endpoint-ip IP  Specify the endpoint IP address of this node instead of being discovered (optional) [$NEXD_LOCAL_ENDPOINT_IP]
   --private-key key       Base64 encoded private key for the local host (dev purposes only - soon to be removed) [$NEXD_PRIVATE_KEY]
   --public-key key        Base64 encoded public key for the local host - agent generates keys by default [$NEXD_PUB_KEY]
   --request-ip IP         Request a specific IP address from Ipam if available (optional) [$NEXD_REQUESTED_IP]
```

Testing Done:

- [x] Local integration tests is passing.
- [x] Tested with aws nodes + home nodes + rdu node setup.
- [x] Mac agent works fine.
- [x] Nexctl works fine.
- [x] wireguard userspace testing looks good. Followed the instructions (https://github.com/nexodus-io/nexodus/blob/main/docs/user-guide/nexd-proxy.md#demo-using-containers)
 
Todo:

- [ ] Windows agent testing
- [ ] Doc updates (will push patch once PR is ready for merge


chirino:
@vishnoianil needs a rebase.

russellb:
I see todo items. Do you want to mark this as a draft until you feel it’s ready for merge?

cmd/nexd/main.go
```diff
@@ -76,37 +75,62 @@ func nexdRun(cCtx *cli.Context, logger *zap.Logger, mode nexdMode) error {
...
 	}
-	if err := nex.Start(ctx, wg); err != nil {
+	if err := nexa.Start(ctx, wg); err != nil {
 		logger.Fatal(err.Error())
 	}
 	wg.Wait()
+}
 
-	return nil
+func runNexRelay(cCtx *cli.Context, ctx context.Context, logger *zap.Logger, controller string) {
```
russellb, line 108: I'd rather not split these, as it creates some duplicated code

cmd/nexd/main.go
```diff
@@ -163,6 +187,55 @@ func main() {
...
+				Action: func(cCtx *cli.Context) error {
+					if runtime.GOOS != Linux {
+						return fmt.Errorf("Relay node is only supported for Linux Operating System")
+					}
+
+					return nexdRun(cCtx, logger, nexdModeRelay)
+				},
+				Flags: []cli.Flag{
+					&cli.BoolFlag{
+						Name:     "enable-discovery",
```
russellb, line 225: So the intention is to allow relay-only or relay+discovery? and not discovery-only?

It looks like docs need to be updated for this as well, right?

cmd/nexd/main.go
```diff
@@ -171,89 +244,95 @@ func main() {
 				Usage:    "Base64 encoded public `key` for the local host - agent generates keys by default",
 				EnvVars:  []string{"NEXD_PUB_KEY"},
 				Required: false,
+				Category: "Wireguard Options",
```
russellb, line 247: perhaps add this at the top of the file to keep the string in one spot since it's repeated several times.

```go
const (
    wireguardOptions = "Wireguard Options"
)
```

integration-tests/integration_test.go
```diff
@@ -18,9 +18,10 @@ import (
 	"github.com/Nerzal/gocloak/v13"
 	"github.com/cenkalti/backoff/v4"
 	"github.com/cucumber/godog"
-	"github.com/nexodus-io/nexodus/internal/cucumber"
+    "github.com/nexodus-io/nexodus/internal/cucumber"
```
russellb, line 21: appears to be some broken whitespace changes here and elsewhere

internal/nexodus/wireguard/route_darwin.go
```diff
@@ -0,0 +1,90 @@
+//go:build darwin
```
russellb, line 1: how did you move these files? git seems to think the files were deleted and then created new. We need git to know that it's a move, not a delete/create.  Try `git mv`.

internal/nexodus/apiserver.go
```diff
@@ -0,0 +1,134 @@
...
+
+type ApiServer struct {
+	username      string
+	password      string
+	controllerURL *url.URL
+	ControllerIP  string
+	client        *client.Client
+	logger        *zap.SugaredLogger
+	skipTlsVerify bool
+}
```
russellb, line 25: Adding all of this code, but not removing any?  It would be good to combine that into the same commit to view it as a single change.

russellb (changes requested):
Just a partial review.

I would prefer that after this is rebased, it's resubmitted as a series of smaller changes. It will be easier to work through iterations of review in smaller chunks than this.

integration-tests/integration_test.go
chirino: `make generate` should fix this.

russellb:
@vishnoianil do you want to keep this open? I know some was submitted in other PRs.

vishnoianil:
> @vishnoianil do you want to keep this open? I know some was submitted in other PRs.

No, Closing it. Some of this work is already in, I have couple of more PR coming in for rest of the stuff.

PR #736, opened 2023-04-03
e2e: Log container exec before, not after the exec

russellb:
A log messsage about running a command in a container was misplaced. In an error case, it wouldn't be shown at all. In all cases, it didn't show up until after the command was already executed. This can be confusing when watching test execution.


nerdalert (commented):
LGTM

PR #737, opened 2023-04-03
e2e: Use types for sets of constants

russellb:
Make the handling of the v6 arg more explicit by making it a named type with two constants that represent possible values. It is also now a boolean instead of an int, since it was really a boolean in practice.

Make a similar change for how IP family is passed around as a string. There were only 2 valid values, so make it a set of typed constants.


nerdalert (approved):
LGTM

PR #745, opened 2023-04-04
e2e: Add nexd proxy e2e tests

### LINKED ISSUE #701: nexd proxy: Add integration tests
I ran out of time before PTO to implement integration tests that go along with PR #700. This issue is to make sure I come back to it.

russellb:
Add a test for `nexd proxy` ingress and egress rules. Both tests have two nodes, one running normal nexd and the other running nexd proxy. One side has an http server and the other runs curl. The roles switch based on which scenario is under test.

I added python3 and curl to the ubuntu test container for this scenario.

There's one hack in a test with a TODO comment to improve. It makes an assumption that our ipam is handing out IPs sequentially. We can't get the IP like we do normally because we can't read wg0 from the container since there's no interface. We could either add a new `nexctl nexd` command to get wireguard state, or we could query the nexodus service for the required device info. The hack at least works for now.

Closes #701

### DIFF
--- a/Containerfile.test
+++ b/Containerfile.test
@@ -57,6 +57,8 @@ RUN apt-get update -qy && \
     tcpdump \
     conntrack \
     psmisc \
+    curl \
+    python3 \
     && \
     apt-get clean
 COPY --from=build /src/dist/nexd /bin/nexd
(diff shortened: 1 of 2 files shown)


nerdalert (approved):
LGTM

PR #748, opened 2023-04-05
e2e: Add pings to proxy tests

russellb:
Add pings between the two nodes before exercising the nexd proxy functionality. This establishing whether basic L3 connectivity is established to differentiate between broken proxy handling or a failure prior to getting that far.

Also tweak the message when `curl` fails, as it's an expected condition if it starts too fast before peering has finished.


chirino (approved):
LGTM

PR #754, opened 2023-04-06
Makefile: Move target timestamp tracking to dist/

russellb:
Move all of the files used to track when a target last ran into dist/. This makes it so we don't have to update the `clean` target and `.gitignore` file for each new dotfile added.


chirino (approved):
lgtm

PR #756, opened 2023-04-06
feat: support storing api tokens between nexd runs so that you don’t …

chirino:
…need to redo the login flow.


nerdalert (approved):
Cool feature, LGTM.

Wonder if the flag should or description should mention the api token since state could mean a few things? Maybe even mention it can prevent having to re-auth in the description.

PR #757, opened 2023-04-06
e2e: Add two additional proxy test scenarios

russellb:
Prior to this commit, there were 3 scenarios:

- single ingress rule
- single egress rule
- 1 ingress rule + 1 egress rule

This commit adds the following to ensure that multiple rules of the same type also work as expected.

- multiple ingress rules
- multiple egress rules

A second commit in this PR changes how testcontainer names are generated so that the test name piece is automatic. This was inspired by having multiple copy/paste errors with the calls to createNode().

### DIFF
--- a/integration-tests/integration_test.go
+++ b/integration-tests/integration_test.go
@@ -9,9 +9,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


nerdalert (approved):
LGTM

PR #758, opened 2023-04-06
fix: `make all` would not settle.

chirino:


internal/api/public/api_invitation.go
```diff
@@ -193,7 +193,7 @@ func (a *InvitationApiService) CreateInvitationExecute(r ApiCreateInvitationRequ
 		return localVarReturnValue, nil, &GenericOpenAPIError{error: err.Error()}
 	}
 
-	localVarPath := localBasePath + "/api/api/invitations"
+	localVarPath := localBasePath + "/api/invitations"
```
russellb, line 196: is this intentional? It's not obvious how it's connected to the Makefile changes.

Makefile
```diff
@@ -147,13 +147,18 @@ dist/.openapi-lint: internal/docs/swagger.yaml | dist
...
+		-v $(CURDIR):/workdir -w /workdir \
+		ghcr.io/swaggo/swag:v1.8.10 \
+		/root/swag init $(SWAG_ARGS) -g ./cmd/apiserver/main.go -o ./internal/docs
+	$(CMD_PREFIX) touch $@
+
 internal/docs/swagger.yaml: dist/.gen-docs
 
 .PHONY: gen-openapi-client
-gen-openapi-client: dist\.gen-openapi-client ## Generate the OpenAPI Client
+gen-openapi-client: dist/.gen-openapi-client ## Generate the OpenAPI Client
```
russellb, line 161: lol ... I think I did this while editing from my phone trying to fix a CI failure.

internal/api/public/api_invitation.go
chirino: Yep. this is generated file.

PR #760, opened 2023-04-07
fix: sessions were being stored in apiserver memory

chirino:
We now store the sessions in redis so that sessions are not Invalidated on apiserver resort and so that your can scale the apiserver past 1 pod.


nerdalert (approved):
LGTM

PR #762, opened 2023-04-07
Add additional checks for relay nodes

nerdalert:
- For container scenarios, check if v4/v6 forwarding is enabled in the proc fs before trying to set it. Setting it will fail in a container and an error will be returned. But when spawned with: --sysctl net.ipv4.ip_forward=1 --sysctl net.ipv6.conf.all.forwarding=1 nexd will pass the settings check and not try and modify the kernel setting.
- Added v6 fwding to the v6 disabled test container, both can be true at the same time (ipv6 disabled with ipv6 fwding enabled).
- Removed relay from the v6 disabled test to keep a v6 disabled container profile.

### DIFF
--- a/integration-tests/integration_test.go
+++ b/integration-tests/integration_test.go
@@ -936,7 +936,7 @@ func (suite *NexodusIntegrationSuite) TestV6Disabled() {
 	})
 
 	// start nexodus on the nodes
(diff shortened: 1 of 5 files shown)


chirino (approved):
lgtm

PR #763, opened 2023-04-07
nexd: Set state-dir by default

russellb:
Having this feature on provides a better user experience. The option is still present if another directory is desired. The default for linux and mac is a pretty standard location for this type of thing: /var/lib/nexd/. The default for windows is the same directory we're using for other nexd things on Windows (C:/nexd).

### DIFF
--- a/cmd/nexd/main.go
+++ b/cmd/nexd/main.go
@@ -258,8 +258,8 @@ func main() {
 			},
 			&cli.StringFlag{
 				Name:    "state-dir",
-				Usage:   "Directory to store state in",
-				Value:   "",
+				Usage:   fmt.Sprintf("Directory to store state in, such as api tokens to reuse after interactive login. Defaults to'%s'", stateDirDefault),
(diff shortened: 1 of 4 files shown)


cmd/nexd/main.go
```diff
@@ -258,8 +258,8 @@ func main() {
 			},
 			&cli.StringFlag{
 				Name:    "state-dir",
-				Usage:   "Directory to store state in",
-				Value:   "",
+				Usage:   fmt.Sprintf("Directory to store state in, such as api tokens to reuse after interactive login. Defaults to'%s'", stateDirDefault),
+				Value:   stateDirDefault,
```
chirino, line 262: Isn't the default already shown?  Don't think your need the Sprintf.

chirino (approved):
LGTM

cmd/nexd/main.go
chirino: We can fix this in a follow up.

PR #764, opened 2023-04-08
Replaces iptables with nftables

nerdalert:
- also resolves an observed ip6tables issue in fedora containers

### DIFF
--- a/Containerfile.test
+++ b/Containerfile.test
@@ -17,7 +17,7 @@ FROM alpine:3.16 as alpine
 RUN apk add --no-cache \
     iputils \
     wireguard-tools \
-    iptables \
+    nftables \
     psmisc \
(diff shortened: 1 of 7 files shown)


nerdalert:
Output from the ec2 multinode deploy:
```
ip-172-31-83-131:~$ sudo nft list ruleset
table ip filter {
	chain FORWARD {
		type filter hook forward priority filter; policy accept;
		iifname "wg0" counter packets 249 bytes 20916 accept
	}
}
table ip6 filter {
	chain FORWARD {
		type filter hook forward priority filter; policy accept;
		iifname "wg0" counter packets 18 bytes 1872 accept
	}
}
```

chirino (approved):
LGTM.

PR #765, opened 2023-04-08
Makefile: Run gen-openapi-client before linters

russellb:
I got a failure on `make -j all` that shows a dependency issue where the linters ran before the api client code was in place:

```
WARN [runner] Can't run linter goanalysis_metalinter: buildir: failed to load package : could not load export data: no export data for "github.com/nexodus-io/nexodus/internal/api/public"
ERRO Running error: 1 error occurred:
        * can't run linter goanalysis_metalinter: buildir: failed to load package : could not load export data: no export data for "github.com/nexodus-io/nexodus/internal/api/public"
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -90,7 +90,7 @@ go-lint-prereqs:
 		exit 1 ; \
 	fi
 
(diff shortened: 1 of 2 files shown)


chirino (approved):
lgtm

PR #767, opened 2023-04-08
installer: Move to contrib dir

russellb:
The `hack` dir contains mostly developer oriented scripts. `contrib` has more user oriented things. Moving the installer script puts it next to the rpm and systemd files.


nerdalert (approved):
LGTM

PR #768, opened 2023-04-08
ci: use makefile for binary builds

russellb:
This ensures custom build flags in the Makefile are not omitted when builds are done in CI. I noticed after installing a binary from s3, the version was not set in the binary. This should fix it.


chirino (approved):
lgtm

PR #771, opened 2023-04-09
Makefile: Update generator dependency handling

russellb:
Update how make determines whether docs and code generators need to run.
By using the touched file method, these were running more than they
should. They should only run if their source is newer than their output.
Since these processes have both sources and outputs that are files, we
should be using them.

This fixes some errors I was having running parallel make. In
particular, `go mod tidy` was running at the same time as the client api
code generator, which would break. At least in this case, the code
generator shouldn't be running at all.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,8 @@ NEXODUS_RELEASE?=$(shell git describe --always --exclude qa --exclude prod)
 NEXODUS_LDFLAGS?=-X main.Version=$(NEXODUS_VERSION)-$(NEXODUS_RELEASE)
 NEXODUS_GCFLAGS?=
 
+SWAGGER_YAML:=internal/docs/swagger.yaml
+
 # Crunchy DB operator does not work well on arm64, use an different overlay to work around it.
 UNAME_M := $(shell uname -m)
 ifeq ($(UNAME_M),arm64)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -147,19 +149,16 @@ dist/.openapi-lint: internal/docs/swagger.yaml | dist
...
 		/root/swag init $(SWAG_ARGS) -g ./cmd/apiserver/main.go -o ./internal/docs
-	$(CMD_PREFIX) touch $@
-
-internal/docs/swagger.yaml: dist/.gen-docs
 
 .PHONY: gen-openapi-client
-gen-openapi-client: dist/.gen-openapi-client ## Generate the OpenAPI Client
-dist/.gen-openapi-client: internal/docs/swagger.yaml | dist
+gen-openapi-client: internal/api/public/client.go ## Generate the OpenAPI Client
+internal/api/public/client.go: internal/docs/swagger.yaml | dist
```
russellb, line 161: @chirino do you know if this is sufficient? In other words, if swagger.yaml got updated, it would always re-write client.go?

chirino: seems right

chirino (approved):
LGTM.

PR #772, opened 2023-04-10
feat: multiple endpoint address per device

chirino:
In the future this will allow us to better support connecting to the best endpoint address if you are behind multiple nat cones.


nerdalert (approved):
LGTM

PR #773, opened 2023-04-10
Replace custom keepalives with wg built-in persistant keepalive

nerdalert:
- Let wireguard manage keepalives. Set to a generally accepted best interval timer of 20s which is plenty to keep NAT bindings open on firewall/middleboxes. It is also one less thread nexd is maintaining. Would also make sense to catch any issues with it now rather than later. If it generates too much traffic in scale testing we can look to optimize as needed. Also it is more rigorously vetted in the wild with peer counts in the hundreds and beyond.

- Use the keepalive functions for a ctl function to probe for peer connectivity that the user can run. The function is batched so as not to fire off hundreds of simultaneous probes and to give priority to wg keepalives.

### DIFF
--- a/cmd/nexctl/main.go
+++ b/cmd/nexctl/main.go
@@ -74,6 +74,11 @@ func main() {
 				},
 			},
 			{
+				Name:   "connections",
+				Usage:  "Run a test of the nexd peer connectivity (host firewalls may block the ICMP probes)",
+				Action: cmdConnStatus,
+			},
+			{
 				Name:  "organization",
 				Usage: "Commands relating to organizations",
 				Subcommands: []*cli.Command{
--- a/internal/nexodus/ctlconnectivty.go
+++ b/internal/nexodus/ctlconnectivty.go
@@ -1,0 +1,106 @@
+package nexodus
+
+import (
+	"encoding/json"
+	"fmt"
+	"go.uber.org/zap"
+	"net"
+)
+
+const batchSize = 10
+
+func (ac *NexdCtl) Connectivity(_ string, keepaliveResults *string) error {
+	res := ac.ax.connectivityProbe()
+	var err error
+
+	// Marshal the map into a JSON string.
+	keepaliveJson, err := json.Marshal(res)
+	if err != nil {
+		return fmt.Errorf("error marshalling connectivty results")
+	}
+
+	*keepaliveResults = string(keepaliveJson)
+
+	return nil
+}
+
+func (ax *Nexodus) connectivityProbe() map[string]KeepaliveStatus {
+	if ax.userspaceMode {
+		// TODO: to be addressed in #703 'nexd proxy: Add support for nexd keepalives'
+		ax.logger.Warn("nexctl connection checks are currently unsupported for nexd user space proxy")
+		return make(map[string]KeepaliveStatus)
+	}
+
+	peerStatusMap := make(map[string]KeepaliveStatus)
+
+	if !ax.relay {
+		for _, value := range ax.deviceCache {
+			// skip the node sourcing the probe
+			if ax.wireguardPubKey == value.PublicKey {
+				continue
+			}
+			pubKey := value.PublicKey
+			nodeAddr := value.TunnelIp
+			if net.ParseIP(value.TunnelIp) == nil {
+				ax.logger.Debugf("failed parsing an ip from the prefix %s", value.TunnelIp)
+				continue
+			}
+			hostname := value.Hostname
+			peerStatusMap[pubKey] = KeepaliveStatus{
+				WgIP:        nodeAddr,
+				IsReachable: false,
+				Hostname:    hostname,
+			}
+		}
+	}
+	connResults := probeConnectivity(peerStatusMap, ax.logger)
+
+	return connResults
+}
+
+// probeConnectivity check connectivity in batches to limit excessive traffic in the case of a large number of peers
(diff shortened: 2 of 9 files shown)


nerdalert:
Output of the `nexctl connections` during a new join showing the convergence of connectivity. (ignore the binary name nexctl-linux in the video):


https://user-images.githubusercontent.com/1711674/230843564-4b6f8739-eb60-455c-8f15-4cc05708845c.mp4

internal/nexodus/ctlconnectivty.go
```diff
@@ -0,0 +1,104 @@
...
+			if ax.wireguardPubKey == value.PublicKey {
+				continue
+			}
+			pubKey := value.PublicKey
+			nodeAddr := value.TunnelIp
+			// strip the /32 from the prefix if present
+			if net.ParseIP(value.TunnelIp) == nil {
+				nodeIP, _, err := net.ParseCIDR(value.TunnelIp)
+				nodeAddr = nodeIP.String()
+				if err != nil {
```
russellb, line 42: This check should come right after `ParseCIDR()`. `nodeIP` can be `nil` here so you can't do `nodeIP.String()` unconditionally. I would just log an error and `continue`. In theory all of this info was validated before it went into the device cache so hopefully it never happens.

internal/nexodus/wg.go
```diff
@@ -111,7 +111,7 @@ func (ax *Nexodus) addPeerOS(wgPeerConfig wgPeerConfig) error {
 		Port: port,
 	}
 
-	interval := time.Second * 0
+	keepaliveInterval := time.Second * 20
```
russellb, line 114: Look above in `addPeerUS()`. Can you change where it is using 5 seconds to match this use of 20? I would probably make it a constant at the top of the file so it's not repeated and accidentally gets out of sync later.

You can also drop the TODO comment above it.

russellb: Note in that comment where it refers to docs that discuss keepalives. Those docs will need to be updated to reflect the changes in this PR.

cmd/nexctl/main.go
```diff
@@ -73,6 +73,11 @@ func main() {
 					return nil
 				},
 			},
+			{
+				Name:   "connections",
+				Usage:  "Run a test of the nexd peer connectivity (host firewalls may block the ICMP probes)",
```
russellb, line 78: Just a thought - it would be nice if we could query the keepalive status from wireguard instead of doing our own probes that are potentially dropped, but I don't see a way to do it.

russellb (changes requested):

russellb:
> Output of the `nexctl connections` during a new join showing the convergence of connectivity. (ignore the binary name nexctl-linux in the video):

nice!  Another way to show it instead of manually rerunning it:

```console
sudo watch nexctl connections
```

cmd/nexctl/main.go
nerdalert: Agree! I noodled on this a bit. I * think * we could using interface rx/tx tracking. Another upside would be if the endpoint has a host based firewall it could be blocking ICMP. The downside is we are inferring that if counters are incrementing bidirectionally things are working. I think its definitely worth a back log to revisit.

internal/nexodus/ctlconnectivty.go
nerdalert: Thanks! This was legacy when type checking was worse and the relay had a different value than the hosts. We can remove it completely.

internal/nexodus/ctlconnectivty.go
```diff
@@ -0,0 +1,104 @@
...
+	if err != nil {
+		return fmt.Errorf("error marshalling connectivty results")
+	}
+
+	*keepaliveResults = string(keepaliveJson)
+
+	return nil
+}
+
+func (ax *Nexodus) connectivityProbe() map[string]KeepaliveStatus {
```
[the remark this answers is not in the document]
nerdalert: Double check me here plx, I'm guilty of not testing with user space but I think this will cover it:

```go
	if ax.userspaceMode {
		// TODO: to be addressed in #703 'nexd proxy: Add support for nexd keepalives'
		ax.logger.Warn("nexctl connection checks are currently unsupported for nexd user space proxy")
		return make(map[string]KeepaliveStatus)
	}
```

vishnoianil:
@nerdalert I was planning to change it to wireguard keepalive as well, because although we probe, but we were not doing anything with reachable peers, until and unless we take any actions against non-rechable peer, i think custom probe is pretty much useless.

Regarding the connectivity check options through nexctl, I think it's good option for dev environment (and probably for kubernetes nodes as well), but i think for edge scenarios, it would be better if we can trigger this connectivity test through the apiserver, where user can click the button to run the connectivity test and it can show the list of all the reachable/unreachable peers in the GUI. wdyt?

nerdalert:
@vishnoianil I think both! We definitely need a health check scenario from the api-server. But the API-server isn't going to have a data path connection that would determine peers. It can check the http connection or poll token refreshes but wouldn't be able to say, spokeA can communicate with spokeB without something in the DP. I know you've thought through the scenarios but, just jotting it down in a reply for brevity. 

For NAT traversal debugging and general connectivity I cant think of anything other than a disaggregated approach for a full picture from a node's pov.

russellb:
@nerdalert @vishnoianil We're going to need a real observability story at some point. This kind of info would be great to report back to the apiserver. I'd love to have central visibility into what peering looks like (direct vs relay) and reachability (at some reasonable interval). Imagine what we could do UI-wise with all of that info.

russellb (approved):

PR #785, opened 2023-04-10
ci: Fix binary uploads

russellb:
A recent change got binary names out of sync with what the later upload step expected. This should fix it.


nerdalert (approved):
LGTM

Issue #548, opened 2023-03-06
Finish deployment automation


dave-tucker:
### Describe the Problem Statement

We would like to run expansive tests against our `qa` deployment...
Then if successful, promote those image SHAs to prod.

### Describe the Enhancement

1. Get ArgoCD to trigger the `e2e-qa` job via webhook when a deploy to the QA environment is complete. We may need to adjust `e2e-qa` to point to the QA deployment.
2. Once the `e2e-qa` job has passed, it should then trigger the `promote` job.

Once promote commits back to the repo to update our deployment, ArgoCD will deploy to prod.

Gregory-Pereira:
Hoping I could help on this one Brent if you havent already started.
/assign

Gregory-Pereira:
To achieve this I need a nexodus-io github admin to provision me a personal access token which will be used by argocd to trigger the e2e-qa workflow. The token should be correctly scoped to make this post request and trigger the workflow. It should be stored in the rosa vault instance. Github Workflow chaining should be implemented for part 2 to trigger the promote job.

PR #790, opened 2023-04-11
modify the qa-ec2 workflow for triggering via argcod notifications webhook

Gregory-Pereira:
Corresponds to: https://github.com/redhat-et/rosa-apps/pull/57
addresses: https://github.com/nexodus-io/nexodus/issues/548

This should enable the workflow to be triggered from a webhook on the nexodus side.


nerdalert (approved):
LGTM

PR #791, opened 2023-04-11
feat: add support to watch for device changes.

chirino:
This will eventually allow nexd to get the desired state from the apiserver more quickly at a lower cost.


nerdalert (approved):
LGTM, I beat this one up pretty well in QA just because its a beast.

- ec2 QA is g2g.
- Disconnects while flipping to a 5g tether and re-converging also looked good.
- local peering all looked good as well.
v nice

Makefile
```diff
@@ -160,15 +160,15 @@ $(SWAGGER_YAML): $(APISERVER_DEPS) | dist
 gen-openapi-client: internal/api/public/client.go ## Generate the OpenAPI Client
 internal/api/public/client.go: internal/docs/swagger.yaml | dist
 	$(ECHO_PREFIX) printf "  %-12s internal/docs/swagger.yaml\n" "[OPENAPI CLIENT GEN]"
-	$(CMD_PREFIX) rm -rf internal/api/public
+	$(CMD_PREFIX) rm -f $(shell find internal/api/public | grep .go | grep -v _custom.go)
```
russellb, line 163: If you want to go hardcore make ... untested and mostly for fun because I can:

```
$(filter-out %_custom.go,$(filter %.go,$(wildcard internal/api/public/*)))
```

Makefile
```diff
@@ -160,15 +160,15 @@ $(SWAGGER_YAML): $(APISERVER_DEPS) | dist
...
 	$(CMD_PREFIX) docker run --rm -v $(CURDIR):/src openapitools/openapi-generator-cli:v6.5.0 \
 		generate -i /src/internal/docs/swagger.yaml -g go \
 		--package-name public \
 		-o /src/internal/api/public \
 		-t /src/hack/openapi-templates \
 		--ignore-file-override /src/.openapi-generator-ignore $(PIPE_DEV_NULL)
 	$(ECHO_PREFIX) printf "  %-12s ./...\n" "[GO FMT]"
-	$(CMD_PREFIX) [ -z "$(shell gofmt -l .)" ] || (echo "$(shell gofmt -l .)" && exit 1)
+	$(CMD_PREFIX) [ -z "$(shell gofmt -l .)" ] || gofmt -w .
```
russellb, line 171: lol, not sure why i didn't think to check the exit code

russellb: actually, I was intentionally not writing to the files here to help avoid permissions problems. I wanted to check and exit with an error without ever touching the files. Is there a reason you needed to change the behavior?

russellb: ... talking to myself over here. I think it's still safe for CI how you have it. The original check doesn't write, and that's what I was trying to avoid. I'm ok with this.

chirino: Well we are generating files which are not formatted right so we need to follow up with a gofmt

chirino: yeah... CI passes. So I think it's safe.

Makefile
chirino: My make foo is weaker than my shell foo.

PR #793, opened 2023-04-11
ci: Speed up rpm builds in CI

### LINKED ISSUE #746: Speed up build-rpm CI job
The build-rpm job is very slow. After recent CI speed improvements, it's now the slowest job. It really shouldn't take this long. I (or someone) should explore ways to speed this up, but while still using `mock`, because that matches what's used in the real build servers on COPR.

We might be able to publish a test container to quay with a populated chroot cached that would speed this up a good bit. There are also several options in mock to explore.

russellb:
Cache most of the chroot used to build rpms in a container image that is prebuilt and pushed to quay.io.

Closes #746

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -528,16 +528,31 @@ cacerts: ## Install the Self-Signed CA Certificate
 dist/rpm:
 	$(CMD_PREFIX) mkdir -p dist/rpm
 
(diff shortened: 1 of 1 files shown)


russellb:
tweaking this a bit more ...

nerdalert (approved):
LGTM

PR #794, opened 2023-04-11
feat: support running individual cucumber feature tests

chirino:
Had to take move the feature tests out of the testify suite to support running each individually. 
Also noticed that e2e tests had a data race.  Enabled the race detector and refactored to avoid it.
Enabled e2e test liniting too.
Also improved messages logged in the e2e tests a bit, allowing you to see which source file is causing the log message.


russellb:
Can you document how to run the individual tests in `docs/developer/integration-tests.md` ?

nerdalert (approved):
LGTM

PR #795, opened 2023-04-11
Build images with latest GA fedora

russellb:
The `test` and `nexd` images were built on `fedora:36`, but `37` is the latest GA version. The `latest` tag always points to the latest GA, so just use that.

The mock container used `38`, but `latest` is a better choice and should work just fine.


chirino (approved):
lgtm

Issue #792, opened 2023-04-11
Database problems in qa env


chirino:
### Describe the bug

in the qa env, I'm seeing errors like this: 

```
2023-04-11T12:32:46.229Z DEBUG database/logger.go:72 INSERT INTO "user_organizations" ("user_id","organization_id") VALUES ('3f1d59aa-3424-4861-980b-2c1ef8fb1006','b4589bdb-4b56-44d6-9cb1-b8c8c35a1c64') ON CONFLICT DO NOTHING {"traceID": "4ae8aa7d2d582113b37e1c88804b5128", "line_number": "/src/internal/handlers/user.go:65", "error": "ERROR: insert or update on table \"user_organizations\" violates foreign key constraint \"fk_user_organization_organization\" (SQLSTATE 23503)", "rows": 0, "elapsed": 0.588042}
2023-04-11T12:32:46.230Z DEBUG database/logger.go:72 INSERT INTO "users" ("id","created_at","updated_at","deleted_at","user_name") VALUES ('3f1d59aa-3424-4861-980b-2c1ef8fb1006','2023-04-11 12:32:46.228','2023-04-11 12:32:46.228',NULL,'hchirino@redhat.com') {"traceID": "4ae8aa7d2d582113b37e1c88804b5128", "line_number": "/src/internal/handlers/user.go:65", "error": "ERROR: insert or update on table \"user_organizations\" violates foreign key constraint \"fk_user_organization_organization\" (SQLSTATE 23503)", "rows": 1, "elapsed": 1.613397}
2023-04-11T12:32:46.230Z ERROR zap@v0.1.0/zap.go:87 can't create user record: ERROR: insert or update on table "user_organizations" violates foreign key constraint "fk_user_organization_organization" (SQLSTATE 23503) {"status": 500, "method": "GET", "path": "/api/users", "query": "filter=%7B%7D&range=%5B0%2C9%5D&sort=%5B%22id%22%2C%22ASC%22%5D", "ip": "172.56.96.78", "user-agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/111.0.0.0 Safari/537.36", "latency": "4.147461ms", "time": "2023-04-11T12:32:46Z", "traceID": "4ae8aa7d2d582113b37e1c88804b5128"}
github.com/gin-contrib/zap.GinzapWithConfig.func1
[...]

PR #797, opened 2023-04-11
feat: enable soft deletes, add more db indexes

chirino:
* a deleted user can now be enabled again if he logs back into the API.


russellb:
related to issue #792

vishnoianil (approved):

PR #802, opened 2023-04-12
Fix local e2e test

vishnoianil:
While moving from test image to nexd image,
dist/nexd and dist/nexctl pre-req targets were
not added back. These binaries are needed to run
 the local e2e.

In Github CI job e2e was working because the job explictly build these binaries.

This patch adds the dist/nexd and dist/nexctl targets to e2e dependency, and removes the explict building of these binary in the build job, as it's not needed anymore.


chirino (approved):
Lgtm

PR #803, opened 2023-04-12
We do need to build binaries in build job

vishnoianil:
because it doesn't use make targets for running
integration test, it invokes directly.


vishnoianil:
YOLOing it, it's a partial revert from previous PR.

PR #806, opened 2023-04-13
Agent Refactor Part-1:Refactor the cli options to add router, relay c…

vishnoianil:
…ommands

Patch also

*.* moves some of the flag value checks in
respective flags actions.

*.* updated integration tests as per new commands

*.* updated documents to reflect the changes


nerdalert:
Ty @vishnoianil! If you don't mind rebasing to pickup the playbook changes (my fault there). When you rebase you can pull out any discovery as I removed it in the playbooks for `dev-ec2` workflows. 

Just parsing out the CLI here. I think Im good with the subcommands of [ proxy | router | relay ]. Let's talk about when to ditch `--discovery` in our standup.

Just for reference, Im pasting all the help outputs:

```
### router ###


nexd router --help
NAME:
   nexd router - Enable child-prefix function of the node agent to enable prefix forwarding.

USAGE:
   nexd router [command options] [arguments...]

OPTIONS:
   --child-prefix CIDR [ --child-prefix CIDR ]  Request a CIDR range of addresses that will be advertised from this node (optional) [$NEXD_REQUESTED_CHILD_PREFIX]
   --help, -h                                   Show help

### proxy ###

nexd proxy --help
NAME:
   nexd proxy - Run nexd as an L4 proxy instead of creating a network interface

USAGE:
   nexd proxy [command options] [arguments...]

OPTIONS:
   --ingress value [ --ingress value ]  Forward connections from the Nexodus network made to [port] on this proxy instance to port [destination_port] at [destination_ip] via a locally accessible network using a value in the form: protocol:port:destination_ip:destination_port. All fields are required.
   --egress value [ --egress value ]    Forward connections from a locally accessible network made to [port] on this proxy instance to port [destination_port] at [destination_ip] via the Nexodus network using a value in the form: protocol:port:destination_ip:destination_port. All fields are required.
   --help, -h                           Show help

### relay ###

nexd relay --help
NAME:
   nexd relay - Enable relay and discovery support function for the node agent.

USAGE:
   nexd relay [command options] [arguments...]

OPTIONS:
   --enable-discovery  Set if this node is to be the discovery node for NAT traversal in an organization (default: false) [$NEXD_ENABLE_DISCOVERY]
   --help, -h          Show help
```

Top level help:

```
nexd --help
NAME:
   nexd - Node agent to configure encrypted mesh networking with nexodus.

USAGE:
   nexd [global options] command [command options] controller-url

COMMANDS:
   version  Get the version of nexd
   proxy    Run nexd as an L4 proxy instead of creating a network interface
   router   Enable child-prefix function of the node agent to enable prefix forwarding.
   relay    Enable relay and discovery support function for the node agent.
   help, h  Shows a list of commands or help for one command

GLOBAL OPTIONS:
   --help, -h           Show help
   --unix-socket value  Path to the unix socket nexd is listening against (default: "/run/nexd.sock")

   Agent Options

   --relay-only  Set if this node is unable to NAT hole punch in a hub zone (Nexodus will set this automatically if symmetric NAT is detected) (default: false) [$NEXD_RELAY_ONLY]
   --stun        Discover the public address for this host using STUN (default: false) [$NEXD_STUN]

   ApiServer Options

   --insecure-skip-tls-verify  If true, server certificates will not be checked for validity. This will make your HTTPS connections insecure (default: false) [$NEXD_INSECURE_SKIP_TLS_VERIFY]
   --password string           Password string for accessing the nexodus service [$NEXD_PASSWORD]
   --state-dir value           Directory to store state in, such as api tokens to reuse after interactive login. Defaults to'/var/lib/nexd' (default: "/var/lib/nexd") [$NEXD_STATE_DIR]
   --username string           Username string for accessing the nexodus service [$NEXD_USERNAME]

   Wireguard Options

   --listen-port port      Wireguard port to listen on for incoming peers (default: 0) [$NEXD_LISTEN_PORT]
   --local-endpoint-ip IP  Specify the endpoint IP address of this node instead of being discovered (optional) [$NEXD_LOCAL_ENDPOINT_IP]
   --private-key key       Base64 encoded private key for the local host (dev purposes only - soon to be removed) [$NEXD_PRIVATE_KEY]
   --public-key key        Base64 encoded public key for the local host - agent generates keys by default [$NEXD_PUB_KEY]
   --request-ip IPv4       Request a specific IPv4 address from IPAM if available (optional) [$NEXD_REQUESTED_IP]
```

LGTM, my only question is around discovery, if we want to pull it out now or on a followup. Ty!

vishnoianil:
@nerdalert if you are comfortable with all your testing of the new discovery mechanism and feels like we can remove it, i can push a patch to clean that up.

>I think the only issue is if/when they overlap. I could see router and proxy conflicted at some point and not making it overly complicated.

This patch helps user space proxy a bit, as it doesn't support --child-prefix and with this change it doesn't have to explicitly check if user specified --child-prefix. But if we think user space proxy would probably support the --child-prefix (or for that matter any new option that agent will support), i think we can get rid of the router subcommand. @russellb wdyt?

vishnoianil:
Looking at the [userspace proxy proposal ](https://github.com/vishnoianil/nexodus/blob/main/docs/development/design/userspace-mode.md) it says `--child-prefix` won't be supported. But in general I am not sure if we would like to use userspace proxy to enable any router specific functionality, but if we do want to support, i think it's right time to discuss it :-). 

On the other hands, looking at the current options we support it looks reasonable to me, but not sure what all option we will support in future and that might push us to revisit it again. :)

nerdalert:
> @nerdalert if you are comfortable with all your testing of the new discovery mechanism and feels like we can remove it, i can push a patch to clean that up.

Comfortable is a bold statement 😆 We just haven't found the right combination of hardware/NAT that will break it! But since we don't have a plan for how to deploy discovery nodes in a SaaS model I think we are all in. The only thing that makes me comfortable with it is WebRTC and SIP have been hole punching via ICE for 15yrs.

nerdalert:
One other thing while discussing UX. If child prefix is in the context of `router` now, might be worth revisiting child-prefix. Not here, just throwing it out to get people thinking. We need to do a UX session fur sur:

```
nexd router --child-prefix 100.230.0.0/22 https://try.nexodus.io
```

Future thing to consider too would be when we forward hosts not on the node itself. E.g.all edge devices can forward through the node. 

Random IOS/EOS-like examples from my previous days as a ccie router jockey:

```
nexd router --advertise-prefix 100.230.0.0/22 https://try.nexodus.io
nexd router --originate-route 100.230.0.0/22 https://try.nexodus.io
nexd router --advertise 100.230.0.0/22 https://try.nexodus.io
nexd router --advertise-prefix 100.230.0.0/22 https://try.nexodus.io
nexd router --export 100.230.0.0/22 https://try.nexodus.io
nexd router --export-network 100.230.0.0/22 https://try.nexodus.io
nexd router --network 100.230.0.0/22 https://try.nexodus.io
nexd router --subnet 100.230.0.0/22 https://try.nexodus.io
```

nerdalert (approved):
LGTM

russellb:
> @nerdalert if you are comfortable with all your testing of the new discovery mechanism and feels like we can remove it, i can push a patch to clean that up.
> 
> > I think the only issue is if/when they overlap. I could see router and proxy conflicted at some point and not making it overly complicated.
> 
> This patch helps user space proxy a bit, as it doesn't support --child-prefix and with this change it doesn't have to explicitly check if user specified --child-prefix. But if we think user space proxy would probably support the --child-prefix (or for that matter any new option that agent will support), i think we can get rid of the router subcommand. @russellb wdyt?

I like having `nexd router`.

I do not think `nexd proxy` would ever support `--child-prefix`.  It's possible we may want to use the pure-userspace stack for routing between different wireguard networks (nexodus organizations or whatever subdivisions we add later), but I would expect that to be an option to `nexd router`. I'm not sure if that's even a useful feature, but it's possible. It would be an easy way to do some routing on a platform that wouldn't let us do it otherwise, I suppose.

cmd/nexd/main.go
```diff
@@ -5,24 +5,32 @@ import (
...
 	"github.com/nexodus-io/nexodus/internal/nexodus"
+	"github.com/nexodus-io/nexodus/internal/util"
 	"github.com/urfave/cli/v2"
 	"go.uber.org/zap"
 )
 
 const (
-	nexodusLogEnv = "NEXD_LOGLEVEL"
+	nexodusLogEnv    = "NEXD_LOGLEVEL"
+	Linux            = "linux"
```
russellb, line 21: Instead of redefining this, I would use the `Linux` constant already defined in the `nexodus` package.

cmd/nexd/main.go
```diff
@@ -5,24 +5,32 @@ import (
...
+	wireguardOptions = "Wireguard Options"
+	agentOptions     = "Agent Options"
+	apiserverOptions = "ApiServer Options"
```
russellb, line 24: Our document on naming conventions suggests we'd use "Nexodus Service Options" here.

PR #810, opened 2023-04-13
fix: `can't create user record: duplicated key not allowed` errors

chirino:
If a user issued multiple concurrent requests to the API server this  could occur if the user had never use the api before.


nerdalert (approved):
LGTM 🎉

PR #814, opened 2023-04-14
Exit the nexd agent in the scenarios where the agent node

vishnoianil:
itself doesn't show up in the peer list.

There are possible scenarios where we would like node agent to exit itself, and not attempt to auto onboard: 

-  when the apiserver db is wiped

- when user manually delete the device from org using either gui or api.

In both the above scenario, the device itself won't show in the peer listing of it's org, so this condition can be used to exit the nexd and avoid onboarding based on the stored apitoken.

### DIFF
--- a/cmd/nexctl/device.go
+++ b/cmd/nexctl/device.go
@@ -3,13 +3,14 @@ package main
 import (
 	"context"
 	"fmt"
-	"github.com/nexodus-io/nexodus/internal/api/public"
 	"log"
+
+	"github.com/nexodus-io/nexodus/internal/client"
 
 	"github.com/google/uuid"
 )
 
-func listOrgDevices(c *public.APIClient, organizationID uuid.UUID, encodeOut string) error {
+func listOrgDevices(c *client.APIClient, organizationID uuid.UUID, encodeOut string) error {
 	devices, _, err := c.DevicesApi.ListDevicesInOrganization(context.Background(), organizationID.String()).Execute()
 	if err != nil {
 		log.Fatal(err)
@@ -43,7 +44,7 @@ func listOrgDevices(c *public.APIClient, organizationID uuid.UUID, encodeOut str
 	return nil
 }
 
-func listAllDevices(c *public.APIClient, encodeOut string) error {
+func listAllDevices(c *client.APIClient, encodeOut string) error {
 	devices, _, err := c.DevicesApi.ListDevices(context.Background()).Execute()
 	if err != nil {
 		log.Fatal(err)
@@ -86,7 +87,7 @@ func listAllDevices(c *public.APIClient, encodeOut string) error {
 	return nil
 }
 
-func deleteDevice(c *public.APIClient, encodeOut, devID string) error {
+func deleteDevice(c *client.APIClient, encodeOut, devID string) error {
 	devUUID, err := uuid.Parse(devID)
 	if err != nil {
 		log.Fatalf("failed to parse a valid UUID from %s %v", devUUID, err)
--- a/internal/client/apiclient.go
+++ b/internal/client/apiclient.go
@@ -16,7 +16,10 @@ import (
 	"golang.org/x/oauth2"
 )
 
-type APIClient = public.APIClient
+type APIClient struct {
+	*public.APIClient
+	options *options
+}
 
 func NewAPIClient(ctx context.Context, addr string, authcb func(string), options ...Option) (*APIClient, error) {
 	opts, err := newOptions(options...)
@@ -118,7 +121,11 @@ func NewAPIClient(ctx context.Context, addr string, authcb func(string), options
 	}
 
 	clientConfig.HTTPClient = oauth2.NewClient(ctx, source)
-	return public.NewAPIClient(clientConfig), nil
+	apiclient := &APIClient{
+		public.NewAPIClient(clientConfig),
+		opts,
+	}
+	return apiclient, nil
 }
 
 type storeOnChangeSource struct {
@@ -171,4 +178,15 @@ func saveTokenToFile(path string, token *oauth2.Token) error {
 	}
 	defer f.Close()
 	return json.NewEncoder(f).Encode(token)
+}
+
+func (as *APIClient) ClearToken() error {
+	if as.options.tokenFile != "" {
+		if _, err := os.Stat(as.options.tokenFile); !os.IsNotExist(err) {
+			if err := os.Remove(as.options.tokenFile); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
 }
--- a/internal/nexodus/nexodus.go
+++ b/internal/nexodus/nexodus.go
@@ -56,7 +56,8 @@ const (
 )
 
 var (
-	invalidTokenGrant = errors.New("invalid_grant")
+	invalidTokenGrant   = errors.New("invalid_grant")
+	deviceNotRegistered = errors.New("device not registered")
 )
 
 // embedded in Nexodus struct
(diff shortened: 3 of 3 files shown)


vishnoianil:
Error it throws:
```
{"level":"fatal","ts":1681453838.7765238,"caller":"nexodus/nexodus.go:457","msg":"Device tzJyMeVGOvnuVTStO0Ex2svoxrNett7if1pVQafCzWg= is not registered with the organization dd265fe2-3889-4b3d-af72-6c3e9a93d5c8. Please re-register the device."}
```

russellb:
I don't understand the api token connection with a wiped database. The authentication credentials should remain the same, even if the agent needs to re-register itself as a new device. Forcing re-registration should be absolute last resort. Can you expand further on the explanation?

chirino:
This for the case where an org admin deletes a device.

vishnoianil:
> This for the case where an org admin deletes a device.

@chirino In sync-up you shared another scenario, in the context of recent production db wipe (although the solution is same), that was related to wiping the apiserver db, but keycloak db is persisting tokens so it might allow the device to fetch the peer listing from the apiserver but won't have the device registered for the org ?

chirino:
So in that context, yea after the wipe, it acts the same as the device being deleted by the user.  Nexd only registers the device on start up. So he needs to exit to re-register.

russellb:
> This for the case where a user deletes a device from the apiserver.

but if the agent is still running, I would expect it to recover itself. It can re-connect / recreate its own device. It should be self-healing a long as its running and has valid credentials.

chirino:
If a user goes in via the ui and deletes the device it would be weird if it automatically comes back.

russellb:
> If a user goes in via the ui and deletes the device it would be weird if it automatically comes back.

but that's the exact same thing that will happen if the daemon restarts. The only thing stopping it with this patch is removing the cached API token, but if it's using user/password auth, it's going to immediately recreate itself assuming it's running properly as a daemon.

It's also weird to delete a device that's still actively running! If you want an actively connected device to get blocked, I think that's a different feature we need, which I would envision as allowing per-device API keys that can be revoked.

chirino:
> but that's the exact same thing that will happen if the daemon restarts.

That's true.  Then again, I think the user/password flags on nexd should get dropped at some point.  It's not a great security story.  It would be better if nexctl could do that oauth flow with the user/password and then save the token for nexd.

vishnoianil:
> > If a user goes in via the ui and deletes the device it would be weird if it automatically comes back.
> 
> but that's the exact same thing that will happen if the daemon restarts. The only thing stopping it with this patch is removing the cached API token, but if it's using user/password auth, it's going to immediately recreate itself assuming it's running properly as a daemon.
Yes, this is good catch, and i think we can solve that as well. But i think first we need to agree whether we see it as a possible scenario to take care or not - device auto rejoins but it's not in the service db. I think it is. In this patch, we exit the nexd, but we don't really have to, we can keep nexd agent running, but doesn't allow it to rejoin, till user comes and intentionally kill it and rejoin. 
> 
> It's also weird to delete a device that's still actively running! If you want an actively connected device to get blocked, I think that's a different feature we need, which I would envision as allowing per-device API keys that can be revoked.
I probably need to understand more about why it's weird to delete the device. But in Edge scenarios where we have multiple devices, I think user can delete the device in case he wants to disconnect those devices from the network, or probably want to migrate it to other org. It's one click operation for admin, compared to login to each device and killing the agent. So in my opinion, deleting the device from UI seems like a valid use case until and unless i am missing something here. But if we feel that UI should not allow admin to delete the device from UI, i think we should enable that delete option for the device, only when it's inactive. 

Blocking the device is a new use case in my opinion, because blocking the device means, we don't even want the device to come onboard (because probably it went rouge) even if user correctly onboard it again. User need to come to UI and unblock it to onboard the device. @russellb or you see the blocking usecase from some other perspective?

russellb:
> > > If a user goes in via the ui and deletes the device it would be weird if it automatically comes back.
> > 
> > 
> > but that's the exact same thing that will happen if the daemon restarts. The only thing stopping it with this patch is removing the cached API token, but if it's using user/password auth, it's going to immediately recreate itself assuming it's running properly as a daemon.
> 
> Yes, this is good catch, and i think we can solve that as well. But i think first we need to agree whether we see it as a possible scenario to take care or not - device auto rejoins but it's not in the service db. I think it is. In this patch, we exit the nexd, but we don't really have to, we can keep nexd agent running, but doesn't allow it to rejoin, till user comes and intentionally kill it and rejoin.

I would rather `nexd` be fixed to auto-heal itself without having to restart.

> > It's also weird to delete a device that's still actively running! If you want an actively connected device to get blocked, I think that's a different feature we need, which I would envision as allowing per-device API keys that can be revoked.
> 
> I probably need to understand more about why it's weird to delete the device. But in Edge scenarios where we have multiple devices, I think user can delete the device in case he wants to disconnect those devices from the network, or probably want to migrate it to other org. It's one click operation for admin, compared to login to each device and killing the agent. So in my opinion, deleting the device from UI seems like a valid use case until and unless i am missing something here. But if we feel that UI should not allow admin to delete the device from UI, i think we should enable that delete option for the device, only when it's inactive.
> 
> Blocking the device is a new use case in my opinion, because blocking the device means, we don't even want the device to come onboard (because probably it went rouge) even if user correctly onboard it again. User need to come to UI and unblock it to onboard the device. @russellb or you see the blocking usecase from some other perspective?

If the intention is to block a device, I would expect that to be done by revoking its credentials. If we can create API tokens per device, this becomes more straight forward.

The reason I think deleting the device as an admin is weird is because  we're mixing up what's responsible for that record. Right now the device registers itself. If you delete it, I'd expect it to create itself again. If we wanted the admin to be able to delete the device, we should have the admin create it, too (which I don't think is a good UX).

If a device's credentials can be revoked so that you know it can no longer attempt to re-register itself, then it would make sense to go delete it.

vishnoianil:
>I would rather nexd be fixed to auto-heal itself without having to restart.

I think if we can clearly define the three events - onboarding, offboarding and Blocking, I think we can fix nexd to auto-heal until it's officially off-boarded, but currently i think we are bit vague around what defines these events (such as we don't really have any specific event for off-boarding, we just kill the agent or delete the device through UX/API). 

>If the intention is to block a device, I would expect that to be done by revoking its credentials. If we can create API tokens per device, this becomes more straight forward.

+1, but i think it doesn't take care of all the scenarios. Like if you delete the wireguard keys on the device, it will generate new wireguard keys and onboard it as a new device. So i think either we need an immutable device identity to implement it correctly (hint: FIDO), or assume that device is not manually accessible or admin accessibly only. 

>The reason I think deleting the device as an admin is weird is because we're mixing up what's responsible for that record. Right now the device registers itself. If you delete it, I'd expect it to create itself again. 

User firing `nexd` command is a manual trigger and the only intention of that command is to onboard the device. So i would argue that it's not really an auto register, but admin doing intentional onboard that result in device creation.

> If we wanted the admin to be able to delete the device, we should have the admin create it, too (which I don't think is a good UX).

I think thats where things get bit complicated with the UX and CLI options and we probably need to do more brainstorming here. We can not provide UX driven onboarding because devices are behind NAT mostly, so API server can't reach them (specifically in Edge cases), so onboarding is always need to be done through CLI by running agent/daemon, which i assume the device owner will do it. But we can control the device through the UX once it's onboarded (if we want to- like offboarding, blocking, moving). If I look at these events/action from device level granularity (and not db record level) than delete operation (or any other device related operation) doesn't look that weird, as far as action is taken by the device owner. 

In summary, I think as far as we can clearly define the actions/events that represent onboarding (eg. running nexd is onboarding event), offboarding (e.g deleting the device from GUI is not off-boarding), blocking (device is specifically marked to block) and other operation, i think we can have a  better control flow and defined behavior for these events.

russellb:
Sounds like a design doc is warranted here.

Issue #702, opened 2023-03-31
nexd proxy: Add UDP proxy support


russellb:
PR #700 included TCP support only. UDP support should be added next.

PR #818, opened 2023-04-14
nexd: Add egress UDP proxy support

russellb:
commit 431a691824d38f34b41a4ccd8e78cc17323f1b06
Date:   Thu Apr 13 13:48:50 2023 -0400

    nexd: Add more debug tools to container
    
    Add `nc` and a couple of custom UDP utilities to do the most trivial
    of UDP testing.
    
commit 2a57dc0155100e174e39ee043d1d8076b0b5c3a0
Date:   Fri Apr 7 09:35:15 2023 -0400

    nexd: Add udp egress proxy support
    
    This commit introduces UDP egress proxy support. I will add ingress
    support in a follow-up change.
    
    UDP proxy support is a bit more complicated than TCP support. We need
    to allow return traffic back to each originating client, so we have to
    do our own primitive connection tracking from within the proxy. These
    flows time out after 1 minute of inactivity.
    
    This timeout may need to be configurable at some point. I pulled 1
    minute from the default value used by Envoy's UDP proxy support.
    
    After a flow times out, subsequent traffic from a client will be seen
    as a new flow to the proxy and a new UDP source port will be allocated
    for the associated traffic forwarded from the proxy.
    
commit 0ce5e8625f0309c196141555b7888a2ab39db4f8
Date:   Fri Apr 14 15:20:53 2023 -0400

    e2e: Add TestProxyEgressUDP
    
    This test sets up a single egress UDP proxy rule. It ensures that a
    UDP packet makes it from a client, through the proxy, and to a server
    on another Nexodus device. It also ensures that a reply packet sent
    from that UDP server makes it back to the original client.
    
### DIFF
--- a/Containerfile.nexd
+++ b/Containerfile.nexd
@@ -16,6 +16,14 @@ RUN CGO_ENABLED=0 go build \
     -ldflags="-extldflags=-static" \
     -o nexctl ./cmd/nexctl
 
+RUN CGO_ENABLED=0 go build \
+    -ldflags="-extldflags=-static" \
+    -o udping ./hack/udping
+
+RUN CGO_ENABLED=0 go build \
+    -ldflags="-extldflags=-static" \
+    -o udpong ./hack/udpong
+
 FROM fedora:latest as fedora
 
 RUN dnf update -qy && \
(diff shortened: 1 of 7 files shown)


russellb:
Ingress UDP proxy support won't be too difficult to add as a follow-up, but it'll have to be next week.

chirino (approved):
LGTM

russellb:
Implements part of #702

PR #819, opened 2023-04-14
docs: Add proxy mode to docs home page

russellb:
Change "IP connectivity" to just "connectivity" since proxy mode allows more than just IP-level connectivity.

Add "L4 Proxy" as a line item on our home page to call out that while it's an L3 solution, there is also an L4 mode for non-privileged environments.


chirino (approved):
LGTM

PR #825, opened 2023-04-15
fix: ui errors model changes, feat: cross link associations better in…

chirino:
… detail pages.

* add a `make ui-fmt` target that is included in the `make generate` target.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -175,10 +175,18 @@ opa-fmt: ## Lint the OPA policies
 	$(ECHO_PREFIX) printf "  %-12s \n" "[OPA FMT]"
 	$(CMD_PREFIX) docker run --platform linux/x86_64 --rm -v $(CURDIR):/workdir -w /workdir docker.io/openpolicyagent/opa:latest fmt --write $(policies)
 
+
+.PHONY: ui-fmt
+ui-fmt: dist/.ui-fmt ## Format the UI sources
+dist/.ui-fmt: $(wildcard ui/*) $(wildcard ui/src/**) | dist
+	$(ECHO_PREFIX) printf "  %-12s \n" "[UI FMT]"
+	$(CMD_PREFIX) docker run --rm -v $(CURDIR):/workdir tmknom/prettier --write /workdir/ui/src/ >/dev/null
+	$(CMD_PREFIX) touch $@
+
 .PHONY: generate
 generate: dist/.generate ## Run all code generators and formatters
 
-dist/.generate: $(SWAGGER_YAML) | dist
+dist/.generate: $(SWAGGER_YAML) ui-fmt | dist
 	$(ECHO_PREFIX) printf "  %-12s \n" "[MOD TIDY]"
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -175,10 +175,18 @@ opa-fmt: ## Lint the OPA policies
 	$(ECHO_PREFIX) printf "  %-12s \n" "[OPA FMT]"
 	$(CMD_PREFIX) docker run --platform linux/x86_64 --rm -v $(CURDIR):/workdir -w /workdir docker.io/openpolicyagent/opa:latest fmt --write $(policies)
 
+
+.PHONY: ui-fmt
+ui-fmt: dist/.ui-fmt ## Format the UI sources
+dist/.ui-fmt: $(wildcard ui/*) $(wildcard ui/src/**) | dist
+	$(ECHO_PREFIX) printf "  %-12s \n" "[UI FMT]"
+	$(CMD_PREFIX) docker run --rm -v $(CURDIR):/workdir tmknom/prettier --write /workdir/ui/src/ >/dev/null
```
russellb, line 183: Could use `$(PIPE_DEV_NULL)`

ui/src/App.tsx
```diff
@@ -62,6 +62,13 @@ const App = () => {
         icon={OrganizationIcon}
         recordRepresentation={(record) => `${record.name}`}
       />
+      <Resource
+        name="organizations"
+        list={OrganizationList}
+        show={OrganizationShow}
+        icon={OrganizationIcon}
+        recordRepresentation={(record) => `${record.name}`}
+      />
```
russellb, line 71: This is a duplicate of the `Resource` above it. Or am I missing something?

ui/src/pages/Users.tsx
```diff
@@ -47,21 +34,28 @@ export const UserShow = () => (
     <SimpleShowLayout>
       <TextField label="ID" source="id" />
       <TextField label="Username" source="username" />
-      <ReferenceArrayField
-        label="Organizations"
-        source="organizations"
+
+      <ReferenceManyField
+        label="Owned Organizations"
```
russellb, line 39: Is there any difference between organizations owned vs. a member of?

russellb (changes requested):

ui/src/App.tsx
chirino: Doh!

ui/src/pages/Users.tsx
chirino: Yes.. You can delete / modify / send invitations for owned organizations.  We can't easily show member of relations since we would need the [ReferenceManyToManyField](https://marmelab.com/react-admin/ReferenceManyToManyField.html) react component and that's not open source.

russellb: Got it. Maybe we can do a custom component to get what we need.  (later, though)

russellb (approved):

PR #827, opened 2023-04-15
feat: add an invitations resource to manage org invitations.

chirino:
* added a GET api path to get a single invitation
* support creating an invitation by user name.


nerdalert (approved):
LGTM

PR #828, opened 2023-04-15
nexd: Set default service URL

russellb:
If nexd is run without a service URL, use a default URL that is set at
build time. If no `NEXODUS_BUILD_PROFILE` is set, it will use the `dev`
profile which will use the local dev service URL. The build profile can
be set to `qa` or `prod` which will set `qa.nexodus.io` or
`try.nexodus.io` as the default URLs.

For the `nexd` container image that gets published to quay.io and the
binary builds published to s3, they are now build with the `prod`
profile in CI.

For those using the rpm, update the default in the sysconfig file
to our public service. This would be the most likely config for an
install using the package.

This change also updates `nexctl` with the same behavior.

### DIFF
--- a/cmd/nexctl/main.go
+++ b/cmd/nexctl/main.go
@@ -24,6 +24,9 @@ const (
 // This variable is set using ldflags at build time. See Makefile for details.
 var Version = "dev"
 
+// Optionally set at build time using ldflags
+var DefaultServiceURL = "https://try.nexodus.io"
+
 var additionalPlatformCommands []*cli.Command = nil
 
 func main() {
@@ -40,7 +43,7 @@ func main() {
 			},
 			&cli.StringFlag{
 				Name:  "host",
-				Value: "https://api.try.nexodus.127.0.0.1.nip.io",
+				Value: DefaultServiceURL,
 				Usage: "Api server URL",
 			},
 			&cli.StringFlag{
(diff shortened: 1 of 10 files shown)


russellb:
I'm not sure why this PR triggered it, but I hit a specific failure on windows unit tests a few times in a row on here, so I pushed a fix for the failure to the same branch.

chirino:
Could we make our release builds change the defaultServiceURL to point at prod?  That way, if you do a local source build your get the local URL, but if you download the binaries from the website, they get defaulted to production?

russellb:
> Could we make our release builds change the defaultServiceURL to point at prod? That way, if you do a local source build your get the local URL, but if you download the binaries from the website, they get defaulted to production?

Yeah, I’ll do that.

chirino (approved):
LGTM

cmd/nexctl/main.go
```diff
@@ -24,6 +24,9 @@ const (
 // This variable is set using ldflags at build time. See Makefile for details.
 var Version = "dev"
 
+// Optionally set at build time using ldflags
+var DefaultServiceURL = ""
```
chirino, line 28: Could we set this to the production URL?  Just in case someone builds from source using `go install github.com/nexodus-io/nexodus/cmd/nexctl`

russellb: done

chirino (approved):

russellb:
@chirino this seems to be passing now

chirino (approved):
Lgtm

PR #829, opened 2023-04-16
Makefile: Ensure code gen runs before go-lint

russellb:
I got the following error on a build:

```
$ make -j all
  [OPENAPI CLIENT GEN] internal/docs/swagger.yaml
make: *** No rule to make target `/Users/rbryant/go/src/github.com/nexodus-io/nexodus/internal/api/public/model_models_add_invitation.go', needed by `dist/.go-lint-linux'.  Stop.
make: *** Waiting for unfinished jobs....
```

It looks like go-lint tried to run at the same time as the client code gen, and it caught it at a time when one of the generated files didn't exist. The fix included here has two parts:

* Make a default target for all of these generated files which depends on the target for `client.go`.

* Set the `client.go` target as an order-only dependency of the go-lint targets. The `openapi-client-gen` order-only pre-req was not sufficient, as the dependency checked was the go source. All generated source will now be seen as depending on the generation of `client.go`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -92,7 +92,7 @@ go-lint-prereqs:
 		exit 1 ; \
 	fi
 
-dist/.go-lint-%: $(NEX_ALL_GO) | go-lint-prereqs gen-docs dist gen-openapi-client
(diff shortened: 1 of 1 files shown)


chirino (approved):
LGTM

PR #831, opened 2023-04-16
feat: nexd now uses the watch api to get device change events.

chirino:
feat: add a Informer() api method that provides a locally cached device list that is synchronized with the apiserver using watch events.

* nexd now reconciles against informer and trigger new reconcile loops when the informer changes.

### DIFF
--- a/integration-tests/api_test.go
+++ b/integration-tests/api_test.go
@@ -196,3 +196,84 @@ func TestConcurrentApiAccess(t *testing.T) {
 	}
 
 }
+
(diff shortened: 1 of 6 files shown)


chirino:
@nerdalert ok this now has nexd using the watch apis..  nodes should converge quicker now.

nerdalert (approved):
LGTM. Tested:
- tethering scenario with ec2 nodes and a couple of remote PAT nodes.

PR #846, opened 2023-04-17
feat: build binary packages that include both nexd and nexctl with the user guide.

chirino:
* update the dashboard page to link to the new packages.


russellb:
I like including `nexd` + `nexctl`. I'm not sure about the user guide since the web presentation is so much nicer than the text.

Maybe we could drop a simple `README.md` in the root of the package that says "see https://docs.nexodus.io, but the source of the docs has been included here for local convenience."

Makefile
```diff
@@ -75,6 +75,37 @@ dist/nexctl-%: $(NEXCTL_DEPS) | dist
...
+
+dist/packages/%:
+
+	$(CMD_PREFIX) mkdir -p $@
+	$(ECHO_PREFIX) printf "  %-12s $@/nexd\n" "[GO BUILD]"
+	$(CMD_PREFIX) CGO_ENABLED=0 GOOS=$(word 1,$(subst -, ,$(shell basename $@))) GOARCH=$(word 2,$(subst -, ,$(shell basename $@))) \
+		go build -gcflags="$(NEXODUS_GCFLAGS)" -ldflags="$(NEXODUS_LDFLAGS)" -o $@/nexd$(if $(findstring windows,$@),.exe,) ./cmd/nexd
+	$(ECHO_PREFIX) printf "  %-12s $@/nexctl\n" "[GO BUILD]"
+	$(CMD_PREFIX) CGO_ENABLED=0 GOOS=$(word 1,$(subst -, ,$(shell basename $@))) GOARCH=$(word 2,$(subst -, ,$(shell basename $@))) \
+		go build -gcflags="$(NEXODUS_GCFLAGS)" -ldflags="$(NEXODUS_LDFLAGS)" -o $@/nexctl$(if $(findstring windows,$@),.exe,) ./cmd/nexctl
```
russellb, line 95: instead of duplicating the build commands, you could have this target depend on the corresponding dist/nexd and dist/nexctl targets. The commands here would just be a copy.

I can give it a try and send you a patch

chirino: not sure how to do that given this target is using %

russellb: did you see the PR i sent you?  I linked it 2 comments up.

chirino: I had not.. merged.  thx!

russellb:
needs a rebase now

russellb (changes requested):
I think if this changes the path and file type for binary uploads to s3, there may be a few more places that need updates:

- `docs/user-guide/agent.md`
- `contrib/nexodus-installer.sh`
- `ui/src/pages/Dashboard.tsx`

chirino:
> * `docs/user-guide/agent.md`

didn't see anything that needs to be changed.

> * `contrib/nexodus-installer.sh`

I've got some changes queued up for this.  But it's hard to test until the the packages are uploaded to S3.  Think we need to merge this before it's safe to change the installer script.

> * `ui/src/pages/Dashboard.tsx`

didn't see anything that needs to be changed.

russellb:
> > * `docs/user-guide/agent.md`

I think I had in mind changes would be needed to explain the updated behavior of the install script (that it installs 2 binaries, at least).
 
> didn't see anything that needs to be changed.
> 
> > * `contrib/nexodus-installer.sh`
> 
> I've got some changes queued up for this. But it's hard to test until the the packages are uploaded to S3. Think we need to merge this before it's safe to change the installer script.

As discussed out-of-band, I think we have outgrown this script and should remove it.

> 
> > * `ui/src/pages/Dashboard.tsx`
> 
> didn't see anything that needs to be changed.

Sorry. I was thinking about when there used to be install instructions embedded in that page. I'm the one that removed them, too.

russellb (approved):
fine with me, but we need to follow up on this quickly with the install script removal and updated docs that tell you where to download the latest binaries.

russellb (approved):
I'm not sure why the build workflow isn't running ...

russellb (changes requested):
@chirino See the attempted build workflow run for this error:

```
 Invalid workflow file: .github/workflows/build.yml#L196
The workflow is not valid. .github/workflows/build.yml (Line: 196, Col: 13): A mapping was not expected
```

russellb:
@chirino passing now!

docs/quickstart.md
```diff
@@ -43,23 +43,12 @@ sudo nexd https://try.nexodus.io
...
-- [Linux arm](https://nexodus-io.s3.amazonaws.com/linux-arm/nexctl)
-- [Mac x86-64](https://nexodus-io.s3.amazonaws.com/darwin-amd64/nexctl)
-- [Mac arm64 (M1, M2)](https://nexodus-io.s3.amazonaws.com/darwin-arm64/nexctl)
-- [Windows x86-64](https://nexodus-io.s3.amazonaws.com/windows-amd64/nexctl.exe)
+- [Linux x86-64](https://nexodus-io.s3.amazonaws.com/qa/nexodus-linux-amd64.tar.gz)
+- [Linux arm64](https://nexodus-io.s3.amazonaws.com/qa/nexodus-linux-arm64.tar.gz)
+- [Linux arm](https://nexodus-io.s3.amazonaws.com/qa/nexodus-linux-arm.tar.gz)
+- [Mac x86-64](https://nexodus-io.s3.amazonaws.com/qa/nexodus-darwin-amd64.tar.gz)
+- [Mac arm64 (M1, M2)](https://nexodus-io.s3.amazonaws.com/qa/nexodus-darwin-arm64.tar.gz)
+- [Windows x86-64](https://nexodus-io.s3.amazonaws.com/qa/nexodus-windows-amd64.zip)
```
russellb, line 51: Need to fix `.tar.gz` to `.tgz` or change the Makefile and build.yml over to `.tar.gz`. I'm taking a poll on Discord for preference!

russellb (commented):
I'm happy with this now and it seems to be working. Since I'm a co-author, someone else should probably approve.

docs/quickstart.md
nerdalert: May be worth considering adding the extraction command and copying it to a directory in $PATH. I'm on the fence, so not slowing it down, just throwing it out there!

nerdalert (approved):
LGTM

docs/quickstart.md
russellb: Yes, you're right. It's worth fixing this so the instructions are complete. I'll go ahead and push an extra patch to this branch for it.

russellb: @nerdalert ok, take another look. It has docs now

russellb (approved):

PR #857, opened 2023-04-18
Add .mailmap file

russellb:
This file is used by some git utilities to group commits that have different author addresses but are actually all the same person.


vishnoianil (approved):

nerdalert (approved):
LGTM

PR #859, opened 2023-04-19
Add scale testing for use against the QA deployment

nerdalert:
- Adds a workflow for running qa playbooks against the qa env.
- Most of the playbooks were reusable from dev-ec2. Only 2 new sets were made since self-signed certs isn't a thing in QA.
- Added a tool to add and delete users to a keycloak instance which generates unique UIDs by the same method as in e2e.
- Scale sizing is chosen at dispatch, which sets the Ansible vars from GH secrets.
- Verified there is no credential leakage in any output.

### DIFF
--- a/.github/workflows/qa-scale.yml
+++ b/.github/workflows/qa-scale.yml
@@ -1,0 +1,163 @@
+# This workflow is for scale testing against the Nexodus qa deployment
+name: qa-scale
+
+concurrency:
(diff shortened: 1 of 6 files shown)


russellb (changes requested):
Can you add a doc under `docs/development/` that explains how this should be used?

russellb (approved):

Issue #852, opened 2023-04-18
Set up Mergify


russellb:
In #815 , @dave-tucker suggested that I look at Mergify to replace the github action I added to do label enforcement. This issue is for setting that up.

While I'm at it, I'll look at using some of its other features.

russellb:
I can't replace the exact configuration currently used. Right now, a PR with the `hold` label gets a failing check status. Custom check statuses are a premium Mergify feature.

If we move to mergify to automate our merges based on criteria, that would replace the current check in a different way. It's a bit more work, though.

dave-tucker:
@russellb ```
pull_request_rules:
  - name: automatic merge for main with > 1 approved reviews, all requested reviews have given feedback and not held
    conditions:
      - "#approved-reviews-by>=1"
      - "#review-requested=0"
      - "#changes-requested-reviews-by=0"
      - base=main
      - label!=hold
```

Needs to also add "CI checks are green" to the list above :)

PR #861, opened 2023-04-19
Auto merge dependabot changes that pass CI

russellb:
Start experimenting with Mergify by creating an auto-merge rule for dependabot changes that pass CI.

A new "build-workflow-complete" job was added to capture the full completion of the build workflow. This makes configuration a lot simpler.

issue #852


dave-tucker (approved):
LGTM

PR #871, opened 2023-04-20
ui: Switch between light and dark theme logos

russellb:
When the UI theme is toggled between light and dark theme, switch which Nexodus logo we display on the dashboard.


nerdalert (approved):
Spun it up, LGTM

PR #872, opened 2023-04-20
README.md: Add Discord link

### LINKED ISSUE #847: Consider setting up chat for the project
We have github, but no good place for live chat discussion. We should pick one and set it up so interested parties have an easy place to jump in and discuss things.

Some options

* Slack - but the message limit on free plans is not great. We don't want to throw away important discussion history.

* Discord - doesn't have limited history like slack. not open source, but it's popular and familiar.

* Rocket.chat - or some other open source slack-like platform. I don't want to self host anything, but this one has a free hosted option: https://www.rocket.chat/pricing -- the free hosted option does have some limits

* Forget all of this and just embrace GitHub discussions as good enough: https://github.com/nexodus-io/nexodus/discussions

russellb:
Set up a place for live discussion. Discord was chosen as a free hosted option that does not limit history.

Closes #847

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,8 +5,10 @@
 
 [![build](https://github.com/nexodus-io/nexodus/actions/workflows/build.yml/badge.svg)](https://github.com/nexodus-io/nexodus/actions/workflows/build.yml)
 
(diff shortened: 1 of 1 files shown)


nerdalert (approved):
LGTM

PR #873, opened 2023-04-20
systemd: Fix autostart on boot

russellb:
The systemd service unit configuration was missing the Install section, which is required for doing `systemctl enable nexodus` to have it automatically started on boot.


nerdalert (approved):
LGTM

PR #874, opened 2023-04-20
rpm: Fix dependency installation

russellb:
When installing nexodus on a new host, nexd failed to start because wireguard-tools didn't get pulled in as a dependency. The reason is because the dependencies were in the wrong section of the spec file. With this change, wireguard-tools will be pulled in automatically as desired.


nerdalert (approved):
LGTM

PR #877, opened 2023-04-20
nexd: Exit if ctlserver can't start

russellb:
If nexd fails to open its socket for the ctlserver, it will now exit. Previously if it failed for some reason (like nexd was already running and listening on this socket), it would just contiously print the error.

If it fails after previously working, `nexd` will stay up and try to recover the ctlserver interface.


nerdalert (approved):
LGTM

PR #880, opened 2023-04-21
nexd: Detect misplaced arguments

russellb:
Arguments placed after the service URL are treated as additional
positional arguments which are all ignored. Detect this and return
with an error making it clear that the arguments were formed
incorrectly.


nerdalert (approved):
LGTM

Issue #885, opened 2023-04-21
Nexd State Machine


dave-tucker:
### Describe the Problem Statement

State management in nexd is becoming more complicated.
We need a mechanism to allow us to simplify that logic and not rely on crash/restart to retrigger things.

PR #886, opened 2023-04-21
design: Add nexd fsm design

dave-tucker:
Updates #885


docs/development/design/fsm.md
```diff
@@ -0,0 +1,57 @@
...
+
+To provide information to the user about the current state, a `nexctl nexd status` command will be added. This can inform the user of the current state and what to do next.
+
+To transition from IDLE to AUTH the "login" transition must occur.
+
+The trigger for this will be a CLI command: `nexctl nexd login`
+
+Once in the "AUTH" state, credentials will be taken from the user - or the device-auth flow will occur. In event of successful login the "login_success" transition occurs to bring nexd into the READY state. In case of "login_failed" we go back to IDLE.
+
+When in the READY state we will immediately issue a transition "start" into the PEERING state to preserve the existing behavior. The commands `nexctl nexd start` and `nexctl nexd stop` will be added to manually issue these transitions so users can choose to start/stop peering as necessary.
```
vishnoianil, line 46: I think the `stop` state is pretty risky. That basically means that the device is connected to the control plane, but it's not controlled by control plane in the `stop` state. So any changes in the peering/policies (which possibly doesn't allow that device to talk to some of the existing devices in-network or out-of-network anymore) won't be enforced, and that's not a good state. I think we should keep `login` and `logout` commands and ditch the `start` and `stop`. Login means you are onboarded and logout means you are offboarded.

docs/development/design/fsm.md
```diff
@@ -0,0 +1,57 @@
...
+    AUTH --> READY: "login_success"
+    AUTH --> IDLE: "login_failed"
+    READY --> PEERING: "start"
+    READY --> IDLE: "logout"
+    PEERING --> READY: "stop"
+    PEERING --> READY: "peering_error"
+    PEERING --> IDLE: "authentication_error"
+    PEERING --> IDLE: "logout"
+```
+
```
vishnoianil, line 30: I think this FSM is assuming that all the states will be forced through nexctl. I think we need to consider UX actions as well in this FSM, until and unless we assumes that UX is just real only. 

Just focusing on the device end-to-end workflow, i think we can take two possible action from UX
1) Delete device --> Map it to nexd logout
2) Block device  (doesn't exist yet) --> Keep the device mark in apiserver db, invalidate the token --> `nexd logout` and move to idle state. Further attempt by user to onboard the same device will be rejected.

docs/development/design/fsm.md
dave-tucker: > That basically means that the device is connected to the control plane, but it's not controlled by control plane in the stop state. 

No that's not the intention. "stop" is not the state. READY is the state.
In READY you are "ready" to connect to the control plane and start peering - you are not connected to the control plane and there is no dataplane setup.

> I think we should keep login and logout commands and ditch the start and stop. Login means you are onboarded and logout means you are offboarded.

Think of the following user story though:

As a Nexodus user, I want to pause my mesh VPN connection to perform a task, then reconnect afterwards without having to reauthenticate with the control plane - i.e you're working on a local VM whose IP address range overlaps with that on the mesh.

Forcing a user through the login/logout flow is suboptimal IMO.

docs/development/design/fsm.md
dave-tucker: > I think this FSM is assuming that all the states will be forced through nexctl.

It is not. "authentication_error" or and "peering_error" are pushed by next in response to API errors.
401 = authentication error, and 5XX but not the slow down error code would be "peering_error".

If we had bi-dir RPC we'd have new transitions for "deleted" and "blocked" which would put trigger the FSM to move to the appropriate state.

docs/development/design/fsm.md
```diff
@@ -0,0 +1,57 @@
...
+nexd may be in one of four states:
+
+1. IDLE: No configuration has been provided
```
chirino, line 33: Maybe a more descriptive state would be `LOGGED_OUT`

docs/development/design/fsm.md
```diff
@@ -0,0 +1,57 @@
...
+2. AUTH: Performing authentication, either interactively or via token
+3. READY: Authenticated and ready to peer
```
chirino, line 35: How about `LOGGED_IN`

docs/development/design/fsm.md
chirino, line 46: Another option if you look at it from bringing network interfaces up and down would be: `nexctl nexd up` and `nexctl nexd down`

docs/development/design/fsm.md
```diff
@@ -0,0 +1,57 @@
...
+4. PEERING: Actively Peering
+
```
chirino, line 37: Not sure if these states resonate better or not:

1. DISCONNECTED: No configuration has been provided, not logged in
2. CONNECTING: Performing authentication, either interactively or via token
3. DOWN: Authenticated and ready to bring the wg interface up
4. UP: wg interface is up and is peering.

docs/development/design/fsm.md
chirino: Also think we should not in the design that doing a `nexctl nexd up` or `nexctl nexd start` implicitly does a `nexctl nexd login` if your not logged in yet.

docs/development/design/fsm.md
vishnoianil: I think we need to define the parallel fsm for control plane as well. Device related changes that are triggered through UX actions by can be pushed to nexd as a device state changes, that nexd can listen to and update it's local FSM.

docs/development/design/fsm.md
vishnoianil: >No that's not the intention. "stop" is not the state. READY is the state.
In READY you are "ready" to connect to the control plane and start peering - you are not connected to the control plane and there is no dataplane setup.

Okay let me try to clarify my picture from control plan perspective. 
When device moves from IDLE->AUTH state, does control plane see the device ? 
When device moves from PEERING -> READY as a result of `stop` command, does that means the agent cleanup it's peering configuration and disconnect from all the peers?

>As a Nexodus user, I want to pause my mesh VPN connection to perform a task, then reconnect afterwards without having to reauthenticate with the control plane - i.e you're working on a local VM whose IP address range overlaps with that on the mesh.

As far as `stop` means that it clears it's local peering state and disconnect from others in the mesh network, and control plane sees that device is gone and update the other devices in the network about that, its seems reasonable to me. 

But if "pause" means it still maintain the connection with other peers, where other peers are still connected and sending packets in, while it edits it's local state, i am not very convenience it's a practical use case and seems like a bad violation of zero trust. 

Just to make sure i am putting my point across clearly, just throwing my local state to action map here 

login --> control plan detect the device, no peering information sent to new and existing device, token stored by new device

start --> control plan send initial peering information to the new device, also sends updated peering information to existing devices in network,

stop -> Agent remove the device from control plan, clear local peering information, control plan sends updated peering information to existing devices in the network

logout -> Control plan invalidate login token, next removes locally stored api token.

vishnoianil: >Another option if you look at it from bringing network interfaces up and down would be: nexctl nexd up and nexctl nexd down it would also avoid confusion from the user thinking your starting or stopping nexd.

But I think It creates another level of confusion because interface up doesn't really mean that it's connected to the network (if we really associated `up/down` with interface up/down). Another option can be `nexctl nexd join`, `nexctl nexd leave`.

chirino: Ok... sounds sound like maybe we need more states? How about: 

DISCONNECTED: No configuration has been provided, not logged in
CONNECTING: Performing authentication, either interactively or via token
UNREGISTERED: Authenticated but devices is not registered with the control plane
REGISTERED: Authenticated and registered with the control plane
DOWN: Registered but the wg interface is not up yet
UP: wg interface is up and is peering.

docs/development/design/fsm.md
russellb: The control plane does not need a fsm. CRUD with validations should be enough for the API.

docs/development/design/fsm.md
russellb: I like `DISCONNECTED` and `CONNECTING.

I like `READY` and `PEERING` more than `DOWN` and `UP`.

It seems like there should be one other state after `PEERING` is complete? Once all peer changes have been applied, I think it would be nice to go to a steady state like `UP`. If it receives updates that need to be applied, then it could go back into `PEERING` while it works on applying those changes.

docs/development/design/fsm.md
```diff
@@ -0,0 +1,57 @@
...
+To provide information to the user about the current state, a `nexctl nexd status` command will be added. This can inform the user of the current state and what to do next.
```
russellb, line 38:
```suggestion
To provide information to the user about the current state, a `nexctl nexd status` command will be updated to include the new states. This can inform the user of the current state and what to do next.
```

docs/development/design/fsm.md
```diff
@@ -0,0 +1,57 @@
...
+
+To transition from IDLE to AUTH the "login" transition must occur.
+
+The trigger for this will be a CLI command: `nexctl nexd login`
```
russellb, line 42: Why wouldn't this remain automatic like it is now? What is the value of the new `nexctl` command? Is it too allow providing a username/password dynamically?  Or?

docs/development/design/fsm.md
russellb, line 46: The value of the start/stop commands isn't clear to me. How is it better than `systemctl start` and `systemctl stop` (or the equivalent for whatever system you're on) ?

I saw mention of having to go back through the auth flow, but that's not really the case since we've either got a cached API token from interactive auth or we have user/password. Either way, stopping/starting shouldn't be a big deal, and it's the workflow I would expect based on how most services work.

Even if we did have start/stop, I would expect people actually stop/start the daemon.

docs/development/design/fsm.md
```diff
@@ -0,0 +1,57 @@
...
+In the PEERING state, we will be performing the reconciliation loop between desired peering state + current tunnel state. Many things can occur here:
+
+1. We may encounter an error (server not reachable) and issue a "peering_error" transition to go back to the READY state. Backoff/Retries can easily be added here to periodically issue a "start" transition when we've arrived at READY this way.
+1. We may encounter an error with authentication - i.e a revoked access token. This will issue an "authentication_error" transition that takes us back to the IDLE state, pending user interaction.
+1. A user may issue the `nextl nexd logout` command that will return us to the IDLE state via a "logout" transition
+1. A user may issue `nexctl nexd stop` command to return us to the `READY` state.
+
+## Alternatives Considered
+
+Unfortunately I don't see any other alternatives here.
```
russellb, line 57: not good ones.  :)

russellb (changes requested):

docs/development/design/fsm.md
dave-tucker: States have been renamed thanks.

dave-tucker: Added better descriptions to the states to help clarify control/data plane connections.

dave-tucker: > The value of the start/stop commands isn't clear to me. How is it better than systemctl start and systemctl stop (or the equivalent for whatever system you're on) ?

Because it allows us a better UX.

On any system with a GUI we're likely going to want a toolbar component and to send notifications to a user.
That toolbar component will want a start/stop button to enable/disable the mesh. fork/exec out to systemctl or the OS'equivalent isn't a particularly elegant or portable way to control our daemon process vs. using our own API.
Screenshot 2023-05-02 at 13 58 52
(see: Tailscale as reference)

Secondly, the control plane connection is used for far more than just getting peers, for example, getting a users current organization memberships, devices etc... We should support cases where a user wants to stop the mesh, browse the list of orgs they are part of, and chose to move their device to another organization

> I saw mention of having to go back through the auth flow, but that's not really the case since we've either got a cached API token from interactive auth or we have user/password. 

Well auth token handling is a topic for another design doc, but point taken. The comment was specifically on recommending a user to use `login/logout` to achieve the use case I'd mentioned - pause VPN while they do something on an overlapping IP address range, or don't want their corporate overlords seeing their packets 😆

russellb: OK - thanks for adding some more detail. Your example here was incredibly helpful:

> Secondly, the control plane connection is used for far more than just getting peers, for example, getting a users current organization memberships, devices etc... We should support cases where a user wants to stop the mesh, browse the list of orgs they are part of, and chose to move their device to another organization

I think I'm OK with adding this, as long as it's not a hard requirement at startup. In other words, just starting the service should automatically push it to login, assuming it has credentials, or to begin the OTP-login flow. That way, I see the added value while also not making anything harder than it is today.

docs/development/design/fsm.md
```diff
@@ -0,0 +1,81 @@
...
+1. REGISTERED: Device is registered with the control plane, but data plane is not established
+1. PEERING: Device is connected to control plane and fetching peer listings
+1. UP: Data plane is established
+
+To provide information to the user about the current state, a `nexctl nexd status` command will be updated to include the new states. This can inform the user of the current state and what to do next.
+
+To transition from UNAUTHENTICATED to AUTHENTICATING the "login" transition must occur.
+
+The trigger for this will be a either:
+- A CLI command, for example: `nexctl nexd login`
```
russellb, line 60: Need a blank line above the list to make md-lint happy

docs/development/design/fsm.md
```diff
@@ -0,0 +1,81 @@
...
+- Automatic if credentials were provided
```
russellb, line 61: How about:

- Automatic at startup, either using provided credentials, or by starting the OTP login flow
- Automatic any time `UNAUTHENTICATED` was reached not as the result of an explicit `logout` request from the admin.

This also means we need to store this state somewhere.  Imagine an admin requests `logout`, and then the service restarts for some reason (say, the package got upgraded). We want `nexd` to remember that `logout` was the last requested state.

docs/development/design/fsm.md
```diff
@@ -0,0 +1,81 @@
...
+    REGISTERED --> PEERING: "fetch_peers"
+    PEERING --> UNREGISTERED: "down"
+    PEERING --> UNREGISTERED: "peering_error"
+    PEERING --> UNAUTHENTICATED: "authentication_error"
+    PEERING --> UNAUTHENTICATED: "logout"
+    PEERING --> REGISTERED: "stop"
+    PEERING --> UP: "peered"
+    UP --> UNAUTHENTICATED: "logout"
+    UP --> PEERING: "fetch_peers"
+    UP --> REGISTERTED: "stop"
```
chirino, line 42: `REGISTERTED` typo?  Did you mean `REGISTERED`

chirino:
So the difference between UP and PEERING is that UP has not yet fetched the peer list? Are we thinking we could get stuck in the UP state and not transition to PEERING due to errors accessing the API server?

russellb:
> So the difference between UP and PEERING is that UP has not yet fetched the peer list? Are we thinking we could get stuck in the UP state and not transition to PEERING due to errors accessing the API server?

`UP` to me was the steady state when everything is configured and working properly.  `PEERING` is that peer configuration changes have been received and `nexd` is still working through processing them.

dave-tucker:
> UP to me was the steady state when everything is configured and working properly. PEERING is that peer configuration changes have been received and nexd is still working through processing them.

Yup. So the current reconcile loop that we have would be:

1. Check that we're in UP state
1. Fetch latest batch of peer changes
1. If peers have changed, trigger "PEERING" state, then (hopefully) back to UP
1. If not, stay in UP state

This also helps us with scaling, since we're effectively buffering peer changes that arrive when we're PEERING.

russellb:
I filed #960 to track and work on the e2e test failure that's happened here a couple times.

docs/development/design/fsm.md
```diff
@@ -0,0 +1,82 @@
...
+    AUTHENTICATED --> UNAUTHENTICATED: "logout"
+    REGISTERING --> REGISTERED: "device_registered"
+    REGISTERING --> UNREGISTERED: "registration_failed"
+    REGISTERED --> UNAUTHENTICATED: "logout"
+    UNREGISTERED --> REGISTERING: "register_device"
+    UNREGISTERED --> UNAUTHENTICATED: "logout"
+    REGISTERED --> UP: "up"
+    REGISTERED --> UNAUTHENTICATED: "logout"
+    REGISTERED --> PEERING: "fetch_peers"
+    PEERING --> UNREGISTERED: "down"
```
russellb, line 34: Would going to `UNREGISTERED` imply removing its own device record?

I don't think we'd want to remove it, since we'd want the device to keep its IPs. In that case `UNREGISTERED` doesn't seem quite right.

Maybe it should be: PEERING --> REGISTERED? If the device is administratively down, it would prevent a transition out of REGISTERED.

... but then REGISTERED doesn't seem to capture the state very well, either.  Maybe we need a `DOWN` that is almost identical to REGISTERED, but some transitions are blocked until it's administratively up again?

dave-tucker: good points. I've added a DOWN state that is like REGISTERED but can only be entered via admin action (i.e stop).
I've stopped anything transitioning to UNREGISTERED, so we'll never unregister the device.
In future we might want to offer the option to do so at some point, but i don't see a use case for it.

The only way out of DOWN is via a "fetch_peers" transition which will eventually get you back into UP - unless anything bad happens along the way.

russellb (approved):

PR #889, opened 2023-04-21
doc: document the design around the watch api

chirino:


docs/development/design/watch-api.md
```diff
@@ -0,0 +1,191 @@
+# Nexodus Device Watch API Implementation Notes
+
+## The Problem
+
+In order to connect or disconnect devices from a Nexodus managed organization, the list of all desired devices in the organization need replicated to all devices in the organization.  The `ListDevicesInOrganization` operation provides this list and can be accessed via the following REST call:
```
russellb, line 5:
```suggestion
In order to connect or disconnect devices from a Nexodus-managed organization, the list of all desired devices in the organization need to be replicated to all devices in the organization.  The `ListDevicesInOrganization` operation provides this list and can be accessed via the following REST call:
```

docs/development/design/watch-api.md
```diff
@@ -0,0 +1,191 @@
...
+
+`GET /api/organizations/{organization_id}/devices`
+
+The `nexd` agent would need to periodically poll this endpoint to get the current desired state from the API server and reconcile the device configuration to add or remove Wireguard peers endpoints as nessary.
+
+As the number of devices (N) in the organization increases, the the number of periodic API requests to `ListDevicesInOrganization` increases linearly.  The size of the reponse list also increases linearly.  Put those two together and you get the reponse bandwith to the API server increasing exponentially (N requests/polling interval x N device elements in the response).  This will clearly lead to scaling challenges with large organizations.
+
+This also has the problem that desired state changes recorded in the apiserver will not take effect immediatly since the `nexd` agents are likely in the middle of their poll interval, and wont poll the API server for a little while.
```
russellb, line 13:
```suggestion
This also has the problem that desired state changes recorded in the apiserver will not take effect immediately since the `nexd` agents are likely in the middle of their poll interval and won't poll the API server for a little while.
```

docs/development/design/watch-api.md
```diff
@@ -0,0 +1,191 @@
+# Nexodus Device Watch API Implementation Notes
+
+## The Problem
+
+In order to connect or disconnect devices from a Nexodus managed organization, the list of all desired devices in the organization need replicated to all devices in the organization.  The `ListDevicesInOrganization` operation provides this list and can be accessed via the following REST call:
+
+`GET /api/organizations/{organization_id}/devices`
+
+The `nexd` agent would need to periodically poll this endpoint to get the current desired state from the API server and reconcile the device configuration to add or remove Wireguard peers endpoints as nessary.
```
russellb, line 9:
```suggestion
The `nexd` agent would need to periodically poll this endpoint to get the current desired state from the API server and reconcile the device configuration to add or remove Wireguard peers endpoints as necessary.
```

docs/development/design/watch-api.md
```diff
@@ -0,0 +1,191 @@
...
+
+As the number of devices (N) in the organization increases, the the number of periodic API requests to `ListDevicesInOrganization` increases linearly.  The size of the reponse list also increases linearly.  Put those two together and you get the reponse bandwith to the API server increasing exponentially (N requests/polling interval x N device elements in the response).  This will clearly lead to scaling challenges with large organizations.
```
russellb, line 11:
```suggestion
As the number of devices (N) in the organization increases, the number of periodic API requests to `ListDevicesInOrganization` increases linearly.  The size of the response list also increases linearly.  Put those two together, and you get the response bandwidth to the API server increasing exponentially (N requests/polling interval x N device elements in the response).  This will clearly lead to scaling challenges with large organizations.
```

docs/development/design/watch-api.md
```diff
@@ -0,0 +1,191 @@
...
+
+This also has the problem that desired state changes recorded in the apiserver will not take effect immediatly since the `nexd` agents are likely in the middle of their poll interval, and wont poll the API server for a little while.
+
+## The Solution: The Watch style API
+
+Inspired by the Kubernetes List/Watch style APIs, the Nexodus `ListDevicesInOrganization` operation can be called with the `watch=true` query parameter.  Furthemore devices have been given a `revision` field that can be used to efficently make change tracking of devices possible.
```
russellb, line 17:
```suggestion
Inspired by the Kubernetes List/Watch style APIs, the Nexodus `ListDevicesInOrganization` operation can be called with the `watch=true` query parameter.  Furthermore, devices have been given a `revision` field that can be used to efficiently make change tracking of devices possible.
```

docs/development/design/watch-api.md
```diff
@@ -0,0 +1,191 @@
...
+
+If a client **watch** is disconnected then that client can start a new **watch** from the last returned `revision`; the client could also perform a fresh watch without specifying a revision and begin again.
+
+This has the effect of having better bandwith scaling characteristics since only change events are sent to N devices when the desired state changes.  If the devices are not actively joining and leaving the organization, this can be verly little bandwidth.  And even if they are joining and leaving at a constant rate, the bandwidth scales with N (not N x N).  In addition to the bandwidth savings, peer devices are pushed change events as soon as possible for processing which means they can converge on the desired state quicker than waiting for the next polling interval.
```
russellb, line 23:
```suggestion
This has the effect of having better bandwidth scaling characteristics since only change events are sent to N devices when the desired state changes.  If the devices are not actively joining and leaving the organization, this can be very little bandwidth.  And even if they are joining and leaving at a constant rate, the bandwidth scales with N (not N x N).  In addition to the bandwidth savings, peer devices are pushed change events as soon as possible for processing which means they can converge on the desired state quicker than waiting for the next polling interval.
```

docs/development/design/watch-api.md
```diff
@@ -0,0 +1,191 @@
...
+
+### Example
+
+When you use the `watch=true` query parameter the HTTP response body (served as `application/json;stream=watch`) consists of a series of JSON documents.  When you don't specify the the `revision` argument you will get all the devices in an organization like in a typical list requests followed by a `bookmark` event.
```
russellb, line 27:
```suggestion
When you use the `watch=true` query parameter, the HTTP response body (served as `application/json;stream=watch`) consists of a series of JSON documents.  When you don't specify the `revision` argument, you will get all the devices in an organization like in a typical list request followed by a `bookmark` event.
```

nerdalert (approved):
LGTM

docs/development/design/watch-api.md
```diff
@@ -0,0 +1,191 @@
...
+       "endpoint_local_address_ip4": "172.17.0.4",
+       "symmetric_nat": true,
+       "hostname": "device2",
+       "revision": 10247
+     }
+   }      
+   ...
+   ```
+
+In the above example, the client gets a bookmark event imediately indicating no changes had occured since revision 10245 but then a little while later a delete event is delivered.
```
russellb, line 92:
```suggestion
In the above example, the client gets a bookmark event immediately indicating no changes had occurred since revision 10245, but then a little while later a delete event is delivered.
```

docs/development/design/watch-api.md
```diff
@@ -0,0 +1,191 @@
...
+defer watch.Close()
+
+// get all the events..
+for {
+    kind, device, err := watch.Receive()    
+}
+
+```
+
+To simplify working against the event based *watch* operations we have implemented a Kubernetes style Informer API.
```
russellb, line 120:
```suggestion
To simplify working against the event-based *watch* operations we have implemented a Kubernetes-style Informer API.
```

docs/development/design/watch-api.md
```diff
@@ -0,0 +1,191 @@
...
+
+A Informer in essence is a local cached copy of the resources that clients are interested in. The local cache will be refreshed via the watch operation. But the use of the watch API is occurring asynchronously and hidden from the clients.
```
russellb, line 122:
```suggestion
An Informer, in essence, is a local cached copy of the resources that clients are interested in. The local cache will be refreshed via the watch operation. But the use of the watch API occurs asynchronously and is hidden from the clients.
```

docs/development/design/watch-api.md
```diff
@@ -0,0 +1,191 @@
...
+
+```go
+informer := client.DevicesApi.
+   ListDevicesInOrganization(context.Background(), orgID).
+   Informer()
+
+devices, _, err := informer.Execute()
+```
+
+Notice that result of `informer.Execute()` is the same as the traditional list style api.  Additionally an informer provides a chanel via `informer.Changed()` that you can use to wait for when you can call `informer.Execute()` to get a changed device list.
```
russellb, line 134:
```suggestion
Notice that the result of `informer.Execute()` is the same as the traditional list-style API.  Additionally, an informer provides a channel via `informer.Changed()` that you can use to wait for when you can call `informer.Execute()` to get a changed device list.
```

docs/development/design/watch-api.md
```diff
@@ -0,0 +1,191 @@
...
+            ... process the devices ....
+    }
+}
+```
+
+### Apiserver Implementation of `watch=true`
+
+The HTTP request handler servicing the `ListDevicesInOrganization` will:
+
+1. Create a **SignalBus** (described later in this do) subscription to `/devices/org={orgId}`
```
russellb, line 156:
```suggestion
1. Create a **SignalBus** (described later in this doc) subscription to `/devices/org={orgId}`
```

docs/development/design/watch-api.md
```diff
@@ -0,0 +1,191 @@
...
+2. Select matching devices from the DB.  It will force device results to be ordered by `revision` so that the last selected result has the highest revision.
+3. Each result will be sent to the client as independent JSON document.  This will continue until it gets an empty result set from the database.
```
russellb, line 158:
```suggestion
3. Each result will be sent to the client as an independent JSON document.  This will continue until it gets an empty result set from the database.
```

docs/development/design/watch-api.md
```diff
@@ -0,0 +1,191 @@
...
+4. At that point it will send the bookmark event the the clien
```
russellb, line 159:
```suggestion
4. At that point, it will send the bookmark event the the client
```

docs/development/design/watch-api.md
```diff
@@ -0,0 +1,191 @@
...
+5. Parks the go routine waiting for the http requent to be terminated or the **SignalBus** subcription to be notified.
```
russellb, line 160:
```suggestion
5. Parks the go routine waiting for the http request to be terminated or the **SignalBus** subscription to be notified.
```

docs/development/design/watch-api.md
```diff
@@ -0,0 +1,191 @@
...
+
+type Subscription interface { 
+    // You can wait on this channel to know when it's been signaled
+    Signal() <-chan bool
+    // close out the subscription.
+    Close()
+}
+```
+
+Since this interface does not hold events and even coalesces multiple Notify calls.  It results in always being non-blocking to the callers of Notify and always have a bounded amount of memory that it uses.
```
russellb, line 189:
```suggestion
Since this interface does not hold events and even coalesces multiple Notify calls, it results in always being non-blocking to the callers of Notify and always having a bounded amount of memory that it uses.
```

russellb (changes requested):
Thanks for the writeup! lgtm other than the spelling issues marked throughout.

chirino:
@russellb thanks for the review!  I really need a grammar checker!

russellb:
> @russellb thanks for the review! I really need a grammar checker!

I use Grammarly. It’s great but I’m not sure others appreciate it when I’m reviewing docs. 😆

russellb (approved):

PR #890, opened 2023-04-22
Use nexctl to validate connectivity in qa

nerdalert:
- On bulk simultaneous joins, IPAM is skipping some addresses for a tbd reason. Using `nexctl connections` probes peers so makes more sense to use and validate in the process.


chirino (approved):
lgtm

PR #894, opened 2023-04-24
bump: the grafana version to v5.0.0-rc1

chirino:


dave-tucker:
Why do we need to use the v5 RC?

chirino:
@dave-tucker mentioned we can't upgrade the version used in our clusters.. so closing for now.

chirino:
Looks like grafana is not installed in the qa/prod env anyways.

dave-tucker (changes requested):

dave-tucker:
I've rolled this into #910 so will close this one for now.

PR #906, opened 2023-04-24
Remove install script

russellb:
We have outgrown the install script for most cases. The preferred install methods are by RPM (for Fedora), with Brew (for Mac), or by downloading prebuild binaries for the rest.

We should put our energy into expanding package availability, while also ensuring that it's so easy to use that downloading a binary is enough to get started.


chirino (approved):
LGTM

PR #908, opened 2023-04-24
fix: use the nexd-wireguard-go binary if it can be found on the path.

chirino:
The nexd-wireguard-go is known to allow port reuse so that nexd can discover it's external endpoint address using STUN requests.


nerdalert (approved):
LGTM

PR #910, opened 2023-04-25
Monitoring Overhaul

dave-tucker:
This replaces #894 

Changes within this patchset are:

From @chirino:
- Bump Grafana Operator to v5.0.0-RC1
- Add new Grafana CRDs to satisfy our kube-lint job

I've then added an overhaul of the monitoring stack which works as follows:

In development we need to deploy operators for this stack to work.
`deploy/observability-operators` deploys these to the `observability` namespace - for no other reason than that's the default for one or more of the operators in question.

In production/QA, the operators are installed for us so we don't need to worry.

The `monitoring` component of the `nexodus` stack has been factored out into `deploy/nexodus-monitoring`.
This will deploy monitoring components into the `nexodus-monitoring` namespace.
This allows for monitoring only the `nexodus` namespace for local development, but we may monitor both `nexodus` and `nexodus-qa` in our ROSA deployment.

Since prometheus, grafana and jaeger will be in their own namespace, we're less likely to hit resource quota issues in qa/prod.

## ROSA

The `rosa` deployment hasn't been tested yet.
We need to open issues to make sure the required operators are installed first, plus we'll need to set up and ArgoCD job for the monitoring stack.

## Jaeger

1. In dev we should set `apiserver` to always point at the OTLP endpoint that may or may not be reachable depending on whether or not the monitoring stack has been deployed. If this doesn't cause panics, great. Otherwise we'll need to manually patch the deployment and a commented out section has been left in the dev overlay for this purpose.
1. In QA/Prod, once ArgoCD has been set up it's safe to assume that the endpoint is reachable so this can always be set.


dave-tucker:
Issue to discuss the OpenShift side of this: https://github.com/redhat-et/rosa-apps/issues/64

deploy/nexodus-monitoring/base/prometheus-service.yaml
```diff
@@ -2,13 +2,11 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: prometheus
-  namespace: observability
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
```
chirino, line 7: Are we going to be able to create cluster roles in the rosa cluster?

chirino: I think we will run into the following:
image

dave-tucker: Not sure. I've moved this into the `dev` overlay for now.
Not sure whether we'll need service accounts, RBAC and all that goodness for ROSA.

dave-tucker: Update: Looks like OpenShift only needs the `ServiceMonitor` - assuming we use OpenShift Metrics - so we'd just keep that in the `base` and move the `Prometheus` object over to the `dev` overlay.
I'll hold off doing that in this PR though as there are probably lots of other things we'll find out in discussion with the folks running the ROSA cluster - i.e they may want to install the prom operator (or may have already done so for another project).

chirino (approved):
LGTM..  I think this is likely to take a few trail and error attempts before we get it all locked in.

PR #914, opened 2023-04-25
bump: required go version 

chirino:
and switch to our fork of the golang.zx2c4.com/wireguard module


russellb:
We're going to have to drop Fedora 27 support before we can move to Go 1.20. I don't mind doing that since Fedora 38 is now GA.

russellb:
I dropped f37 support, so this should be unblocked

russellb (approved):

russellb:
@Mergifyio rebase

PR #915, opened 2023-04-25
mergify: Enforce actionlint check

russellb:
When the github actions config is changed, enforce that the `actionlint` check was successful before automatically merging the PR.


nerdalert (approved):
LGTM

PR #917, opened 2023-04-26
Remove zone from the remaining docs and flags

nerdalert:
- Also added a gitignore line for each cmd directory.


nerdalert:
In theory I would have thought this would work for the gitignore but it doesn't:
```
cmd/**/*
!cmd/**/*.go
```

I know make dist resolves it but old habits die hard. I need this one.

russellb (approved):

PR #935, opened 2023-04-29
ci: Use go 1.20 in publish-rpms

russellb:
Use the common Go env setup for the publish-rpms job. It was still
using Go 1.19, so rpm builds were broken.


nerdalert (approved):
LGTM

PR #936, opened 2023-04-29
ci: Remove more steps from publish-rpms

russellb:
Now that this job is using our custom go setup action, there's more that
can be removed since it's already taken care of.


nerdalert (approved):
LGTM

PR #939, opened 2023-04-30
Makefile: add version target

russellb:
Add a target to print the version string that is embedded in binaries at
build time.


nerdalert (approved):
LGTM

chirino (approved):
lgtm

PR #952, opened 2023-05-01
nexd: drop the requirement to install the wireguard tools.

chirino:
The wg cli tool was being used to configure the WG interface.  We now do it with the wgctl go module.


nerdalert (approved):
LGTM, I installed on a linux node that had never had wireguard installed just to be safe. Worked like a champ 🥇

chirino:
Need to test a little more on darwin... since we don't have e2e for that one yet.

chirino:
I've now verified it's not need on Darwin either.

PR #953, opened 2023-05-01
Network policy design doc

nerdalert:
See doc for details.

### DIFF
--- a/deploy/nexodus/overlays/released/kustomization.yaml
+++ b/deploy/nexodus/overlays/released/kustomization.yaml
@@ -22,10 +22,10 @@ components:
 #       namespace: nexodus
 images:
   - name: quay.io/nexodus/apiserver
-    newTag: ac5c4e70a46c13275045b5723dc01bfbada887f1
+    newTag: 3e4444a62cee24c8d5ed05652b6b91a82ce431d6
   - name: quay.io/nexodus/envsubst
-    newTag: ac5c4e70a46c13275045b5723dc01bfbada887f1
+    newTag: 3e4444a62cee24c8d5ed05652b6b91a82ce431d6
   - name: quay.io/nexodus/frontend
-    newTag: ac5c4e70a46c13275045b5723dc01bfbada887f1
+    newTag: 3e4444a62cee24c8d5ed05652b6b91a82ce431d6
   - name: quay.io/nexodus/go-ipam
-    newTag: ac5c4e70a46c13275045b5723dc01bfbada887f1
+    newTag: 3e4444a62cee24c8d5ed05652b6b91a82ce431d6
--- a/docs/development/design/security-groups.md
+++ b/docs/development/design/security-groups.md
@@ -1,0 +1,360 @@
+# Design Document: Security Groups
+
+## Introduction
+
+The goal of this design document is to outline the implementation of security groups. Security groups will contain a set of rules that specify allowed ports and protocols on the driver interface. This will be managed using nftables and initially support Linux devices only.
+
+## Phased Implementation
+
+### Phase I - Default Security Group for Organization
+
+- The default security group for an organization will be applied to all devices on startup.
+- A device can only have one policy group applied at any given time in phase I. This means each organization will have a security group. Users in the organization can CRUD the security group. AAlternatively we can scope it to only the organization owner that can modify the group.
+- When a new set of rules are applied, clear the appropriate chain and re-apply the rules. This can be more elegantly managed but introduces a great deal of complexity when rules are overlapping and may not be installed in the tables because another rule superceded it. See the section [Rule Deconfliction](#rule-deconfliction).
+
+### Phase II - User-owned Security Groups
+
+- Users can create their own security groups for more granular policies per device. The user owns the security group. It seems reasonable that other members of an org should be able to use that security group. This would be similar to how an ec2 security group gets applied to a node would be the same here with a patch.
+- A device can only have one policy group applied at any given time.
+
+### Phase III - Robust Admin Organization Policies
+
+- Overarching admin policies can be overlaid on top or in lieu of individual user policies. Further exploration needs to happen here.
+- A device can potentially have multiple policy groups applied at the same time, this could be via a seperate chain or inserts on the rule ordering.
+
+## Default Security Group and Rules
+
+Inbound rules:
+Ultimately, the default security group rules for inbound traffic by default drops all inbound traffic unless there is a match of traffic in an established state. This match is referring to traffic that is part of an existing connection initiated by the device. In Phase I all inbound traffic will be allowed until a user-friendly mechanism to install rules in the UI are complete.
+
+Outbound rules:
+The agent will add a deny rule at the end of the egress chain only when an explicit allow rule is provisioned by the user.
+
+- The default for Phase I security group is permit any traffic in both directions. There will be one nftables named `nexodus` containing two chains `nexodus-inbound` and `nexodus-outbound`. While these chains could be completely empty by default, I would propose the inbound chain have some basic permit any rules accompanied by a drop all rule. This is primarily to give some burn in time on any potential issues along with getting accustomed to defining a default policy since the explicit allow will eventually become an implicit deny by default on inbound traffic only if we follow the ec2 style model. The egress table will allows all traffic by default an implicit allow all, meaning an accept chain with no rules. If the user defines a policy blocking some protocol, destination address or destination ports those allows would be added, followed by a drop rules.
+- Ordering will be done by the order the user installs the rules. This is possible since there are no denies. As a reference, you can compare EC2 rules to Azure rules for not allowing deny statements vs allowing deny statements. Order begins to matter when denies are in place. This adds complexity which for our use case does not add any clear value.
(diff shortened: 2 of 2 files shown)


docs/development/design/network-policy.md
```diff
@@ -0,0 +1,256 @@
...
+- A device can only have one policy group applied at any given time.
+
+### Phase III - Robust Admin Organization Policies
+
+- Overarching admin policies can be overlaid on top or in lieu of individual user policies.
+- A device can potentially have multiple policy groups applied at the same time.
+
+## Default Security Group Rules
+
+The default security group rules will be an explicit allow, adopting the EC2 approach of an implicit deny. The agent will add an explicit deny at the end of the ingress or egress chain. The user only sees an implicit deny with explicit allows just as in Ec2.
```
russellb, line 26: "an explicit allow" ... of what?  all traffic?

docs/development/design/network-policy.md
```diff
@@ -0,0 +1,256 @@
...
+- The same applies to source port and destination ports, `PortFrom:8080` coupled with `PortTo:9000` would equate to a rule of `8080-9000` being permitted. `PortFrom:0 PortTo:0` will be read as `ip permit <protocol> any`. `PortFrom:443 PortTo:443` would be equivalent to `ip permit <protocol> 4434`.
+- L3 `IpRanges` are applied based on the direction field they are located in SecurityGroups. `InboundRules` have the IP prefix applied to the `saddr` field in nftables in the input chain, while `OutboundRules` are applied to the `daddr` field in the outbound chain.
+- In regard to L4 ports, destination ports are what will be supported. For example, an ingress rule of `input tcp dport 22 counter accept` would apply port 22 to the L4 dport value in the nexodus-inbound chain meaning any host can connect to the node on port 22.
+
+## Security Group User Interface
+
+- The user can add rules via web UI, the nexctl tool or the http API. There would not be support for adding or manipulating rules via the agent.
```
russellb, line 36: It's worth calling out here that with our current security model, there is no way to enforce that an agent can't make changes. In other words, we need device-specific tokens that are have more limited access controls. That way a compromise of a single device can't be used to make changes in the Nexodus API.

This is a problem much more broad than this specific proposal, though. If we don't have an issue, we should (I will check). Once we do have an issue, let's link it from here to explain more about how agent credentials could be prevented from making security policy changes.

docs/development/design/network-policy.md
```diff
@@ -0,0 +1,256 @@
...
+    "from_port": 0,
+    "to_port": 0
+  }
+```
+
+The actual rule in nftables would only insert the LPM (Longest Prefix Match). The resulting chain would look as follows.
+
+```plaintext
+table inet nexodus {
+	chain nexodus-inbound {
```
russellb, line 62: Can you fix up the markdown-lint warnings?

docs/development/design/network-policy.md
```diff
@@ -0,0 +1,256 @@
...
+	}
+
+	chain nexodus-outbound {
+		type filter hook output priority filter; policy accept;
+	}
+```
+
+The same LPM rule optimizations also applies to IPv4 and IPv6.
+
+## Default Security Group
```
russellb, line 74: Perhaps merge this with the previous section about default security group rules? I was wondering what the default was when reading that section.

docs/development/design/network-policy.md
```diff
@@ -0,0 +1,256 @@
...
+
+The default for Phase I security group would be as follows.
```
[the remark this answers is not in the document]
nerdalert: For sure, I'm almost finished nailing the default rules down in code. Will provide the actual rules.

docs/development/design/network-policy.md
```diff
@@ -0,0 +1,256 @@
...
+
+### Phase II - User-owned Security Groups
+
+- Users can create their own security groups for more granular policies per node. The user owns the security group. Seems reasonable other members of an org should be able to use that security group. Similar to how an ec2 security group gets applied to a node would be the same here with a patch.
+- A device can only have one policy group applied at any given time.
+
+### Phase III - Robust Admin Organization Policies
+
+- Overarching admin policies can be overlaid on top or in lieu of individual user policies.
+- A device can potentially have multiple policy groups applied at the same time.
```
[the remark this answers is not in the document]
nerdalert: This is a tbd. I need to dig into ec2 some and see whats feasible. Input on authorization from the team would also help with this. Ty.

docs/development/design/network-policy.md
nerdalert: Oops, I duped that. Pretty much wrote the same section twice 🤦 The more detail there the better though so not a total waste.

docs/development/design/network-policy.md
```diff
@@ -0,0 +1,256 @@
+# Design Document: Security Groups Implementation
```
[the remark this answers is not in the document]
nerdalert: Good call, I changed it to `Network Policy`.

chirino:
Should we allow defining rules that use symbolic names for devices in addition to IP ranges?  This would be super handy for rules between the devices in the organization.  We could use device hostnames or maybe some device labels that administrators can apply when they register the machines.

nerdalert:
@chirino I'm all for it. Do tags like you mentioned in EC2 cover it? Or are you thinking more along the lines of a list of addresses or wild caed DNS once that is implemented.

nerdalert:
@russellb added details and examples on CRUD operations

docs/development/design/network-policy.md
```diff
@@ -0,0 +1,360 @@
...
+
+- The default security group for an organization will be applied to all devices on startup.
+- A device can only have one policy group applied at any given time in phase I. This means each organization will have a security group. Users in the organization can CRUD the security group. AAlternatively we can scope it to only the organization owner that can modify the group.
+- When a new set of rules are applied, clear the appropriate chain and re-apply the rules. This can be more elegantly managed but introduces a great deal of complexity when rules are overlapping and may not be installed in the tables because another rule superceded it. See the section [Rule Deconfliction](#rule-deconfliction).
+
+### Phase II - User-owned Security Groups
+
+- Users can create their own security groups for more granular policies per device. The user owns the security group. It seems reasonable that other members of an org should be able to use that security group. This would be similar to how an ec2 security group gets applied to a node would be the same here with a patch.
+- A device can only have one policy group applied at any given time.
```
russellb, line 18: This doesn't sound like a natural addition to phase 1. I don't think a user should be able to drop the org-admin level policy that was added in phase 1.

The RBAC I would expect in this phase is:

- Users CRUD their own policies.
- The org owner should be able to get details of user-level policies applied to each device in an organization.
- Org owners should NOT be able to CRUD or even list all policies defined by a user.

It's questionable whether all users should be able to view the user-level policies defined for other users' devices in the same organization. I'm tempted to err on the side of don't make it visible. It could be opened up more later.

docs/development/design/network-policy.md
```diff
@@ -0,0 +1,360 @@
...
+
+### Phase III - Robust Admin Organization Policies
+
+- Overarching admin policies can be overlaid on top or in lieu of individual user policies. Further exploration needs to happen here.
+- A device can potentially have multiple policy groups applied at the same time, this could be via a seperate chain or inserts on the rule ordering.
```
russellb, line 23: I think this is pretty much what I concluded should be phase 2.

to summarize, Phases:

1. org-admin managed Policy
2, user-level policy that can be added on top of org-level policy

russellb (changes requested):

docs/development/design/network-policy.md
russellb: @chirino commented

> I'm not sure hiding device level policy from peers adds more security to the system. If two devices can peer with each other, I think you could find upsides to being able to inspect each other's policies.
> 
> For example if device A and B allow all outbound connections (the default) but deny all inbound the you could skip setting up peering between them.

That works for me. "if they can peer, viewing policy seems reasonable" sold me.

So forget my last paragraph and replace with: All org members should be able to view the user-level policies assigned to a device in the same organization.

docs/development/design/network-policy.md
```diff
@@ -0,0 +1,360 @@
+# Design Document: Network Policy
```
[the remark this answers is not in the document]
nerdalert: I would prefer leaving the api as SecurityGroups containing SecurityRules. I thought you wanted the doc renamed.

docs/development/design/network-policy.md
```diff
@@ -0,0 +1,360 @@
...
+- Protocol
+- FromPort
+- ToPort
+- Destination Address
+- Destination Port
+- OrganizationID
+
+## CRUD Actions with API for Security Groups
+
+In this section, we will outline the CRUD actions that can be performed using the API for managing security groups.
```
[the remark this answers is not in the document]
nerdalert: I have no idea how the user side will work other than s/user/org/ feel free to add. Its why its pushed to the next phase. Some of this is I dont know what I dont know.

russellb:
Sent a set of proposed changes here: https://github.com/nerdalert/apex/pull/2

russellb (approved):

PR #956, opened 2023-05-02
nexd: Avoid duplicates in the newPeers list

russellb:
The Reconcile logic had a flaw where the same peer could be appended to
the `newPeers` list twice. I haven't observed a direct failure as a
result, but I spotted it because of some suspicious debug output while
debugging a different issue.


chirino (approved):
LGTM

PR #957, opened 2023-05-02
Address proxy test reliability

russellb:
- nexd: Reject proxy rules with a blank dest host
- e2e: Update nexdStatus() check to include tunnel IPs
- e2e: Run proxy tests in parallel again

commit bd241e2827326427129a6c35330d20b73e0aa59f
Date:   Tue May 2 15:19:05 2023 -0400

    nexd: Reject proxy rules with a blank dest host
    
    Proxy rules with a blank destination host should be considered invalid.
    Add a check for this and a test case.
    
    I have observed a failure in e2e tests where proxy rules were configured
    this way as a result of an earlier error. The original error would have
    been easier to spot if the rule had been outright rejected.
    
commit a86ebc9e9ef6531a204cb99bf3e8e4e6643eea67
Date:   Tue May 2 15:27:56 2023 -0400

    e2e: Update nexdStatus() check to include tunnel IPs
    
    Update the method that checks for nexd to report its status as READY to
    also validate that it can return its tunnel IPs. I am seeing cases in
    e2e tests where attempts to get this IP sometimes come back blank
    because the request was issued too soon.
    
    This really should not be necessary. If we had a better state machine
    for nexd, we could know that nexd is running and the data plane is up.
    Right now, READY just means nexd has successfully connected with the
    control plane. Attempts to get data plane info may happen too soon. This
    is a hack to make sure nexd has had enough time to get its own local
    config.
    
    Related: https://github.com/nexodus-io/nexodus/pull/886
    
commit f0ba0c106453ff802d6e541ee817a81510fc612c
Date:   Tue May 2 16:04:29 2023 -0400

    e2e: Run proxy tests in parallel again
    
    Previous commits addressed the failures I was able to reproduce. I'm now
    able to run them in parallel reliably locally.
    
### DIFF
--- a/integration-tests/helper_test.go
+++ b/integration-tests/helper_test.go
@@ -332,18 +332,46 @@ func (helper *Helper) runCommand(cmd ...string) (string, error) {
 
 // nexdStatus checks for a Running status of the nexd process via nexctl
 func (helper *Helper) nexdStatus(ctx context.Context, ctr testcontainers.Container) error {
-	timeoutCtx, cancel := context.WithTimeout(ctx, time.Second*1000)
+	timeoutCtx, cancel := context.WithTimeout(ctx, time.Second*60)
 	defer cancel()
 	running, _ := util.CheckPeriodically(timeoutCtx, time.Second, func() (bool, error) {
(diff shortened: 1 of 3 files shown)


chirino (approved):
LGTM

PR #959, opened 2023-05-02
nexd: Avoid segfault in error handling

russellb:
I just hit a seg fault on this fmt.Errorf() line because resp was nil.
Update the code to only include the header if resp is not nil.


chirino (approved):
lgtm

Issue #960, opened 2023-05-03
TestProxyInvalidConfig periodic failure in CI


russellb:
The `TestProxyInvalidConfig` test is occasionally failing in CI. It appears to be failing at the same spot each time it does fail.

Example test run: https://github.com/nexodus-io/nexodus/actions/runs/4872240432/jobs/8690269022?pr=886

```
    helper_test.go:67: proxy_test.go:607: Running command on /TestProxyInvalidConfig-node1: --username kittehf4700c9d-e9b4-11ed-89a2-6045bd09c431 -*** proxy --ingress tcp:90000:127.0.0.1:8080
    proxy_test.go:609: 
        	Error Trace:	/home/runner/work/nexodus/nexodus/integration-tests/proxy_test.go:609
        	Error:      	An error is expected but got nil.
        	Test:       	TestProxyInvalidConfig
```

https://github.com/nexodus-io/nexodus/actions/runs/4872240432/jobs/8690269022

```
    helper_test.go:67: proxy_test.go:607: Running command on /TestProxyInvalidConfig-node1: --username kittehf4700c9d-e9b4-11ed-89a2-6045bd09c431 -*** proxy --ingress tcp:90000:127.0.0.1:8080
    proxy_test.go:609: 
        	Error Trace:	/home/runner/work/nexodus/nexodus/integration-tests/proxy_test.go:609
        	Error:      	An error is expected but got nil.
        	Test:       	TestProxyInvalidConfig
```

https://github.com/nexodus-io/nexodus/actions/runs/4855535480/jobs/8654230289

```
    helper_test.go:67: proxy_test.go:607: Running command on /TestProxyInvalidConfig-node1: --username kitteh7579bda6-e871-11ed-9e63-000d3a33f669 -*** proxy --ingress tcp:90000:127.0.0.1:8080
    proxy_test.go:609: 
        	Error Trace:	/home/runner/work/nexodus/nexodus/integration-tests/proxy_test.go:609
        	Error:      	An error is expected but got nil.
        	Test:       	TestProxyInvalidConfig

```

russellb:
Reproduced a failure locally, though it was a different test case than the 3 quoted above:

```
[...]

PR #961, opened 2023-05-03
e2e: Add logging to TestProxyInvalidConfig

russellb:
I'm seeing failures in this test in CI that I have not been able to
reproduce locally. In the failure case, a test case will return without
an error. Add logging of any output returned in case that gives any
hints.

I also refactored the test to DRY it up.

Related to #960 


chirino (approved):
lgtm

Issue #789, opened 2023-04-10
Cached tokens should be service specific


russellb:
Now that `nexd` caches refresh tokens, it's possible for `nexd` to be in a broken state if you try to switch which service you're using from the same host. For example, I ran against `try.nexodus.io` and then tried to switch to `qa.nexodus.io`.

```
$ sudo nexd https://qa.nexodus.io
{"level":"info","ts":1681152386.07538,"caller":"nexodus/keys_darwin.go:24","msg":"Existing key pair found at [ /usr/local/etc/wireguard/public.key ] and [ /usr/local/etc/wireguard/private.key ]"}
{"level":"fatal","ts":1681152386.352436,"caller":"nexd/main.go:106","msg":"get user error: Get \"https://api.qa.nexodus.io/api/users/me\": oauth2: cannot fetch token: 400 Bad Request\nResponse: {\"error\":\"invalid_grant\",\"error_description\":\"Invalid refresh token\"}"}
```

1. Our cached tokens should probably be tied to the service they came from.
2. If the cached token is invalid for whatever reason, `nexd` should fallback gracefully to starting auth over without it.

russellb:
This was resolved by #1505

PR #962, opened 2023-05-03
Handle token issues when swapping api-servers

nerdalert:
- Response header needed to be nil checked to avoid a segfault in the case where the token is invalid and no repsonse header is received.
- Added checks for both invalid_grant and invalid_token to give a user message on how to resolve the token error.
- The checks both error messages because when you swap api-servers using a token for another server, the first error message you get is invalid_token and all subsequent messages are invalid_grant in that order.


russellb:
looks related to #789

russellb (approved):
lgtm for fixing the seg fault.

I think we have more work to do to improve the scenario of switching between servers

#789

PR #963, opened 2023-05-03
docs: Fix doc title copied from template

russellb:
Fix the document title for the design doc about the project name. It
was still the title copied from the template.

Also update the template to include the word "TEMPLATE" so it's a
little easier to find in the list of docs on the docs site.


chirino (approved):
LGTM

PR #967, opened 2023-05-04
Fix minor issues in the fsm design doc

vishnoianil:


docs/development/design/fsm.md
```diff
@@ -48,11 +47,12 @@ nexd may be in one of these states:
 
 1. UNAUTHENTICATED: No configuration has been provided
 1. AUTHENTICATING: Performing authentication, either interactively or via token
-1. AUTHENTICATED: User is succesfully authenticated
+1. AUTHENTICATED: User is successfully authenticated
 1. UNREGISTERED: Authenticated but device has not been registered with the control plane
 1. REGISTERED: Device is registered with the control plane, but data plane is not established
 1. PEERING: Device is connected to control plane and fetching peer listings
 1. UP: Data plane is established
+1. DOWN: Data plane is not established
```
russellb, line 55:
```suggestion
1. DOWN: Data plane has been administratively turned off while retaining a connection to the control plane.
```

docs/development/design/fsm.md
```diff
@@ -28,7 +28,6 @@ stateDiagram-v2
     REGISTERED --> UNAUTHENTICATED: "logout"
     UNREGISTERED --> REGISTERING: "register_device"
     UNREGISTERED --> UNAUTHENTICATED: "logout"
-    REGISTERED --> UNAUTHENTICATED: "logout"
```
russellb, line 31: Why is this line removed?

russellb (changes requested):

docs/development/design/fsm.md
vishnoianil: @russellb looking at the current fsm graph, transition to DOWN Is not always administratively down, "peering_error" can transition it to DOWN as well.

docs/development/design/fsm.md
vishnoianil: duplicate: see line 28

vishnoianil: generated graph, shows duplicate lines from REGISTERED-> UNAUTHENTICATED (https://github.com/dave-tucker/apex/blob/b39acf40d7c51cdfdb50db6759aa1cf09151b36c/docs/development/design/fsm.md)

docs/development/design/fsm.md
russellb: Ah, you're right.

```suggestion
1. DOWN: Data plane is not established, either due to an error or it has been placed down administratively using nexctl
```

russellb (approved):

PR #964, opened 2023-05-03
e2e: Fix major flaw in TestProxyInvalidConfig

russellb:
It turns out this test was never actually testing what it was supposed
to. The intent was to start `nexd` with invalid proxy rules and verify
that it exited with an error. The problem is that a very critical
argument was left out ... `nexd`. What was happening in every test case
was that it tried to run a command called `--username`. Of course, this
fails. Every so often, containerExec() would not return an error, but
since I added debug output, I was able to spot the error message that
clued me in to what went wrong.

```
    proxy_test.go:619: nexd output: �OCI runtime exec failed: exec
failed: unable to start container process: exec: "--username":
executable file not found in $PATH: unknown
 ```

Big oops.

This change fixes it so the test actually runs `nexd`. It also stops
printing the debug output every time and only dumps it if the command
unexpectedly returns success.

Closes #960

Signed-off-by: Russell Bryant <rbryant@redhat.com>

PR #968, opened 2023-05-04
fix: test container log output had 8 bytes of binary junk before the …

chirino:
…output.

We now strip it out to simplify looking at test results.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,6 +4,7 @@ go 1.20
 
 require (
 	github.com/Nerzal/gocloak/v13 v13.3.0
+	github.com/ahmetb/dlog v0.0.0-20170105205344-4fb5f8204f26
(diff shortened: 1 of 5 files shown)


russellb:
Did you find any reference that explains what the 8 bytes of output is?

russellb:
also, check out this terrible thing I did because I didn't understand what the junk was in the output: https://github.com/nexodus-io/nexodus/blob/main/integration-tests/util_test.go#L303

I bet you can remove that and it's fine now after your change

russellb:
e2e test failure in the last run is fixed by: https://github.com/nexodus-io/nexodus/pull/964

russellb (approved):
note to self on what the 8 bytes are: 
https://ahmet.im/blog/docker-logs-api-binary-format-explained/

Issue #787, opened 2023-04-10
Use a broader list of public STUN servers


russellb:
Right now we hit the same two google stun servers. It would be better to spread our load across a broader set of public stun servers. Some reasons:

- Avoid getting blocked if our rate of stun queries gets too high as the project grows
- Provider higher availability by having a larger set of servers to try in case the google servers went down at some point.

This repo maintains a regularly updated list of publicly available stun servers: https://github.com/pradt2/always-online-stun

We could pull this list periodically and fall back to a hard coded set of servers in the code if we can't get this list.

russellb:
> @russellb I am wondering if for long term, we should host our own stun servers across different (aws) regions (although it has cost). Relying on public stuns is a risky dependency in my opinion, and not sure if customer prefer to use public stun servers for their edge devices.

Using known public ones seems like a good default starting point and also seems good enough for our demo service.

Maybe Nexodus should allow stun servers to be customized at different levels (defaults for the whole service, an organization, or a device). That would provide a lot of flexibility.

PR #971, opened 2023-05-04
fix: #787 Use a broader list of public STUN servers

chirino:


internal/nexodus/stun.go
```diff
@@ -0,0 +1,36 @@
...
+	stunServers = []string{}
+)
+
+func init() {
+	stunServers = strings.Split(stunServersTxtFile, "\n")
+	for i := range stunServers {
+		stunServers[i] = strings.TrimSpace(stunServers[i])
+	}
+	// #nosec G404
+	currentStunServer = rand.Intn(len(stunServers))
```
russellb, line 25: A minor improvement would be to use `rand.Shuffle()` so that every `nexd` instance isn't going through the list in the same order.

https://pkg.go.dev/math/rand#Shuffle

nerdalert (approved):
LGTM

PR #972, opened 2023-05-05
Add license checking for Go dependencies

russellb:
Add a Makefile target and a CI job that checks the licenses of Go
dependencies. If a license is enountered that is either unknown or is
of a more restrictive type than we'd like to allow, this will result
in an error. An example would be if someone accidentally introduced an
AGPL licensed dependency, this would catch it.

More details on the checker: https://github.com/google/go-licenses

To see what is categorized in the types we disallow, "forbidden" or
"restrictive", see:
https://github.com/google/licenseclassifier/blob/main/license_type.go

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -112,6 +112,17 @@ jobs:
       - name: OPA Lint and Tests
         run: make opa-lint
 
+  go-licenses:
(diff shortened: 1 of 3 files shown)


nerdalert (approved):
LGTM

PR #980, opened 2023-05-06
docs: Drop mention of zones

russellb:
I noticed this document still mentioned zones. It was only in the
section that instructed the reader to set up a relay. Since a relay is
not required in most casees, I just dropped the section instead of just
adjusting the zone references.


nerdalert (approved):
LGTM

PR #981, opened 2023-05-06
Add network-router functionality

nerdalert:
- Adds network router functionality in conjunction with child prefix.
- When a child-prefix is specified, that route gets propagated  across the peer group. When --network-router is passed with child prefix, a series of nftable chains are added that enable traffic to be masqueraded via a physical device on the node.
- The physical device is discovered via existing helpers in the code currently used to choose the address with a default gateway attached to the dev. If the route is found via netlink, the interface tied to the route will be used. If a route tied to an interface is not in the device's main routing table, the interface with the default route is used. There could be scenarios where the network is not directly attached but would be picked up via the default route.
- e2e tests add two networks simulating two disparate networks that have no direct connectivity to each other and can only talk to the edge network-router that will advertise their prefix and perform snat.
- v4 support only until v6 is added to child-prefix. Will add v6 once #983 merges on another PR.


nerdalert:
This is functioning but I would rather wait until we get nftables in to use some of the methods there for provisioning the tables/chains/rules. I also need to think on how to test it in e2e and write docs.

This has some nice scenarios, one in particular would be the point-to-site scenario, where the site could be a air-gapped infra or infra that you only want access to always be funneled through a node that is the only node permitted into the site. 

Example Topology:

```
                                  +--------------+
                                  |  Nex-device  |
                                  | (WireGuard)  |
                                  +--------------+
                                         | wg0
                                         | Tunnel
                                         | wg0
          +--------------+        +--------------+        +---------------+
          | Non-Wireguard|--------| Nex-Exit-Node|--------| Non-Wireguard |
          |   Device-1   |  eth0  | (WireGuard)  |  eth0  |   Device-3    |
          +--------------+        +--------------+        +---------------+
                                         |
                                         | eth0
                                 +---------------+
                                 | Non-Wireguard |
                                 |   Device-2    |
                                 +---------------+

```

nerdalert:
Im getting backed up by Security Groups. Going to open this for merge and will just refactor to the nftable helpers in the Security Group PR. This PR is blocking exit nodes as well, which is ready for push as soon the the origin node bits merge with this.

russellb:
looks like this one depends on #1027

russellb:
@nerdalert #1027 covers `--exit-node`, but is there a design doc that covers `--network-router`?

I'm starting to think that we need one doc that covers all of these:

- `--child-prefix`
- `--network-router`
- `--exit-node`

These things are tightly related, and I'd like to be able to think about them all together.

internal/nexodus/net_router.go
```diff
@@ -0,0 +1,50 @@
...
+		return fmt.Errorf("unable to determine default interface for the network router node, please specify using --local-endpoint-ip: %w", err)
+	}
+
+	nx.netRouterInterfaceMap = make(map[string]*net.Interface)
+
+	// iterate over childPrefixes and find the best matching interface for each prefix based on the device's
+	// default namespace routing table. If no match is found, use the interface containing the default gateway.
+	for _, prefix := range nx.childPrefix {
+		if util.IsIPv6Prefix(prefix) {
+			nx.logger.Debugf("IPv6 is not currently supported for --net-router: %s", prefix)
```
russellb, line 25: might be work a Warning instead of Debug here

russellb (approved):
nice work!!

Issue #982, opened 2023-05-07
Brief bug summary


nerdalert:
### Describe the bug

If you join a node 3 times with a child prefix you will hit a bug. TLDR we are releasing it if the prefix is already leased. I need to think through a clean way to deal with existing prefix leases and what happens when either the same node or a new node requests the same prefix. In the meantime we need to patch UpdateDevice to check if the existing device.ChildPrefix in the api-server device entry matches the child prefix in the request.ChildPrefix update.

### How to Reproduce

Join using:

```
NEXD_LOGLEVEL=debug  ./nexd-dev-linux-amd64 --username 'admin' --password 'floofykittens' router --child-prefix 10.180.0.0/20  https://try.nexodus.127.0.0.1.nip.io
```

- 1st Join

```
ipam=> SELECT * FROM "prefixes_a0227d60_cef4_41da_8dac_792de6d611a4";
      cidr      |                                                                                                              prefix
----------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 100.100.0.0/16 | {"IPs": {"100.100.0.0": true, "100.100.0.1": true, "100.100.255.255": true}, "Cidr": "100.100.0.0/16", "Version": 1, "IsParent": false, "Namespace": "", "ParentCidr": "", "ChildPrefixLength": 0, "AvailableChildPrefixes": {}}
 200::/64       | {"IPs": {"200::": true, "200::1": true}, "Cidr": "200::/64", "Version": 1, "IsParent": false, "Namespace": "", "ParentCidr": "", "ChildPrefixLength": 0, "AvailableChildPrefixes": {}}
 10.180.0.0/20  | {"IPs": {"10.180.0.0": true, "10.180.15.255": true}, "Cidr": "10.180.0.0/20", "Version": 0, "IsParent": false, "Namespace": "", "ParentCidr": "", "ChildPrefixLength": 0, "AvailableChildPrefixes": {}}
(3 rows)
```

- 2nd Join

```
time="2023-05-07T03:20:28Z" level=info msg="DEBUG_BREAK0 --> [10.180.0.0/20]"
time="2023-05-07T03:20:28Z" level=info msg="DEBUG_BREAK1 prefix 10.180.0.0/20"
time="2023-05-07T03:20:28Z" level=info msg="DEBUG_BREAK2 {}"
time="2023-05-07T03:20:28Z" level=info msg="DEBUG_BREAK3 prefix 10.180.0.0/20"
[...]

PR #983, opened 2023-05-07
Check if the ChildPrefix is the same as the existing in UpdateDevice

nerdalert:
- Adds a function and tests to compare the existing vs new prefix request.
- Adds e2e that tests for this condition.
- Tracked in #982
- IPAM errors out of a default route is added as a child prefix
  since that creates an overlap. Its not important to store an
  exit-node prefix in IPAM, so this skips adding a 0.0.0.0 or ::/0
  as a child-prefix.
- Adds ipv6 child prefix support.

### DIFF
--- a/integration-tests/integration_test.go
+++ b/integration-tests/integration_test.go
@@ -393,7 +393,7 @@ func TestHubOrganization(t *testing.T) {
 	require.NotContainsf(node2dump, node3IP, "found deleted device node still in wg show wg0 dump tables of a device")
 }
 
-// TestChildPrefix tests requesting a specific address in a newly created organization. This will start nexd three
+// TestChildPrefix tests requesting a specific address in a newly created organization for v4 and v6. This will start nexd three
 // different times. The first makes sure the prefix is created and routes are added. The second is started and then killed.
 // The third start of nexd is to validate the child-prefix was not deleted from the ipam database. TODO: test changing the child-prefix
 func TestChildPrefix(t *testing.T) {
@@ -406,10 +406,12 @@ func TestChildPrefix(t *testing.T) {
 	password := "floofykittens"
 	username, cleanup := helper.createNewUser(ctx, password)
 	defer cleanup()
-	node3LoopbackNet := "172.16.10.101/32"
 	node2LoopbackNet := "172.16.20.102/32"
-	node3ChildPrefix := "172.16.10.0/24"
-	node2ChildPrefix := "172.16.20.0/24"
+	node3LoopbackNet := "172.16.10.101/32"
+	node2LoopbackNetV6 := "200:2::1/64"
+	node3LoopbackNetV6 := "200:3::1/64"
+	node2ChildPrefix := "172.16.20.0/24,200:2::/64"
+	node3ChildPrefix := "172.16.10.0/24,200:3::/64"
 
 	// create the nodes
 	node1, stop := helper.CreateNode(ctx, "node1", []string{defaultNetwork}, enableV6)
@@ -439,17 +441,31 @@ func TestChildPrefix(t *testing.T) {
 		"router", fmt.Sprintf("--child-prefix=%s", node3ChildPrefix),
 	)
 
-	// add loopbacks to the containers that are contained in the node's child prefix
+	// add v4 loopbacks to the containers that are contained in the node's child prefix
 	_, err = helper.containerExec(ctx, node3, []string{"ip", "addr", "add", node3LoopbackNet, "dev", "lo"})
 	require.NoError(err)
 	_, err = helper.containerExec(ctx, node2, []string{"ip", "addr", "add", node2LoopbackNet, "dev", "lo"})
+	require.NoError(err)
+	// add v6 loopbacks to the containers that are contained in the node's child prefix
+	_, err = helper.containerExec(ctx, node3, []string{"ip", "-6", "addr", "add", node3LoopbackNetV6, "dev", "lo"})
+	require.NoError(err)
+	_, err = helper.containerExec(ctx, node2, []string{"ip", "-6", "addr", "add", node2LoopbackNetV6, "dev", "lo"})
 	require.NoError(err)
 
 	// parse the loopback ip from the loopback prefix
 	node3LoopbackIP, _, _ := net.ParseCIDR(node3LoopbackNet)
 	node2LoopbackIP, _, _ := net.ParseCIDR(node2LoopbackNet)
+	// parse the loopback ipv6 from the loopback prefix
+	node3LoopbackIPv6, _, _ := net.ParseCIDR(node3LoopbackNetV6)
+	node2LoopbackIPv6, _, _ := net.ParseCIDR(node2LoopbackNetV6)
 
-	// address will be the same, this is just a readiness check for gather data
+	// readiness check
+	err = helper.nexdStatus(ctx, node2)
+	require.NoError(err)
+	err = helper.nexdStatus(ctx, node3)
+	require.NoError(err)
+
+	// gather the wg0 v4 addresses
 	node3IP, err := getContainerIfaceIP(ctx, inetV4, "wg0", node3)
 	require.NoError(err)
 	node2IP, err := getContainerIfaceIP(ctx, inetV4, "wg0", node2)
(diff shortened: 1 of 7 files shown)


integration-tests/integration_test.go
```diff
@@ -468,6 +480,85 @@ func TestChildPrefix(t *testing.T) {
...
+		helper.runNexd(ctx, node3,
+			"--username", username, "--password", password,
+			"router", fmt.Sprintf("--child-prefix=%s", node3ChildPrefix),
+		)
+
+		// address will be the same, this is just a readiness check for gather data
+		node3IP, err = getContainerIfaceIP(ctx, inetV4, "wg0", node3)
+		require.NoError(err)
+		node2IP, err = getContainerIfaceIP(ctx, inetV4, "wg0", node2)
+		require.NoError(err)
```
russellb, line 526: If you don't specifically care about getting it from `wg0`, there is also a `getTunnelIP()` helper that will get you the same IP, but via `nexctl nexd get tunnelip`. Using that interface should work for all configurations (normal mode, proxy mode, future drivers that aren't wg0 ...).

A major difference is that `getTunnelIP()` does not do retries, while right now `getContainerIfaceIP()` does. We've got retries in multiple places all dealing with the same issue (`nexd` might not be up yet). I think it would nice to simplify by having one place that does that, probably `nexdStatus()`.

internal/handlers/device.go
```diff
@@ -171,20 +171,27 @@ func (api *API) UpdateDevice(c *gin.Context) {
...
-				if _, ok := prefixAllocated[prefix]; ok {
-					if err := api.ipam.ReleasePrefix(ctx, device.OrganizationID, prefix); err != nil {
-						return err
-					}
-				} else {
-					// otherwise we need to allocate it
-					if err := api.ipam.AssignPrefix(ctx, device.OrganizationID, prefix); err != nil {
-						return err
+			// check if the updated device child prefix matches the existing device prefix
+			if !childPrefixEquals(device.ChildPrefix, request.ChildPrefix) {
```
russellb, line 175: Please combine this with the `if` statement that includes this block.

```suggestion
			if request.ChildPrefix != nil && !childPrefixEquals(device.ChildPrefix, request.ChildPrefix) {
```

internal/handlers/device.go
```diff
@@ -171,20 +171,27 @@ func (api *API) UpdateDevice(c *gin.Context) {
...
+				prefixAllocated := make(map[string]struct{})
+				for _, prefix := range device.ChildPrefix {
+					prefixAllocated[prefix] = struct{}{}
+				}
+				for _, prefix := range request.ChildPrefix {
+					isDefaultRoute := util.IsDefaultIPv4Route(prefix) || util.IsDefaultIPv6Route(prefix)
+					// If the prefix is not a default route, process IPAM allocation/release
+					if !isDefaultRoute {
```
russellb, line 183: This is a style preference. I prefer inverting the logic to reduce levels of nesting.

```suggestion
					if isDefaultRoute {
					    continue
					}
```

internal/util/ip_utils.go
```diff
@@ -56,3 +56,31 @@ func AppendPrefixMask(ipStr string, maskSize int) (string, error) {
...
+	if err != nil {
+		ipAddr = net.ParseIP(ip)
+		if ipAddr == nil {
+			return false
+		}
+	}
+
+	ipv6Default := net.ParseIP("::")
+	return ipAddr.Equal(ipv6Default) || (ipNet != nil && ipNet.IP.Equal(ipv6Default) && ipNet.Mask.String() == "<nil>")
+}
```
russellb, line 86: A nice addition would be a wrapper that checks both for you:

```go
func IsDefaultIPRoute(ip string) bool {
    return IsDefaultIPv4Route(ip) || IsDefaultIPv6Route(ip)
}
```

internal/handlers/device.go
```diff
@@ -298,8 +305,11 @@ func (api *API) CreateDevice(c *gin.Context) {
 		}
 		// allocate a child prefix if requested
 		for _, prefix := range request.ChildPrefix {
-			if err := api.ipam.AssignPrefix(ctx, org.ID, prefix); err != nil {
-				return fmt.Errorf("failed to assign child prefix: %w", err)
+			// Skip the prefix assignment if it's an IPv4 or IPv6 default route
+			if !util.IsDefaultIPv4Route(prefix) && !util.IsDefaultIPv6Route(prefix) {
```
russellb, line 309: See later where I suggest a helper wrapper that turns this into a single function call.

internal/handlers/device.go
```diff
@@ -448,3 +458,20 @@ func (api *API) DeleteDevice(c *gin.Context) {
...
+	for _, value := range existingPrefix {
+		countMap[value]++
+	}
+	for _, value := range newPrefix {
+		countMap[value]--
+		if countMap[value] < 0 {
+			return false
+		}
+	}
+	return true
```
russellb, line 476: This looks correct to me, but I really wish there was a first-class implementation of Sets for go. What we really want here is to view each set of prefixes as a set and they are equal if the difference between the sets is an empty set.

We could use this library if we wanted to turn the childPrefix storage into a set natively. It would handle things like detecting if the same thing is specified twice as input, it only ends up in the set once. It would also simplify all of the operations where you're having to create maps on the fly to help with comparisons.

https://github.com/deckarep/golang-set

internal/handlers/user.go
```diff
@@ -311,7 +311,7 @@ func (api *API) DeleteUserFromOrganization(c *gin.Context) {
 			Where("user_id = ?", userID).
 			Where("organization_id = ?", orgID).
 			Delete(&UserOrganization{}); res.Error != nil {
-			c.JSON(http.StatusInternalServerError, models.NewApiInternalError(fmt.Errorf("WTF failed to remove the association from the user_organizations table: %w", res.Error)))
+			c.JSON(http.StatusInternalServerError, models.NewApiInternalError(fmt.Errorf("failed to remove the association from the user_organizations table: %w", res.Error)))
```
russellb, line 314: lol.

internal/handlers/device.go
```diff
@@ -171,20 +171,27 @@ func (api *API) UpdateDevice(c *gin.Context) {
...
-				} else {
-					// otherwise we need to allocate it
-					if err := api.ipam.AssignPrefix(ctx, device.OrganizationID, prefix); err != nil {
-						return err
+			// check if the updated device child prefix matches the existing device prefix
+			if !childPrefixEquals(device.ChildPrefix, request.ChildPrefix) {
+				prefixAllocated := make(map[string]struct{})
+				for _, prefix := range device.ChildPrefix {
+					prefixAllocated[prefix] = struct{}{}
+				}
```
russellb, line 179: A comment further down by `childPrefixEquals()` I talk about storing prefixes in a Set data structure instead of a slice of strings. If we did that, you wouldn't have to do this loop at all..

internal/handlers/device.go
```diff
@@ -298,8 +305,11 @@ func (api *API) CreateDevice(c *gin.Context) {
 		}
 		// allocate a child prefix if requested
 		for _, prefix := range request.ChildPrefix {
```
russellb, line 307: As far as I can tell, nothing is validating that the child prefixes are valid. They could be garbage and we're relying on `api.ipam.AssignPrefix()` to return an error. I don't think the code should make it that far if the input is obviously malformed.

internal/util/ip_utils_test.go
```diff
@@ -121,3 +123,43 @@ func TestAppendPrefixMask(t *testing.T) {
...
+// TestIsIPv4DefaultRoute tests the IsIPv4DefaultRoute function.
+func TestIsDefaultIPv4Route(t *testing.T) {
+	testCases := []struct {
+		input    string
+		expected bool
+	}{
+		{"0.0.0.0", true},
+		{"0.0.0.0/0", true},
+		{"192.168.1.1", false},
+		{"192.168.1.1/24", false},
```
russellb, line 136: I'd add a test case with garbage input. Right now this code is called on invalidated user input. I'd like to change that so the input is validated before it gets this far, but might as well make sure the code doesn't blow up (now or after future changes) on garbage.

russellb (changes requested):
There are a couple of suggestions in here that go beyond the scope of this PR, so I don't feel them as required. That includes:

1. Cleanup to how we check `nexd` readiness in the integration steps. It's not very consistent right now, but that's not new in this PR.
2. I thought out loud about changing the data structure used to story the child prefixes because multiple comparisons needed in this PR are Set operations. Storing them as a set would really clean things up, IMO. Child prefixes aren't introduced here, though, so it's beyond the scope of this PR.

integration-tests/integration_test.go
```diff
@@ -437,15 +441,23 @@ func TestChildPrefix(t *testing.T) {
 		"router", fmt.Sprintf("--child-prefix=%s", node3ChildPrefix),
 	)
```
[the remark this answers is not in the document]
nerdalert: There was an IP wg0 address check further down. Swapped it with a status call.

internal/handlers/user.go
nerdalert: lol

russellb (approved):

PR #984, opened 2023-05-07
apiserver: implement ext_autz grpc endpoint

chirino:
We were stuffing the AccessToken in a browser cookie. With this change we get the AccessToken from the redis session.  This lowers the amount of data placed into browser cookies to just being the Redis session id.

This will also give us a place to implement per device API token in the future.

### DIFF
--- a/cmd/apiserver/main.go
+++ b/cmd/apiserver/main.go
@@ -2,17 +2,21 @@ package main
 
 import (
 	"context"
+	auth "github.com/envoyproxy/go-control-plane/envoy/service/auth/v3"
+	redisStore "github.com/go-session/redis/v3"
+	"github.com/go-session/session/v3"
+	"github.com/nexodus-io/nexodus/internal/signalbus"
+	"github.com/nexodus-io/nexodus/internal/util"
 	"github.com/redis/go-redis/v9"
+	"google.golang.org/grpc"
 	"log"
+	"net"
 	"net/http"
 	"os"
 	"os/signal"
 	"sync"
 	"syscall"
 	"time"
-
-	"github.com/nexodus-io/nexodus/internal/signalbus"
-	"github.com/nexodus-io/nexodus/internal/util"
 
 	agent "github.com/nexodus-io/nexodus/pkg/oidcagent"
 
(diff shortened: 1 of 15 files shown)


russellb:
@dave-tucker fyi this one has a review request for you

cmd/apiserver/main.go
```diff
@@ -234,7 +247,24 @@ func main() {
 
 				store := inmem.New()
 
-				api, err := handlers.NewAPI(ctx, logger.Sugar(), db, ipam, fflags, store, signalBus)
+				sessionStore := cookie.NewCookieStore(
```
dave-tucker, line 250: Style nit - would prefer to set it like this.
But do we still need to support both?
```go
var sessionStore session.Store
if redisServer != "" {
  // setup redis store
} else {
  // setup cookie store
}
```

cmd/apiserver/main.go
```diff
@@ -290,23 +320,38 @@ func main() {
...
 					ReadTimeout:       5 * time.Second,
 					ReadHeaderTimeout: 5 * time.Second,
 					WriteTimeout:      10 * time.Second,
 				}
+				defer util.IgnoreError(httpServer.Close)
 
 				go func() {
-					if err = server.ListenAndServe(); err != nil {
+					if err = httpServer.ListenAndServe(); err != nil {
+						log.Fatal(err)
```
dave-tucker, line 340: I know we're already `log.Fatal` ing, but could we not do that please? especially from a goroutine.
it's going to immediately os.Exit(1) and deferred functions aren't going to get called.
we should probably propagate the error back to the main func on a channel then exit from there.

dave-tucker (changes requested):
Overall looks good, but the goroutine handling in apiserver@main could use some attention

cmd/apiserver/main.go
chirino: You're likely right.. we only test the redis case anyways.  I'll comment it out just in case we want to bring it back in the future.

cmd/apiserver/main.go
chirino: I can fix that.

cmd/apiserver/main.go
```diff
@@ -315,7 +360,34 @@ func main() {
 					<-ctx.Done()
 				})
 				wg.Wait() // context is canceled via signal.
-				server.Close()
+
+				// Try to do a graceful shutdown for 5 seconds...
+				doneEvents := make(chan struct{})
```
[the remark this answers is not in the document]
chirino: So I think we still need a channel, but maybe we don't need a counter.  I'll just add a goroutine that waits on the wg and signals the channel.

russellb:
@chirino if you rebase this I'll take a look at it

akoserwal:
Lgtm for managing the access token in redis cache

akoserwal (approved):

PR #997, opened 2023-05-08
Refactored WG Peers

nerdalert:
- Mostly much needed cleanup and simplification.
- Renamed some files to match current style.


internal/nexodus/wg_peers.go
```diff
@@ -0,0 +1,200 @@
+package nexodus
```
russellb, line 1: git seems to think you removed one file and added a new one, so it's not showing the changes as a diff. Do you remember how you did it?  Was it a `git mv`?

nerdalert: I used git mv and still see it as a delete and add. Reverted the filename to make the review not insane.

russellb (approved):

PR #998, opened 2023-05-09
Doc design proposal to use a shared CG-NAT address space by default.

chirino:


vishnoianil:
Just throwing alternate option. If nexodus doesn't allow user to allocate the overlapping subnet, that can also address this issue? That way we can keep the same model, and share the devices across multiple org as well.  do you issues with that approach (Except we need to do some work with IPAM)?

vishnoianil:
> If nexodus doesn't allow user to allocate the overlapping subnet
If an org is assigned with 10.100.1.0/24, the same subnet is not going to be assigned to the another org. If another org is trying to assign 10.100.0.0/16, it won't be allowed as well, because 10.100.1.0/24 is assigned to the org. We can provide list of available subnets to the users as well to pick from.

We are allocating one big subnet for sharing device, but also allowing users to create their own private subnet (bullet 5), that will restrict them from sharing the device. I am wondering rather than enforcing the big subnet on all the user (and assuming that they want to share the devices), can't we leave it to user, to allocate a big subnet if they really want to share the devices across it's organization. That will give them freedom of how they want to manage their device (shared, private etc)?

docs/development/design/shared-cidr.md
```diff
@@ -0,0 +1,13 @@
+# Why we should be using a single CG-NAT IPv4 Address space in Nexodus
```
russellb, line 1: Would you mind tweaking this to follow the template?

https://docs.nexodus.io/development/design/TEMPLATE/

Some of the "alternatives considered" are discussed in PR comments.

russellb (changes requested):
In general, I'm convinced that this seems valuable. I think everyone should have a chance to weigh in.

I also have a nitpick request to update to follow the design doc template.

vishnoianil:
>I feel like the "Simple UX Above Features" guiding principle applies here.
https://docs.nexodus.io/development/development/

In general I agree with that principal, but subnetting is one of core feature of Nexodus, so It's not just only a matter of simplicity here. Nexodus is at L3 layer, where Network Admin's role is pretty critical. Specifically for the on-prem deployment  they would prefer to have some control over subnetting as per their requirements, so I am not very convince having a big global range as a default is a good choice for that scenario, but I might be wrong as well, so would love to hear others opinion on it.

Secondary concern is big ranges generally means a big blast radius, for 

Scale: Can we handle mesh network of 4M devices (that's possible worst case scenario, which practically we might not hit, but it's a attach surface)

Security: one compromised device can impact larger number of connected devices, specifically devs are not really good with keeping tab on where the device is shared etc (we all have shutdown the firewall while developing stuff).

Performance: Although It's a big ip address space, but the sharing of device is not by default, it's something that user need to initiate, so every IP address in this space need to be dealt with individually (storage, iptable rules etc), so possibly 4 millions entries in iptables, rather than subnet based entries ?

Migration: In case we need to change our subnetting strategy - it's easier to open the subnets, rather than restricting the wider subnet.

I think whatever we decide here, leaving a flexibility in deciding the subnet range probably would be helpful in long run (specifically for the on-prem scenarios)

chirino:
@vishnoianil so I'm just proposing that the default orgs use the simple shared IPAM namespace.  Custom orgs can be configured to have a non-shared IPAM namespace.  So those advanced network admins, can get that if they really want it.

chirino:
@russellb updated to use the template layout.

docs/development/design/shared-cidr.md
```diff
@@ -0,0 +1,32 @@
...
+
+Currently, each org gets a new IPAM address space for IP address allocation. This means the same IP address could be allocated to devices in different organizations. This would prevent a device in from being shared with other organizations.
+
+Instead, organizations should default to using the shared CG-NAT address space which is 100.64.0.0/10, i.e. IP addresses from 100.64.0.0 to 100.127.255.255. This is about 4 Million addresses.
+
+## Proposal
+
+Why we should be using a single CG-NAT IPv4 Address space in Nexodus:
+
+1. Giving devices a stable IP identity is paramount.  This identity should be preserved even if the device is moved from one organization to another organization.
```
russellb, line 15: How would this work when some organizations use the shared space and some use a custom CIDR?

docs/development/design/shared-cidr.md
```diff
@@ -0,0 +1,32 @@
...
+4. If in the future we do start running out of address spaces and are we are willing to give up on #2, then we allocate additional CG-NAT address spaces.  Another option would be to use an IPv6 address space instead.
+5. This also avoids needing a network administrator to pre-plan network CIDRs and how they need to be allocated to all the devices.
+6. We can still allow an Organization to configure a different address space than the shared CG-NAT address space.  This would give control back to network administrators that want more control over the IP allocations.
+7. Just because organizations share an address space does not mean the devices will be able to interconnect across organizations since WG public keys would not be shared between devices in different organizations.
+8. All Nexodus devices are /32 routed at the current time.  In the future we may want to group devices into subnets to reduce the size of the routing tables that needs to be distributed, but even then we could take the subnet slices from the CG-NAT space.  Yes we should have to make the slices small to avoid wasting unallocated IP addresses.  Still this is not a current problem.
```
russellb, line 22: One impact is that your route to a relay is a single summarized route, but this would make it fall back to a list of /32 routes that all point to the relay. It's the same cost as peering directly, though. Is that what you're thinking?

russellb (commented):
What is the API impact? Is it just that the create-org API call now has the CIDR as optional?

What would that show when it's using the default behavior?  Would it be blank? Or is there some other way to indicate "using the global range, not an org-specific one" ?

docs/development/design/shared-cidr.md
chirino: Point 8 was more about the possibility of allocating small subnet slices like a /3 to a group of devices that all local to each other and have one them be the relay.  But I think that's likely to really complicate things.

docs/development/design/shared-cidr.md
chirino: This is the same as asking how would different TMobile customers (aka organizations) use the shared TMobile CG-NAT addresses assigned to their mobile phones.

chirino:
> What is the API impact? Is it just that the create-org API call now has the CIDR as optional?

Correct.  We add a flag, to the create org call, which when set, you get the previous behaviour of using a non-shared IPAM on your custom defined CIDR.

> What would that show when it's using the default behavior? Would it be blank? Or is there some other way to indicate "using the global range, not an org-specific one" ?

#1007 implement it by adding a `private_cidr=true` field the org to indicate it's not the shared CG-NAT ipam that's in use.

docs/development/design/shared-cidr.md
russellb: but if a device is moved from an org using the shared CG-NAT space to an org using a different, private CIDR, then it can't take its IP with it. I believe that's what I was trying to get at.

chirino: yeah.. correct.. movement would only be allowed if your IPAM namespace is not changing.

nerdalert (approved):
LGTM

nerdalert:
SGTM. I feel like this drives network slicing/VPCs and DNS up the priority ladder for next features. Slicing being lower hanging fruit that could probably be done in the same implementation as this. DNS being a heavier lift. Ty!

PR #999, opened 2023-05-09
rpm: Build centos-stream-9 rpms in the right root

russellb:
You'd think `centos-stream-9` is the right mock root to build CentOS
Stream 9 packages, but it is not. When you enable a copr repo on CentOS
Stream 9, it enables the `epel-9` repository for that copr project, not
the `centos-stream-9` repo.

When doing this build with mock, `epel-9` is not sufficient. You must
also specify the base distro that `epel-9` is applied on top of, so
there we must specify `centos-stream+epel-9`.


nerdalert (approved):
LGTM

PR #1002, opened 2023-05-10
nexd: Drop back to using only google stun servers

russellb:
Since expanding the stun server list, we have seen a number of instances
where the results are not what we expect. When we previously used google
only, the results were much more reliable.

This PR drops back to google-only. It is still an improvement over the
past as we now have all 5 servers listed from google instead of 2.


nerdalert (approved):
LGTM

PR #1003, opened 2023-05-10
nexd: Move peer setup msg to debug

russellb:
Recent commits moved other peering setup details to be debug messages,
so this one stands out as not helpful on its own with the debug
counterparts.


nerdalert (approved):
LGTM

PR #1005, opened 2023-05-10
Add logging for symmetric nat discovery

nerdalert:
- The symmetric disocvery logs are only on the initial statrtup. As we add a reconciler we may want to dial the successful log back but in the meantime this will help with debugging. Lets get stable with STUNs before we roll to symmetric checking in a reconcile.
- Removed a duplicate log printing the NAT binding results and made the debug msg the same across all three OSs.

### DIFF
--- a/internal/nexodus/nexodus.go
+++ b/internal/nexodus/nexodus.go
@@ -574,7 +574,6 @@ func (ax *Nexodus) reconcileStun(deviceID string) error {
 			}
 		}
 	}
-	ax.logger.Debugf("reflexive binding is %s", reflexiveIP)
 
 	return nil
 }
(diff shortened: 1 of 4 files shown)


russellb (changes requested):
have some real CI failures to address

internal/nexodus/nexodus.go
```diff
@@ -673,6 +672,19 @@ func (ax *Nexodus) symmetricNatDisco() error {
 		isSymmetric = stunAddr.String() != stunAddr2.String()
 	}
 
+	if stunServer1 != "" {
```
russellb, line 675: Is `stunServer1` right? It looks like you wanted `stunAddr` and `stunAddr2` ?

russellb (changes requested):

internal/nexodus/nexodus.go
nerdalert: 😵‍💫 good catch, here is the output:

```
2023-05-10T22:24:26.481Z	DEBUG	nexodus/nexodus.go:676	first NAT discovery STUN request returned: 71.31.152.254:39802
2023-05-10T22:24:26.481Z	DEBUG	nexodus/nexodus.go:682	second NAT discovery STUN request returned: 71.31.152.254:39802
```

russellb (approved):

PR #1006, opened 2023-05-11
feat: e2e tests now run with a local stun server

chirino:
* Moved common stun related bits to the stun package
* added a stun.Server implementation
* nexd supports a `—stun-server` flag


internal/stun/server.go
```diff
@@ -0,0 +1,134 @@
...
+		if err != nil {
+			if errors.Is(err, net.ErrClosed) {
+				return nil
+			}
+			return err
+		}
+		requestBytes := buf[:n]
+
+		if !stun.IsMessage(requestBytes) {
+			s.Log.Info("Not a STUN request")
```
russellb, line 98: I'd make this a debug message. Otherwise, it could be a low-key DoS by filling up a device's disk with log messages by sending it a stream of tiny UDP packes

cmd/nexd/main.go
```diff
@@ -73,6 +74,11 @@ func nexdRun(cCtx *cli.Context, logger *zap.Logger, mode nexdMode) error {
 		logger.Info("Starting in L4 proxy mode")
 	}
 
+	stunServers := cCtx.StringSlice("stun-server")
+	if stunServers != nil {
+		stun.SetServers(stunServers)
+	}
```
russellb, line 80: How about validating that there are at least 2 since we expect at least 2 for symmetric NAT discovery?  They also need to be different.

cmd/nexd/main.go
```diff
@@ -325,6 +331,12 @@ func main() {
 				EnvVars:  []string{"NEXD_STATE_DIR"},
 				Category: nexServiceOptions,
 			},
+			&cli.StringSliceFlag{
+				Name:     "stun-server",
+				Usage:    "stun server to use discover our endpoint address",
```
russellb, line 336: and adding a note here that we require 2 different stun servers

russellb (changes requested):
nice work! I think this is pretty close, just a few comments

internal/stun/server_test.go
```diff
@@ -0,0 +1,20 @@
...
+import (
+	"fmt"
+	"github.com/nexodus-io/nexodus/internal/stun"
+	"github.com/nexodus-io/nexodus/internal/util"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/zap"
+	"testing"
+)
+
+func TestListenAndStart(t *testing.T) {
```
russellb, line 12: It looks like this is failing in CI

cmd/nexd/main.go
chirino: Well, in the case of e2e tests we only set 1 stun server.

cmd/nexd/main.go
chirino: But we won't have two in the e2e tests... or should we spin up two stun servers?

cmd/nexd/main.go
russellb: how about spinning up 2 on 2 different ports for e2e?

chirino:
updated PR.. we now validate at least two stun servers are configured and we start up two local servers for e2e testing.

russellb:
#1008 to get past the go-licenses failure

russellb (approved):

PR #1009, opened 2023-05-11
Trim stun server list to prevent an empty string being returned

nerdalert:
- resolves getting an empty string as a stun server

### DIFF
--- a/internal/stun/stun.go
+++ b/internal/stun/stun.go
@@ -17,9 +17,12 @@ var (
 )
 
 func init() {
-	servers := strings.Split(stunServersTxtFile, "\n")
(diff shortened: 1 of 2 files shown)


nerdalert:
Resolves an issue of getting empty stun servers. You might have noticed it with warn messages like this on Linux nodes:

```
WARN	nexodus/nexodus.go:229	transaction error: write ip 0.0.0.0->:0: sendmsg: destination address required
```

russellb (approved):
good catch

chirino (approved):
lgtm

chirino:
@nerdalert rebased off main for ya.

PR #1014, opened 2023-05-12
feat: store the wg public and private keys in the state dir.

chirino:


internal/nexodus/keys.go
```diff
@@ -12,18 +14,112 @@ import (
...
+	if err != nil {
+		return err
+	}
+	err = os.WriteFile(privKeyFile, data, privateKeyPermissions)
+	if err != nil {
+		return err
+	}
+
+	// TODO: decide if we should delete the old keys.
+	//err = os.Remove(oldPubKeyFile)
```
chirino, line 103: @russellb do you think we should delete the old keys?

internal/nexodus/keys.go
```diff
@@ -12,18 +14,112 @@ import (
...
+
+	// skip if the new key files exist
+	if canReadFile(pubKeyFile) && canReadFile(privKeyFile) {
+		return nil
+	}
+
+	oldPubKeyFile := "public.key"
+	oldPrivKeyFile := "private.key"
+	if !ax.userspaceMode {
+		switch runtime.GOOS {
```
chirino, line 62: @russellb so now with this change, even if you are in user space mode, we store the keys in the state dir.  I assume this difference is so that a non-root user could write the keys... I imagine that a user space mode user would typically configure the state dir right?

russellb: I wouldn't expect a regular user to have write access to the dir where keys are stored. They shouldn't have read access, either (at least to the private key and any cached tokens).

For userspaceMode, I believe I had it just using the working directory? An improvement is probably a user-specific config dir. `~/.config/nexodus/` would probably work for linux and mac.

internal/nexodus/keys.go
russellb: No, because we don't know where the old keys came from. If we didn't generate them, we need to leave them in place. A user may have other uses of wireguard on the system.

I wonder if it should be a preference order kind of thing.

1. Use `/etc/wireguard` if keys are there.
2. Use `statedir/` if keys are there.
3. Generate new keys and place in `statedir/`

russellb (changes requested):

internal/nexodus/keys.go
chirino: I think we should NOT look at /etc/wireguard at all at some point the future since that is likely to be user managed WG interface configuration.  I see this copy step as a temporary migration path to ensure smooth upgrade to our current users.

chirino: I think we should NOT look at /etc/wireguard at all at some point the future since that is likely to be user managed WG interface configuration.  I see this copy step as a temporary migration path to ensure smooth upgrade to our current users.

russellb: OK, how about:

1. Use statedir keys.
2. If /etc/wireguard has keys AND the public key matches what's already configured on the device, copy it / use it.
3. Generate new keys in statedir.

russellb: and actually, I'm not sure statedir (/var/lib/) is the right place for keys. I think it's /etc. We could use /etc/nexodus instead of /etc/wireguard, though.

For reference: https://refspecs.linuxfoundation.org/FHS_3.0/fhs/index.html

> /var contains variable data files. This includes spool directories and files, administrative and logging data, and transient and temporary files.

and for /var/lib

> This hierarchy holds state information pertaining to an application or the system. State information is data that programs modify while they run, and that pertains to one specific host. Users must never need to modify files in /var/lib to configure a package's operation, and the specific file hierarchy used to store the data must not be exposed to regular users. [[40]](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch05s08.html#ftn.idm236090988912)

russellb: and this page has some commentary about storing user-specific configuration, like the keys when running in non-privileged mode:

https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch03s08.html#homeReferences

It suggests `~/.nexodus/...`. If we have our own dir, we could use it for both state (cached tokens) and config (keys).

internal/nexodus/keys.go
chirino: Should we check to see if the process is running as root, then use the /var/lib dir otherwise use a ~/.config dir?

russellb: I think we need two types of paths for two different types of files.

1. `stateDir` -- Variable State information (cached API tokens for now). This should default to `/var/lib/nexodus` for linux/mac and `C:/nexodus` for Windows.

2. `configDir` -- For nexodus configuration (wireguard keys for now). This should  remain in `/etc`. Creating a new `/etc/nexodus` is fine to have it isolated.

In both cases, if write access to the default paths is not available, fall back to `~/.nexodus` on linux/mac and just fail on Windows, since I don't know what's normal there.

internal/nexodus/keys.go
chirino: I see device keys and user auth tokens to be about the same.  It's all nexd runtime state to me.  Given we generate the keys and the tokens, you could argue it's not configuration so it should NOT live in /etc.

I'm kind of partial to putting all that stuff in:
* if user == root: /var/lib/nexodus
* if user != root: ~/.nexodus

russellb: The keys are associated with device identity, while the token is session-related. We may or may not create the keys. We will always create the cached tokens.

internal/nexodus/keys.go
chirino: I see wireguard keys as being variable state information.  I don't really see why an admin would want to configure the wiregaurd keys.  Wouldn't he be happy with the ones generated by the system?   if he wants preserve identity after a move, then he needs to preserve the state dir.

russellb: https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch05s08.html

> This hierarchy holds state information pertaining to an application or the system. State information is data that programs modify while they run, and that pertains to one specific host. **Users must never need to modify files in /var/lib to configure a package's operation, and the specific file hierarchy used to store the data must not be exposed to regular users.**

chirino: Why does a user need to modify the keys?

internal/nexodus/keys.go
chirino: It would be simpler if we just simply did not support the user creating the keys.  In fact that could introduce more failure conditions.  What if he reuses the same keys across different devices?

russellb: They would have a broken setup if a key was reused. I imagine the two `nexd` instances would fight over resetting the values on the same device.

I don't know how to tell the difference between keys we generated vs. ones placed on disk for us, though, so that'll always be a possibility.

internal/nexodus/keys.go
russellb: They represent the only unique per-device identity. If someone is already using wireguard, they may already have a repository of per-device keys. It's also possible that they may want to create a device ahead of time so they know upfront which IP is associated with each device.

We don't make this very easy to do in practice since neither the UI nor `nexctl` support creating a device.

I have a concrete use case for this. When running `nexd` in a Pod, I put the keys in a Secret, because I need `nexd` to come up as the same device, even after it gets restarted for some reason. Setting this up in advance is easier than trying to grab the keys from the first time `nexd` runs and stashing them.

I'm sure we can do better in the future with per-device API keys, but for now managing keys is the only identity control we have.

chirino:
going to close this out, I'm going to cook a better version that consolidates all internal nexd state into a single file.

PR #1023, opened 2023-05-15
Redo the monitoring stack

dave-tucker:
- Installs operators from OLM since we should (hopefully) get equivalent functionality in the operators installed in OpenShift.
- Install the Grafana Operator by default in the `nexodus-monitoring` namespace
- Adds a `promtail` component that will watch pod logs and forward to Loki. This is running as a Deployment right now, but a DaemonSet would probably make more sense since we need one on every worker node. This is enabled by default in the dev and rosa/qa stacks.
- Adds `make deploy-monitoring-stack` make target which deploys grafana, loki and tempo for use in local development.
- Grafana - at 127.0.0.1.nip.io is only entrypoint into the stack for PromQL, TraceQL and LogQL queries.


nerdalert (approved):
LGTM

PR #1027, opened 2023-05-15
Exit node design doc

nerdalert:


docs/development/design/exit-node.md
```diff
@@ -0,0 +1,145 @@
...
+    - HTTPS port 443 to the API server's destination IP(s). If an exit node becomes unavailable, the exit node client device would no longer be able to reach the device. This would orphan the node and require a user to run nexctl to disable the exit node subscription. To allow an administrator or user of many devices to be able to dis-enroll a device from the web UI, we need to maintain api-server connectivity even if an exit node is unavailable.
+    - DNS UDP 53. This should like be discovered via host discovery. There could potentially be scenarios where a DNS is only available via the Nexd mesh but that would be fairly atypical and should be tackled with Nexodus DNS implementation. This rule could also be narrowed if there was a host entry added to the device the agent is running on.
```
russellb, line 90:
```suggestion
    - DNS UDP 53. This should likely be discovered via host discovery. There could potentially be scenarios where a DNS is only available via the Nexd mesh but that would be fairly atypical and should be tackled with Nexodus DNS implementation. This rule could also be narrowed if there was a host entry added to the device the agent is running on.
```

Should this be both TCP and UDP port 53?

docs/development/design/exit-node.md
```diff
@@ -0,0 +1,145 @@
...
+    - STUN requests, UDP 3478 and 19302 for the RFC defined and Google STUN ports. This is required because once a device is forwarding all traffic through the exit node, the reflexive address for the client device would take on the same reflexive address of the exit node.
```
russellb, line 91: We'll have to document this well. This one seems like it could break other peer-to-peer apps that people intend to use over the top of Nexodus.

Is there a way to mark our own stun requests and only route those differently? That could be tracked as an issue for  later, if needed. I just don't want to forget about it.

docs/development/design/exit-node.md
```diff
@@ -0,0 +1,145 @@
...
+
+Nex_Device_4 -- "eth0(n)" --> STUN
+Nex_Device_4 -- "eth0(n)" --> DNS
+Nex_Device_4 -- "eth0(n)" --> API
+```
+
+## Alternatives Considered
+
+- We could take a namespace approach but that requires moving physical interfaces into namespaces which can be pretty invasive on a machine. The proposed approach makes teardown as simple as removing a netfilter table and two routing tables with a few policy routing rules.
+- For Darwin support we will need to take a path of granular routes which is more invasive and less flexible for out of band services that we can achieve with native Linux routing capabilities, but it could be considered an alternative option. It will require some Darwin networking research to try and send the stun through a policy based route. An exit node being Linux only seems reasonable for now. Below is an example:
```
russellb, line 138: +1 for linux-only support for exit nodes. That seems very reasonable to me.

russellb (changes requested):
fairly minor comments, I think

docs/development/design/exit-node.md
nerdalert: Here is the basic lab scenario commands for the agent side with annotations I have coded up ready to push:

```
# fwmark set on the interface configuration
REMOTE_KEY=6/CwH/gzz9jdKnxeVpWFivycFQMBkniLeBNFQq0+f04=
sudo wg set wg0 peer $REMOTE_KEY allowed-ips 0.0.0.0/0 persistent-keepalive 25  endpoint 54.227.102.183:41823

# This command enables the src_valid_mark functionality for all network interfaces. This is required for routing marked packets properly with WireGuard.
sudo sysctl -q net.ipv4.conf.all.src_valid_mark=1
# This command adds a rule to the routing policy database (RPDB) that says, "If a packet does not have the firewall mark 51820, look up the routing table 51820."
sudo ip -4 rule add not fwmark 51820 table 51820
# This command adds a rule to the RPDB that says, "When looking up the main routing table, ignore the source address prefix length." This is useful for avoiding unnecessary routing cache updates when using policy-based routing.
sudo ip -4 rule add table main suppress_prefixlength 0
# This command adds a default route to the routing table 51820, which says that all traffic should be sent through the WireGuard interface (wg0).
sudo ip -4 route add 0.0.0.0/0 dev wg0 table 51820

# Nftables

# These commands create an nftables table nexodus-stun-mangle and a chain OUTPUT within it. The purpose of this chain is to mangle (alter) packets. It adds a rule that sets the mark 0x4B66 for packets with UDP destination port 19302.
sudo nft add table ip nexodus-stun-mangle
sudo nft add chain ip nexodus-stun-mangle OUTPUT { type route hook output priority mangle\; policy accept\; }
sudo nft add rule ip nexodus-stun-mangle OUTPUT meta l4proto udp udp dport 19302 counter mark set 0x4B66
# These commands create an nftables table nexodus-stun-snat and a chain POSTROUTING within it. The purpose of this chain is to perform source NAT (SNAT) for outgoing packets. It adds a rule that applies masquerading (SNAT) to packets going out through the enp0s1 interface.
sudo nft add table ip nexodus-stun-snat
sudo nft add chain ip nexodus-stun-snat POSTROUTING { type nat hook postrouting priority srcnat\; policy accept\; }
sudo nft add rule ip nexodus-stun-snat POSTROUTING oifname "enp0s1" counter masquerade
# This command adds a default route to the routing table 19302, which says that all traffic should be sent through the enp0s1 interface with a gateway at 192.168.64.1.
sudo ip -4 route add 0.0.0.0/0 table 19302 via 192.168.64.1 dev enp0s1
# This command adds a rule to the RPDB that says, "If a packet has the firewall mark 19302, look up the routing table 19302." This is used to route marked packets with destination port 19302 using the custom routing table created in step 11.
sudo ip -4 rule add fwmark 19302 table 19302

```

docs/development/design/exit-node.md
```diff
@@ -0,0 +1,145 @@
...
+ENDPOINT ADDRESS         PUBLIC KEY
+x.227.102.183:44728     6/CwH/gzz9jdKnxeVpWFivycFQMBkniLeBNFQq0+f04=
+
+$ wget -qO - icanhazip.com
+x.227.102.183
+
+$ sudo nexctl nexd exit-node disable
+Successfully disabled exit node client on this device
+
+# After subscribing to the exit-node, the client is now originating traffic from the exit node.
```
Jahmilli, line 49: minor thing but im guessing this comment should be on line 42 after we've run `sudo nexctl nexd exit-node enable`.

At this point it looks like we've unsubscribed from the exit node

docs/development/design/exit-node.md
```diff
@@ -0,0 +1,145 @@
...
+Nex_Device_4 -- wg0 --> Nexodus_Network_Router_B
+Nex_Device_5 -- wg0 --> Nexodus_Network_Router_B
+Nex_Device_6 -- wg0 --> Nexodus_Network_Router_B
+```
+
+### Client Out of Band Services
+
+- Since all traffic is being tunneled there is a high likelihood that at some point the device will lose connectivity to the exit node and become orphaned. Because of that, we are baking in as many safeguards as possible and options to recover connectivity.
+
+1. Nexodus services will travel out of band via the device's interface containing a default route via Linux policy routing. A new routing table is created with a default route pointing to the same default route in the default namespace. Then using combinations of nftables and rules in the Linux RPDB (routing policy database) traffic destined for the OOB services is mangled and forwarded bypassing the Nexodus driver interface (for example wg0). The initial out of band services are:
```
russellb, line 88:
```suggestion
1. Traffic to Nexodus services will travel out of band via the device's interface containing a default route via Linux policy routing. A new routing table is created with a default route pointing to the same default route in the default namespace. Then, using combinations of nftables and rules in the Linux RPDB (routing policy database), traffic destined for the OOB services is mangled and forwarded, bypassing the Nexodus driver interface (for example wg0). The initial out of band services are:
```

docs/development/design/exit-node.md
```diff
@@ -0,0 +1,145 @@
...
+   - HTTPS port 443 to the API server's destination IP(s). If an exit node becomes unavailable, the exit node client device would no longer be able to reach the device. This would orphan the node and require a user to run nexctl to disable the exit node subscription. To allow an administrator or user of many devices to be able to dis-enroll a device from the web UI, we need to maintain api-server connectivity even if an exit node is unavailable.
```
russellb, line 89: Instead of port 443, I would specify it as the API server port. `nexd` will know if it's on a nonstandard port since it has to be started with the URL to the API server.

russellb:
```suggestion
   - HTTPS port 443 to the API server's destination IP(s). If an exit node becomes unavailable, the exit node client device would no longer be able to reach the device. This would orphan the node and require a user to run nexctl to disable the exit node subscription. To allow an administrator or user of many devices to be able to dis-enroll a device from the web UI, we need to maintain API server connectivity even if an exit node is unavailable.
```

docs/development/design/exit-node.md
```diff
@@ -0,0 +1,145 @@
...
+   - DNS UDP 53. This should like be discovered via host discovery. There could potentially be scenarios where a DNS is only available via the Nexd mesh but that would be fairly atypical and should be tackled with Nexodus DNS implementation. This rule could also be narrowed if there was a host entry added to the device the agent is running on.
```
russellb, line 90:
```suggestion
   - DNS UDP 53. This should likely be discovered via host discovery. There could potentially be scenarios where a DNS is only available via the Nexd mesh but that would be fairly atypical and should be tackled with Nexodus DNS implementation. This rule could also be narrowed if there was a host entry added to the device the agent is running on.
```

docs/development/design/exit-node.md
```diff
@@ -0,0 +1,145 @@
...
+   - STUN requests, UDP 3478 and 19302 for the RFC defined and Google STUN ports. This is required because once a device is forwarding all traffic through the exit node, the reflexive address for the client device would take on the same reflexive address of the exit node.
```
russellb, line 91: I wrote the comment below before I realized we had this discussion in a previous comment thread. It would be great if you could update the doc to include the reasoning why you think this type of traffic should be handled correctly.

----

I don't think this is quite right. This will break any other application that uses STUN.

I think we need to find a way to only send our own stun requests outside of the mesh. Any other application using STUN should go through the exit node.

docs/development/design/exit-node.md
```diff
@@ -0,0 +1,145 @@
...
+2. To enable forwarding if the driver interface is down, A routing table with a default route to the driver interface such as wg0 is created. A forwarding mark is added to the driver of `51820` to denote Wireguard that says, if a packet does not have the firewall mark 51820, perform a lookup in the routing table 51820. Inside of that routing table 51820 is a default route pointing to wg0.
```
russellb, line 92:
```suggestion
2. To enable forwarding if the driver interface is down, A routing table with a default route to the driver interface such as wg0 is created. A forwarding mark is added to the driver of `51820` to denote Wireguard that says, if a packet does not have the firewall mark 51820, perform a lookup in the routing table 51820. Inside routing table 51820 is a default route pointing to wg0.
```

docs/development/design/exit-node.md
```diff
@@ -0,0 +1,145 @@
...
+DNS["DNS UDP Port 53"]
+API["API-Server TCP 443"]
+STUN["STUN Servers UDP [3478,19302]"]
+end
+
+Nex_Device_4 -- "eth0(n)" --> STUN
+Nex_Device_4 -- "eth0(n)" --> DNS
+Nex_Device_4 -- "eth0(n)" --> API
+```
+
```
russellb, line 134:
```suggestion
### Nexd Proxy Mode

TODO - russellb to fill this out later (tl;dr - it should be easy)
```

docs/development/design/exit-node.md
russellb:
```suggestion
- For Darwin support, we will need to take a path of granular routes which is more invasive and less flexible for out-of-band services than we can achieve with native Linux routing capabilities, but it could be considered an alternative option. It will require some Darwin networking research to try and send the stun through a policy-based route. An exit node being Linux only seems reasonable for now. Below is an example:
```

russellb (changes requested):

docs/development/design/exit-node.md
```diff
@@ -0,0 +1,145 @@
...
+This document proposes exit node functionality in Nexodus. An exit node in the context of Nexodus is where traffic from a device will drain to a designated Nexodus agent via a peering and default route associated with the Nexodus exit node.
+
+## Proposal
+
+The initial scope of work if for Linux only. There are two main components, the exit node which originates all traffic leaving the Nexodus mesh and the exit node client that subscribes to the exit node service.
+
+### Exit Node
+
+- An exit-node can be defined from the startup by specifying a child prefix network to advertise by passing the default route cidr notation of `0.0.0.0`.
+- One additional flag is required to make sure a user intends on originating a default route that will forward traffic from a nexd driver interface such as `wg0` to another port such as `eth0`. The `--exit-node` flag also informs the agent to perform source NAT for traffic leaving the Nexodus mesh to allow return traffic to return to the originating exit-node and then sent back to the pre-nat source device.
```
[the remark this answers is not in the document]
nerdalert: Agree, the PR implements:

```
nexd router --exit-node
```

docs/development/design/exit-node.md
nerdalert: Really the only way to do it with PBR is run our own STUN that no other applications would use. Maybe marking traffic somehow but I would need to circle back on that. One thing to limit the issue would be only peel off `19302` the custom Google port since that is what we are currently using and leave the RFC port 3478 alone in the short term.

docs/development/design/exit-node.md
nerdalert: Thanks, done!

docs/development/design/exit-node.md
```diff
@@ -0,0 +1,145 @@
...
+
+```shell
+ nexd router --child-prefix 0.0.0.0/0 --exit-node https://try.nexodus.127.0.0.1.nip.io
+```
+
+### Exit Node Client
+
+- In order for a client to use the exit node, the user will need to explicitly subscribe to the exit node. Since exit nodes can be disruptive to the user's connection by tunnelling all traffic to the exit node, it makes sense to have the extra step needed to explicitly join an exit node. The user can join via `nexctl` or via the web UI.
+
+An example workflow of an agent node looks like the following (open to adding client to the command):
```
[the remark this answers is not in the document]
nerdalert: Added.

docs/development/design/exit-node.md
nerdalert: Resolving as its being discussed later in the review.

docs/development/design/exit-node.md
russellb: I'm OK just documenting this as a caveat of using exit nodes for now

docs/development/design/exit-node.md
```diff
@@ -0,0 +1,157 @@
...
+- To begin with, there will be at most one active exit node per organization (or once a VPC/slice capability is implemented). Looking forward, the exit-node client should have the ability to choose from multiple exit nodes in an organization or slice of an organization and choose from multiple exit nodes based on the user's criteria, such as latency or geolocation. Since these are default routes that overlap being advertised to the mesh, the driver implementation will dictate whether more than one exit node can be concurrently active. In the case of the Wireguard driver, a peer's public key is associated to the route, so it is not possible to overlap prefixes.
+- The exit node startup would look like the following example:
+
+```shell
+nexd router --exit-node
+```
+
+### Exit Node Client
+
+- In order for a client to use the exit node, the user will need to explicitly subscribe to the exit node. Since exit nodes can be disruptive to the user's connection by tunneling all traffic to the exit node, it makes sense to have the extra step needed to explicitly join an exit node. The user can join via `nexctl` or via the web UI.
```
russellb, line 24:
```suggestion
- In order for a client to use the exit node, the user will need to explicitly subscribe to the exit node. Since exit nodes can be disruptive to the user's connection by tunneling all traffic to the exit node, it makes sense to have the extra step needed to explicitly join an exit node. The user can join via `nexctl`, via the web UI, or via a command-line argument to `nexd`.
```

russellb (approved):

Issue #1024, opened 2023-05-15
nexodus-monitoring: Point Grafana to OpenShift for Auth


dave-tucker:
### Describe the Problem Statement

Instead of using our default plaintext credentials, we should use OpenShift auth as documented [here](https://grafana-operator.github.io/grafana-operator/docs/examples/openshift/readme/).
Hopefully that will allow folks with access to ROSA to use the same credentials to access Grafana.

PR #1030, opened 2023-05-16
deploy: OpenShift Auth for Grafana / Promtail as a DaemonSet

dave-tucker:
- Run Promtail as a DaemonSet and attempt to add correct incantations to get it started on OpenShift.
- Authenticate access to Grafana using OpenShift

Fixes: #1024


nerdalert (approved):
LGTM, just some lint errors

PR #1036, opened 2023-05-17
Remove redundant testcontainer teardown

nerdalert:
- The test suite will teardown containers. When debugging the latest dependabot testcontainer release, e2e was hanging on teardown. This will let us bump to the latest testcontainers-go release once merged  #1018 
- Also removed some residual wg commands that were silently failing that I noticed in the runtime logs.


dave-tucker (approved):
LGTM

PR #1042, opened 2023-05-18
docs: Document demo1 from the nexlink design

russellb:
The nexlink design doc (docs/development/design/nexlink.md) described
3 demos to be produced before implementing nexlink itself. This is
demo1.

This shows how to run `nexd proxy` in a Pod to expose a Kubernetes
Service to a Nexodus Organization.


russellb:
I labeled this as `no-tests-required`, but writing an integration test for this scenario is possible. It wouldn't cover any functionality of `nexd` that is not already exercised in the tests. It would just validate that the manifests work as described. It didn't feel worth it to me, but let me know if anyone feels differently.

nerdalert (approved):
LGTM

Issue #1043, opened 2023-05-18
Beef up Symmetric NAT discovery on initialization


nerdalert:
- Add a retry on initial symmetric NAT discovery. If there is a failed STUN lookup on initialization, we currently drop back to assuming the NAT is not symmetric. This will build peers with no connectivity since they cannot be hole-punched.
- Add a warning if there is no relay defined in the ORG that no connectivity is available until a relay is bootstrapped.

russellb:
This should be fixed by doing regular reconciliation on the NAT state and adjusting the configuration accordingly. I'm going to close this as a duplicate of #423, since that issue is generally about fixing changes between NAT types.

Issue #1044, opened 2023-05-18
If a node is behind Symmetric NAT, the reflexive address is flapping between the two returns.


nerdalert:
### Describe the bug

- we don't store both addresses and since they wont match in a symmetric scenario there is a 50/50 chance it will show up as a new address and 
- This also triggers some flapping since the reflexive address is getting advertised to all peers. Someone could do a skip on symmetric peers in `Reconcile` if so inclined.

### How to Reproduce

Log into nex behind symmetric nat

### Expected Behavior

-

russellb:
I mentioned this on Discord, but I think it's a bit worse. Now that we go through a longer list of stun servers, I expect the result is different every time.

I don't think the addresses from stun queries are useful to store at all in the case of symmetric NAT. No other Nexodus device has any use for them. The NAT binding that needs to be discovered and stored is the one to the relay node. We had that behavior back before we switched to stun-based discovery, but I think we need to go back to that now once we've determined that opening a port with stun isn't going to work.

russellb:
>. This also triggers some flapping since the reflexive address is getting advertised to all peers. Someone could do a skip on symmetric peers in Reconcile if so inclined.

with one caveat, if a device was previously not behind symmetric NAT but is now, reconciliation is required.

I'd worry less about skipping things during reconcile and more on ensuring incorrect updates don't get sent out. Trying to be too clever during reconciliation could cause some accidental inconsistencies. I don't optimizing there is worth it unless it's observed as a bottleneck.

russellb:
@nerdalert I think this may be a duplicate of #1001

PR #1045, opened 2023-05-19
Add a retry to failed STUN lookups on initilization

nerdalert:
- Adds a retry to symmetricNatDisco() in case one of the two lookups fail.
- Adds a log message if a node is Symmetric but the org does have a relay.
- Issue #1043

### DIFF
--- a/cmd/nexd/main.go
+++ b/cmd/nexd/main.go
@@ -100,6 +100,7 @@ func nexdRun(cCtx *cli.Context, logger *zap.Logger, mode nexdMode) error {
 		Version,
 		userspaceMode,
 		cCtx.String("state-dir"),
+		ctx,
 	)
 	if err != nil {
 		logger.Fatal(err.Error())
--- a/internal/nexodus/nexodus.go
+++ b/internal/nexodus/nexodus.go
@@ -156,6 +156,7 @@ func NewNexodus(
 	version string,
 	userspaceMode bool,
 	stateDir string,
+	ctx context.Context,
 ) (*Nexodus, error) {
 
 	if err := binaryChecks(); err != nil {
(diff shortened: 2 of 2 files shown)


russellb (commented):
a couple of thoughts / questions:

1) I believe symmetricNatDisco() only called once at startup. so we don't currently detect if a device moves between symmetric and not-symmetric? Is there an issue on this already, or am I thinking about it wrong?

2) Adding more retries will make startup hang for longer on a network where stun is blocked. I guess that's not a big deal right now because it's not going to work anyway if stun doesn't work. Prior to this change, it'll at least error out and give you an error after a minute or so. With this change it's going to hang for much longer before logging any errors. Maybe if it at least logged an error (or warning?) every time a stun request failed, it will at least still give an indication about why it's hanging. 

This reminds me of #1044. When behind symmetric NAT, we need to get our NAT binding from the relay, not STUN. Once that's working, it's the same thing in the case of stun being blocked. nexd should still work with a relay (as long as udp isn't blocked all together). In other words, if stun keeps failing, at some point I think nexd should fall back to concluding we have to assume it's Symmetric NAT. We still have a chance of getting working connectivity there, but not if we're stuck on stun.

What do you think?

russellb (changes requested):
meant to hit "Request changes"

nerdalert:
@russellb I backed down the retryTimer here to 1 second so it will take less time on startup. This also helps with a false positive if one stun doesn't return anything it will attempt to get an accurate return from both. The blocking is definitely not ideal and might be worth re-implementing the STUN lookup with a done timer channel at some point.

russellb:
> @russellb I backed down the retryTimer here to 1 second so it will take less time on startup. This also helps with a false positive if one stun doesn't return anything it will attempt to get an accurate return from both. The blocking is definitely not ideal and might be worth re-implementing the STUN lookup with a done timer channel at some point. Ty for the review.

Thanks for the updates. This lgtm.

russellb (approved):

PR #1050, opened 2023-05-20
Add the ability to get wg information from the interface via nexctl

nerdalert:
- The adds `nexctl nexd peers list` to display peer information from the wg interface
- Also consolidates `nexctl nexd connections` under peers to now be `nexctl nexd peers ping` which is a cleaner UX.
- This also adds back debugging functionality we lost from wg binary removal.

### DIFF
--- a/internal/nexodus/ctlpeers.go
+++ b/internal/nexodus/ctlpeers.go
@@ -1,0 +1,31 @@
+package nexodus
+
+import (
+	"encoding/json"
+	"fmt"
+)
+
+func (ac *NexdCtl) ListPeers(_ string, result *string) error {
+	if ac.ax.userspaceMode {
+		return fmt.Errorf("userspace mode not currently supported")
+	}
+
+	iface := ac.ax.defaultTunnelDev()
+	if iface == "" {
+		return fmt.Errorf("no tunnel interface found")
+	}
+
+	peers, err := DumpPeers(iface)
+	if err != nil {
+		return fmt.Errorf("error getting list of peers: %w", err)
+	}
+
+	peersJSON, err := json.Marshal(peers)
+	if err != nil {
+		return fmt.Errorf("error marshalling list of peers: %w", err)
+	}
+
+	*result = string(peersJSON)
+
+	return nil
+}
--- a/cmd/nexctl/local_unix.go
+++ b/cmd/nexctl/local_unix.go
@@ -50,11 +50,6 @@ func init() {
 				},
 			},
 			{
-				Name:   "connections",
-				Usage:  "Run a test of the nexd peer connectivity (host firewalls may block the ICMP probes)",
-				Action: cmdConnStatus,
-			},
-			{
 				Name:  "get",
 				Usage: "Get a value from the local nexd instance",
 				Subcommands: []*cli.Command{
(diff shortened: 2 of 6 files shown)


nerdalert:
Example output from each command:

- `nexctl nexd peers list`

```
$ sudo nexctl nexd peers list
PUBLIC KEY                                       ENDPOINT                 ALLOWED IPS                     LATEST HANDSHAKE     TRANSMITTED     RECEIVED
6/CwH/gzz9jdKnxeVpWFivycFQMBkniLeBNFQq0+f04=     54.227.102.183:51820     [100.100.0.0/16 200::/64]       None                 313464          0
FNHhpJpeQcpKC+QCLHqwX9CahOo/PQEG8AWNj9RF9lU=     35.173.211.114:56611     [100.100.0.2/32 200::2/128]     None                 312724          0
kv8RizjzRoy1xMJaph7V+JoCFQVtT77tg2mXq6tTQm4=     192.168.64.1:52870       [100.100.0.4/32 200::4/128]     112 seconds ago      40260           36596
```

- `nexctl nexd peers ping`


```
$ sudo nexctl nexd peers ping
HOSTNAME                     WIREGUARD ADDRESS     CONNECTION STATUS
brents-mbook     100.100.0.4           ✓ Reachable
ip-10-10-0-47                100.100.0.2           x Unreachable
ip-10-180-7-194              100.100.0.3           x Unreachable
```

internal/nexodus/ctlpeers.go
```diff
@@ -0,0 +1,31 @@
...
+func (ac *NexdCtl) ListPeers(_ string, result *string) error {
+	var iface string
+
+	if runtime.GOOS == Darwin.String() {
+		iface = darwinIface
+	} else {
+		iface = wgIface
+	}
+
+	peers, err := DumpPeers(iface)
```
russellb, line 18: This won't work for `nexd proxy`. I can think of a few options for this PR. I am OK with any of them. If you choose #1 or #2, please file an issue so I can come back and implement it later.

1. Check `ac.ax.userspaceMode` and return an error.

2. Add a fallback implementation if `userspaceMode` is true that returns data we have in our peer cache. We'd have the IPs and such, but not state like last handshake time.

3. Use the userspace stack configuration interface to get the data you want. For reference, see https://www.wireguard.com/xplatform/#configuration-protocol and search for `IpcGet()`. If you use `IpcGet()` you will have all the data needed.

4. It's possible that the wgctl library would work, too. The README says it can speak the userspace configuration protocol, so I'm wondering if we can get it to work.

internal/nexodus/ctlpeers.go
```diff
@@ -0,0 +1,31 @@
...
+)
+
+func (ac *NexdCtl) ListPeers(_ string, result *string) error {
+	var iface string
+
+	if runtime.GOOS == Darwin.String() {
+		iface = darwinIface
+	} else {
+		iface = wgIface
+	}
```
russellb, line 16: It's probably worth using `defaultTunnelDev()` which implements this same logic, but also handles the `proxy` mode.

russellb (changes requested):

internal/nexodus/ctlpeers.go
nerdalert: Good catch. I am choosing the easy path and opened an issue one of us can circle back on. Mainly because I need the functionality we removed with the wg binary for tshooting CI output in QA this week. Ty.

russellb: Yeah, that's good with me! I figured you would. I just needed to get the options / ideas for how to do it out of my head.

russellb (approved):

PR #1062, opened 2023-05-22
e2e: Improve reliability of TestProxyInvalidConfig

russellb:
This test case occasionally fails when `nexd` did fail, but
`containerExec()` did not return an error. Add a backup check for a
fatal log message as another confirmation that a failure occurred.


nerdalert (approved):
LGTM

PR #1068, opened 2023-05-22
hack: Add more nexd variations to bash history

russellb:
When running the nexd container, it comes with some sample nexd
commands in the bash history. Add variations of the commands that
include setting the log level to debug.


nerdalert (approved):
LGTM

PR #1072, opened 2023-05-23
nexd: Remove some firstTime code for reconciliation

russellb:
In general, to have the most robust reconciliation possible, we should
no reconciliation related code that runs differently the first time
vs. subsequent times. This change moves the code in that direction.

nexd was blocking during startup on completion of the first reconcile
call. It dow defers that to the reconciliation go routine. The only
functional difference is that if the first reconciliation fails, we no
longer exit the process. It will keep trying and logging each failed
attempt, which I believe is more desirable behavior.

The Reconcile() method took a `firstTime` parameter, but the
difference in behavior was very minor. The only functoinal differences
were an additional debug log message and a fatal log message if
setting up the wireguard config failed. Instead, nexd will come back
around and try again on each reconcile attempt.

The Reconcile() method also had a bug in building the newPeers list. A
previous commit, d616d84011b447e9f6cfd2e64cef615211c84902, fixed a
case where peers would be duplicated in the newPeers list.
Unfortunately, that only fixed it in the copy of the code run for the
first reconcile. This change includes the same fix for every
reconciliation run. Now there should no longer ever be duplicates in
that list.

DeployWireguardConfig() no longer takes a `firstTime` parameter. The
intention was to ensure handlePeerRoute() and handlePeerTunnel() are
always called, but it happens anyway, because the newPeers list is the
full list of peers the first time it is called. We now also ensure the
full list is processed again anytime the wireguard configuration fails
and not just the first time.

### DIFF
--- a/internal/nexodus/nexodus.go
+++ b/internal/nexodus/nexodus.go
@@ -5,7 +5,6 @@ import (
 	"crypto/tls"
 	"errors"
 	"fmt"
-	"github.com/nexodus-io/nexodus/internal/stun"
 	"net"
 	"net/http"
 	"net/netip"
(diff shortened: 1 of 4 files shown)


nerdalert (approved):
LGTM! ty

PR #1077, opened 2023-05-25
Store peer configs as a map

russellb:
- nexd: Store peer configs in a map
- nexd: Utilize map in DeployWireguardConfig
- nexd: Reduce some peer config duplication

commit c0c46eb6118df68ad0ff964e383804e762bf2173
Date:   Thu May 25 16:19:28 2023 -0400

    nexd: Reduce some peer config duplication
    
    There were a couple places where some code was slightly duplicated
    when building peer configs. This cleans it up a bit.
    
commit f16f158821659fed9f2667265ccb3b84623fd8a6
Date:   Thu May 25 16:00:25 2023 -0400

    nexd: Utilize map in DeployWireguardConfig
    
    This method was O(n^2) before. We can now get the wireguard config for
    each updated peer as a map lookup instead of a linear search.
    
commit b5c491d9e9eac61f81d55fd4b25a8dd3a1235db8
Date:   Thu May 25 15:57:00 2023 -0400

    nexd: Store peer configs in a map
    
    This is another case where we have a set of peers we need to cross
    reference with another set of peer info. Accessing it by public key
    will be faster than linear searches for larger organizations.
    
    There should be no logic changes in this patch, only a change in the
    storage. Benefits of the new storage will come in future patches as
    more code is introduced that needs to reference the last known config.
    
### DIFF
--- a/internal/nexodus/nexodus.go
+++ b/internal/nexodus/nexodus.go
@@ -129,7 +129,7 @@ type Nexodus struct {
 
 type wgConfig struct {
 	Interface wgLocalConfig
-	Peers     []wgPeerConfig
+	Peers     map[string]wgPeerConfig
 }
 
 type wgPeerConfig struct {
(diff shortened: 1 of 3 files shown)


nerdalert (approved):
LGTM

PR #1080, opened 2023-05-26
Proposal for QUIC based driver for Nexodus

vishnoianil:


docs/development/design/quic-driver.md
```diff
@@ -0,0 +1,156 @@
...
+
+Before diving into the proposed solution, I think it is important to list out the requirements for the QUIC based tunnel driver. These requirements are pretty much derived from wireguard supported features.
+
+- Tunnel creation across
+  - public cloud
+  - private cloud
+  - on-prem (including devices behind home router)
+- Support tunnel creation in above scenarios with
+  - an easy NAT in place (UDP hole punching)
+  - a symmetric NAT in place (through a relay server)
```
nerdalert, line 19: - Do we add mutli-point from a device's tun interface here? 
- I feel like the other bare bones requirement is being able to source a STUN from the device's interface source port e.g. `SO_REUSEPORT`.

docs/development/design/quic-driver.md
```diff
@@ -0,0 +1,156 @@
...
+
+### Major challenges in implementing this approach
+
+- HTTP DATAGRAM "connect-ip" extension is not yet supported by any of the QUIC protocol implementations, so that's a major challenge in implementing this approach.
+- Once "connect-ip" support is available, we need http proxies to support this extension, or we can put effort into adding that support to the proxy of our choice.
+  - Existing proxies that support QUIC are listed [here](https://github.com/quicwg/base-drafts/wiki/Implementations)
+- Overall stability and performance of the implementation will be questionable at the early stage.
+
+## References
+
```
nerdalert, line 89: Great list!

docs/development/design/quic-driver.md
vishnoianil: Yeah, I was planning to add more details about the feature set (mainly discovery), but i thought thats probably getting more into the architecture level details, so i was thinking of adding those details in the architecture document. But i can add some more details here, hopefully that will help clarify it further.

docs/development/design/quic-driver.md
```diff
@@ -0,0 +1,156 @@
...
+B(http-proxy/QUIC)<-.-> IP2( IP PACKET)
+```
+
+This solution can also be extended to support the user space proxy use cases as well. Similar to `wireguard-go`, a QUIC aware HTTP proxy can be implemented to support the use space proxy use cases.
+
+### Major challenges in implementing this approach
+
+- HTTP DATAGRAM "connect-ip" extension is not yet supported by any of the QUIC protocol implementations, so that's a major challenge in implementing this approach.
+- Once "connect-ip" support is available, we need http proxies to support this extension, or we can put effort into adding that support to the proxy of our choice.
+  - Existing proxies that support QUIC are listed [here](https://github.com/quicwg/base-drafts/wiki/Implementations)
```
russellb, line 85: Are these the reasons that you didn't propose using `connect-ip` and instead suggest using QUIC DATAGRAM directly?

russellb, line 85: Are these the reasons that you didn't propose using `connect-ip` and instead suggest using QUIC DATAGRAM directly?

russellb (changes requested):
Thank you for your work done so far! I have some high-level feedback.

I do not support adding support for tunneling IP traffic over QUIC directly to Nexodus. That is a general enough feature that I would like to see as much of it implemented outside of Nexodus as possible. A really good analogy is to look at the architecture of wireguard and how we can consume it. You can configure wireguard tunnels directly and then Nexodus orchestrates that configuration across many hosts. I envision the same architecture for any other tunneling method, including QUIC.

Most of the content here is still relevant. I would just rephrase the document to suggest that the intent is to document an exploration and PoC to determine whether QUIC could work as a future protocol for Nexodus. I would not include a proposal for integration at this time.

docs/development/design/quic-driver.md
vishnoianil: Yes, pretty much.

vishnoianil:
> 



> Thank you for your work done so far! I have some high-level feedback.
> 
> I do not support adding support for tunneling IP traffic over QUIC directly to Nexodus. That is a general enough feature that I would like to see as much of it implemented outside of Nexodus as possible. A really good analogy is to look at the architecture of wireguard and how we can consume it. You can configure wireguard tunnels directly and then Nexodus orchestrates that configuration across many hosts. I envision the same architecture for any other tunneling method, including QUIC.

Yes, agree, and that's how I am implementing it. It has no or minimum impact on the orchestrator, most of the code is contained within nexodus agent code. It will rely on similar format configuration like wireguard. So from the high level architecture perspective, it's very similar to wireguard, except wireguard implementation is kernel based, and QUIC implementations are userspace implementations. Happy to address any questions to clarify the low level details of how QUIC (DATAGRAM) and wireguard are architecturally similar, and QUIC based driver can be implemented in similar way like wireguard based driver. 
> 
> Most of the content here is still relevant. I would just rephrase the document to suggest that the intent is to document an exploration and PoC to determine whether QUIC could work as a future protocol for Nexodus.

After the current PoC I am pretty confident that QUIC can be be used as a alternate driver for Nexodus. I am happy to add more specific details if you think that will help understand the how's and what's better.

> I would not include a proposal for integration at this time.
Sure, but wondering if the integration proposal will help clarify the first concerned you raised (regarding the tunneling architecture).

russellb:
I envision three design-doc-level deliverables.

1. a doc that describes why alternative transports are interesting to Nexodus and explores whether QUIC is a viable alternative. It should have supporting references to other projects and our own PoC code. It must discuss both the tunneling mechanism and the security architecture.  The document should also list alternatives considered. We looked at ipsec, even had a PoC about it, so I'd like that referenced in our write-up that explains our survey of future transport options. It should be clear why QUIC is considered the best choice.

2. a design for a project that supports tunneling IP traffic that is as easy as wireguard. This is completely separate from Nexodus. One way to look at it is consider the ease of setting up IP tunneling between hosts manually with wireguard. I would like to see a project that provides a similar experience using QUIC. It can reference #1, but should talk about the first steps about how we make this usable by anyone. 

3. Once #2 is far enough along, we could have a proposal for integration with Nexodus with a library similar to wireguard-go. Nexodus should be seen as an orchestrator for configuring core technology built as proposed in #2.

My suggestion is that this doc should be updated to focus on the scope proposed in #1. Do you agree with the proposed scope of the first doc? I think is important before we get back into any details.

PR #1083, opened 2023-05-27
Allow nexd to choose an organization to connect to

### LINKED ISSUE #1082: Allow nexd to choose which organization to connect to
A user can be a member of multiple organizations, but `nexd` doesn't support it. It requires the user to be a member of exactly one org or it will exit with an error.

russellb:
- nexd: Fix relay of non-default CIDRs
- nexd: Support choosing an organization

commit 410d68231fc8b2da2ec74897b1915cf83bd93136
Date:   Sat May 27 08:48:45 2023 -0400

    nexd: Fix relay of non-default CIDRs
    
    nexd assumed that the valid CIDRs for a relay were the default CIDRs.
    Relays wouldn't have worked for an organization with non-default
    CIDRs.
    
commit 7c7385a7b16d4697f5b7b8f90ca7ce755a197751
Date:   Fri May 26 21:13:01 2023 -0400

    nexd: Support choosing an organization
    
    By default, nexd will pick the organization for you as long as the
    user is only a member of a single organization. Otherwise, an
    organization must be specified as a command line parameter or env var.
    
    Closes #1082
    
commit c43f6d57137cdd1477e1d2625237ca9388333f05
Date:   Sat May 27 17:38:13 2023 -0400

    e2e: Add TestChooseOrganization
    
    Add an e2e test that validates that you can specify which organization
    nexd should connect to.
    
commit 72b229ac2d87f95f6dd8bd0cb611f53a40030cb1
Date:   Sat May 27 17:45:11 2023 -0400

    ui: Allow creating organizations
    
    Turn on creating organizations through the UI.
    
commit 3bd996b57a932303b113079cc37cf8f10dec6f30
Date:   Sat May 27 18:19:13 2023 -0400

    fflags: Enable multi-org support by default
    
    Multi-org support has been behind a feature flag since it was not
    fully implemented. Previous commits added UI support and nexd support,
    so it seems fine to turn it on everywhere now. I left the feature flag
    in place in case something goes wrong and we want to turn it back off.
    
### DIFF
--- a/cmd/nexd/main.go
+++ b/cmd/nexd/main.go
@@ -3,13 +3,14 @@ package main
 import (
 	"context"
 	"fmt"
-	"github.com/nexodus-io/nexodus/internal/stun"
 	"os"
 	"os/signal"
 	"runtime"
 	"sort"
 	"sync"
 	"syscall"
+
+	"github.com/nexodus-io/nexodus/internal/stun"
 
 	"github.com/nexodus-io/nexodus/internal/nexodus"
 	"github.com/nexodus-io/nexodus/internal/util"
@@ -101,6 +102,7 @@ func nexdRun(cCtx *cli.Context, logger *zap.Logger, mode nexdMode) error {
 		userspaceMode,
 		cCtx.String("state-dir"),
 		ctx,
+		cCtx.String("org-id"),
 	)
 	if err != nil {
 		logger.Fatal(err.Error())
@@ -335,6 +337,13 @@ func main() {
 				Name:     "stun-server",
 				Usage:    "stun server to use discover our endpoint address.  At least two are required.",
 				EnvVars:  []string{"NEXD_STUN_SERVER"},
+				Category: nexServiceOptions,
+			},
+			&cli.StringFlag{
+				Name:     "org-id",
+				Usage:    "Organization ID to use when registering with the nexodus service",
+				EnvVars:  []string{"NEXD_ORG_ID"},
+				Required: false,
 				Category: nexServiceOptions,
 			},
 		},
--- a/deploy/nexodus/base/apiserver/deployment.yaml
+++ b/deploy/nexodus/base/apiserver/deployment.yaml
@@ -98,11 +98,6 @@ spec:
                 configMapKeyRef:
                   name: apiserver
                   key: NEXAPI_TRACE_INSECURE
-            - name: NEXAPI_FFLAG_MULTI_ORGANIZATION
-              valueFrom:
-                configMapKeyRef:
-                  name: apiserver
-                  key: NEXAPI_FFLAG_MULTI_ORGANIZATION
             - name: NEXAPI_REDIRECT_URL
               valueFrom:
                 configMapKeyRef:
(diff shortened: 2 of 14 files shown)


russellb:
I tested this manually so far. I created a 2nd org using `nexctl`.  Then:

1. I validated that `nexd` exits with an error if you don't specify an org. It also helpfully includes the org names and Ids in the log before exiting.
2. I validated that I can join either org and get an IP associated with that org.

nerdalert:
@russellb Good PR, brings up an important topic.

Wondering if we should have a default flag on org default resources. We are getting into a similar conversation we had with a default zone vs. user specified zones. The cleanest UX is still giving the CLI options to nexd but we are heading down a broader path of loading up the CLI.

At the least, these are the big categories of optionality a user would be offered if an organization is scaled out to have an EC2 like experience.

```
sudo nexd \
--org-id 12345678-1234-1234-1234-123456789012  \
--vpc-id 9876543-1234-1234-1234-123456789012  \
--security-group 1111111111-1234-1234-1234-123456789012  \
https://try.nexodus.io
```

A default column in the tables could be as simple as the first one created. The alternative is if someone experiments and creates a new resource, they are permanently locked out of the following user experience.

```
sudo nexd \
https://try.nexodus.io
```

Other thing worth considering is giving the option to use a name when calling the org. I know names aren't unique per se, but I think scopes would allow for non-overlapping org names globally unless the user explicitly specified an overlapping org that he/she owns.  It could be a check and a message back to the user to use the UUID if they overlap. Would also be for making names unique in an org based on the user scoping.

```
sudo nexd \
--org-id red  \
--vpc-id red-backend  \
--security-group red-airgap \
https://try.nexodus.io
```

Plenty of this is beyond the scope of this PR but the framework could be laid out by starting with orgs.

russellb:
Thanks, @nerdalert !  Great topic.  My initial reaction:

1. I like the idea of having one as the default so `--org-id` (or similar) is never a hard requirement.
2. I wonder if the default should just be the user's personal org and skip exposing it as an option.
3. I also wonder if the user's default / personal org should be restricted to only the user and not allow invitations to others.
4. If we expand the hierarchy one more level (so networks are a construct inside of an org instead of the org itself), then inviting other users into a specific network seems OK.
5. A shorter point: a default to a network that will only ever have devices owned by the same user seems like a safe and reasonable default.

nerdalert:
1.  Agree
2. I think we need to expose all of them. The original use case in your PR where a user is a member of two orgs is very valid. A generic use case being a bunch of nodes that need administrative access by various team members, could be an org.
3. I would say make it shareable since the use cases is more enterprise focused than individual users with no relation. SaaS vs. Enterprise usage feels a bit at odds with each other on this one.
4. I think the easy answer is can I do this with EC2. Can I have an org with a bunch of VPCs and security groups that everyone in the org has access to? Also feels slightly orthogonal to the everyone gets an org approach but I much prefer an EC2 feel here. consumer vs. business environment.
5. I tend to think of an Org as a collection of users in practice. If a user has the rights to come into an org and consume the resources, they should likely be permitted to use any VPC in the org unless they are not in a permit list for that VPC. If a user stands this up for his environment, its not going to be 50 organizations for 50 users.

I feel like we really need to start exposing these knobs in a UI to start experimenting with different workflows. I definitely lean more towards the sharing of resources rather than the isolation due to how VPNs and Cloud resources are accessed and shared today.

nerdalert (approved):

PR #1095, opened 2023-05-29
nexd: Choose a default organization

### LINKED ISSUE #1094: Choose a default organization if no org is specified to nexd
#1083 added support in `nexd` for a user that belongs to multiple organizations. When a user is a member of multiple orgs, `nexd` requires an org to be specified as a command line argument or env var. In some comments on the PR, @nerdalert and I discussed some possible UX improvements. In particular, we discussed the idea of having a default organization that is chosen if none is specified, so that the simplest form of running nexd still works, `nexd URL`.

russellb:
A recent change added support in nexd for users that are members of
multiple organizations. This change restores the original simple form
of running nexd: `nexd URL`.

If a user is a member of multiple organizations and no org ID was
specified, choose the default organization created for the user, which
is the one that has the same name as the user.

In the future we may want to add support in the API and UI for
specifying which organization is the default, but this seems like a
good enough start, and it just keeps the same behavior from before
when only 1 org (the user's org) was supported.

Closes #1094


russellb:
I should update the docs to discuss this behavior and update the e2e test suite to verify that the expected org is chosen by default.

russellb:
marking as a draft while i finish adding test cases and docs

PR #1099, opened 2023-05-30
tests: Resolve ipam port conflict

russellb:
The unit tests fail when I try to run them on the same host as a
running kind stack development environment because port 9090 is
already in use. Change the port used in the test suite.


nerdalert (approved):
LGTM

Issue #965, opened 2023-05-03
Peering failure due to invalid local connectivity assumption


russellb:
I ran into a case where two instances of `nexd` failed to peer. The reflexive addresses for these two devices is the same, so they know they're behind the same outer-layer of NAT. However, the local endpoint addresses are not able to reach each other directly. `nexd` chooses to try to peer using these addresses, but it won't work and `nexd` never tries an alternative.

The scenario all resides in my laptop on my home network. One `nexd` instance is running in a container. Another `nexd` instance is running in a Pod within a `kind` cluster. The clusterIP for the Pod is what is reported as the local endpoint, but that is not reachable from outside of the `kind` cluster.

```
                           ▲
                           │
                           │
                           │
                 ┌─────────▼─────────┐
                 │                   │
                 │                   │
                 │  Home Router      │
                 │                   │
                 │                   │
                 └─────────▲─────────┘
                           │
                           │
                           │
                           │
                           │
┌──────────────────────────┼──────────────────────────────────────────────┐
│  Laptop                  │                                              │
│                          │                                              │
│                          │                                              │
│                          │      ┌─────────────────────────────┐         │
│                          │      │  Container running as a     │         │
│                          │      │  Node in a kind cluster     │         │
│   ┌───────────────────┐  │      │                             │         │
│   │                   │  │      │  ┌────────────────┐         │         │
│   │ Container running │  │      │  │ Pod in kind    │         │         │
│   │   nexd            ├──┴──────┤  │ cluster running│         │         │
│   │                   │         │  │ nexd proxy     │         │         │
│   └───────────────────┘         │  └────────────────┘         │         │
│                                 │                             │         │
│                                 └─────────────────────────────┘         │
│                                                                         │
└─────────────────────────────────────────────────────────────────────────┘
```

russellb:
#1431 should address this by falling back to relay connectivity when better methods fail

PR #1118, opened 2023-06-04
nexd: Allow nexd to change peer configs

russellb:
This change should introduce no functional changes, but it reorganizes the code in an important way.

Previously, the only time `nexd` would reconfigure its peers is when it saw a configuration change come in for that peer from the API. The logic is now changed such that the peer will be configured anytime that nexd's peer configuration calculations decide. It may change based on a config change from the API, or it may change because `nexd` decides it should change for some other reason.

A future change will take advantage of this shift to start making peer configuration changes based on nexd's own observations on the connection health to that peer.

Part of #965


nerdalert (approved):
LGTM, double checked with reflexive address changes as well (5g), all looks good. Just needs a rebase.

PR #1133, opened 2023-06-05
e2e: Add additional validation to nft list output

russellb:
Add some additional validation to the output of `nft list` to validate
that it is empty or not-empty as expected.  While working on some
changes, this made a certain failure much more obvious.


nerdalert (approved):
LGTM

PR #1134, opened 2023-06-06
nexd: Adjust security group reconciliation

### LINKED ISSUE #1132: Revisit security group setup during nexd startup
This is something that came up while reviewing #973.

The code merged into main runs slightly different code during startup (calling `processSecurityGroupRules()` directly) from what is run in the reconcile loop (calling `reconcileSecurityGroups()`. I am trying to remove as much startup-flow-specific code as possible so that as much processing as possible is written in a way that it can be done continuously.

I plan to take a look at some refactoring here after merging.

russellb:
Adjust the security group reconciliation logic so that it can also be
relied on at startup. This removes some startup-specific code and just
relies on the reconcile loop to do the right thing.

CLoses #1132

### DIFF
--- a/internal/nexodus/nexodus.go
+++ b/internal/nexodus/nexodus.go
@@ -448,15 +448,6 @@ func (ax *Nexodus) Start(ctx context.Context, wg *sync.WaitGroup) error {
 		}
 	}
 
-	// apply the default security group rules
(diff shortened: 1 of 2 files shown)


nerdalert (approved):
LGTM

Issue #1084, opened 2023-05-27
500 error when creating a duplicate organization


russellb:
Creating duplicate organizations produces a 500 error from the apiserver. 

To reproduce, run this command twice against a dev environment.

```console
nexctl --username admin --password floofykittens --host https://api.try.nexodus.127.0.0.1.nip.io organization create --name test123 --cidr "192.168.80.0/24" --cidr-v6 "211::/64" --description "foo"
```

russellb:
I believe the response here should be a `409 Conflict`.

PR #1137, opened 2023-06-07
apiserver: Returning 409 conflict instead of 500 when creating organi…

Jahmilli:
Returning 409 conflict instead of 500 when creating organization with same name.

Closes https://github.com/nexodus-io/nexodus/issues/1084


Jahmilli:
Will need to have a look at the test (either remove it or look at updating the code) however would like to get a second opinion on it as in its current state it is failing.

I've had a look at setting up a test to validate we do return the correct response code if we create a duplicate organization, however, i get 2 different errors when running the code through kind and testing manually, 
(i.e `nexctl --username admin --password floofykittens --host https://api.try.nexodus.127.0.0.1.nip.io organization create --name test123 --cidr "192.168.80.0/24" --cidr-v6 "211::/64" --description "foo"`) 
and the other approach just via an actual test. 

When running manually (using Postgres as the db), the error I'm getting is `duplicated key not allowed` and this can be asserted against in the code via `if errors.Is(res.Error, gorm.ErrDuplicatedKey) {`. However, when running via a test, the error is `UNIQUE constraint failed: organizations.name` and the condition isn't met resulting in a 500 being returned. 

I believe this is due to using Sqlite for tests instead of Postgres, however, just wanted to know if anyones encountered this before?

russellb:
> The problem i'm seeing is the condition in my implementation `if errors.Is(res.Error, gorm.ErrDuplicatedKey) {` only works when running manually (using Postgres instead of Sqlite), but when running tests, the condition isn't met resulting in a 500 being returned instead of a 409.
> 
> I believe this is due to using Sqlite for tests instead of Postgres, however, just wanted to know if anyones encountered this before?

It sounds like you've correctly diagnosed the issue. I haven't seen this exact thing, but there are definitely several cases where we've hit differences in behavior when using different databases. @chirino had to do a bunch of work trying to get cockroachdb working.

I'm glad you also tested it manually! It would have been easy for us to miss that it didn't work against postgres.

I've got two suggestions for next steps:

1. You'll have to add a second check on the error. I would just leave a comment in the code explaining that you observed the two different errors depending on the underlying db.
2. You've got a unit test, which is great. I would also add an integration test. The integration test suite runs against a full running service. Between the unit test and the integration test, you will have test coverage for both databases.

For the integration test, I suggest writing a cucumber API test. I haven't written one of these tests yet. I would suggest reaching out to @chirino if you have questions. You can see the other examples in `integration-tests/features/`. There are a few files in there covering organizations APIs, but it looks like creating organizations isn't there yet. Hopefully the others give you enough of a head start to see how it works.

russellb:
Good find! We can file an issue to revisit the unit test later once gorm is updated if you'd like. Having a working integration test is enough for me to put the fix in.

russellb:
Assuming all of the CI passes, this looks good to me. I'd like to give @chirino a chance to look over your new test case to make sure I don't miss something with how that should be done.

Thank you!!

chirino (approved):
LGTM!  Good work!

PR #1140, opened 2023-06-07
Change no security group supoport for non-linux to informational

nerdalert:
- Change this from debug to informational to notify the user security groups are only supported on Linux.

### DIFF
--- a/internal/nexodus/nexodus.go
+++ b/internal/nexodus/nexodus.go
@@ -265,7 +265,7 @@ func (ax *Nexodus) Start(ctx context.Context, wg *sync.WaitGroup) error {
 	}
 
 	if runtime.GOOS != Linux.String() {
-		ax.logger.Debug("Security Groups are currently only supported on Linux")
+		ax.logger.Info("Security Groups are currently only supported on Linux")
 	}
 
 	var options []client.Option


nerdalert:
The alternative approach, is to remove the OS check on `reconcileSecurityGroups` but this would incur polling for security groups from these devices which I don't think is worth it imo. This is what it would look like.

```go

// reconcileSecurityGroups will check the security group and update it if necessary.
func (ax *Nexodus) reconcileSecurityGroups(ctx context.Context) {
	//if runtime.GOOS != Linux.String() {
	//	return
	//}

	// lookup this device in the device cache and check if the security-group ID is nil
	existing, ok := ax.deviceCache[ax.wireguardPubKey]
	if ok {
		if existing.device.SecurityGroupId == uuid.Nil.String() {
			ax.securityGroup = nil
			if err := ax.processSecurityGroupRules(); err != nil {
				ax.logger.Error(err)
			}
			return
		}
	}

	// TODO: This needs some scrutiny, ax.securityGroup could be nil, empty with a uuid.nil 0000000-000.. or a valid UUID.
	// are there any scenarios where a group has been created/updated/patched that the device is not
	// updated about via reconcile while still maintaining nil checks?
	if ax.securityGroup == nil {
		return
	}

	// if the security group ID is not nil, lookup the ID and check for any changes
	responseSecGroup, _, err := ax.client.SecurityGroupApi.GetSecurityGroup(ctx, ax.org.Id, ax.securityGroup.Id).Execute()
	if err != nil {
		// if the group ID returns a 404, clear the current rules
		if strings.Contains(err.Error(), securityGroupNotFound.Error()) {
			ax.securityGroup = nil
			if err := ax.processSecurityGroupRules(); err != nil {
				ax.logger.Error(err)
			}
			return
		}
		ax.logger.Debugf("Error retrieving the security group: %v", err)
		return
	}

	if !reflect.DeepEqual(responseSecGroup, ax.securityGroup) {
		if runtime.GOOS == Linux.String() {
			ax.logger.Debugf("Security Group change detected: %+v", responseSecGroup)
		} else {
			ax.logger.Info("Security Group change detected, but security groups currently only supported on Linux")
		}
		if responseSecGroup.Id != ax.securityGroup.Id ||
			!reflect.DeepEqual(responseSecGroup.InboundRules, ax.securityGroup.InboundRules) ||
			!reflect.DeepEqual(responseSecGroup.OutboundRules, ax.securityGroup.OutboundRules) {
			ax.securityGroup = responseSecGroup
			if err := ax.processSecurityGroupRules(); err != nil {
				ax.logger.Error(err)
			}
		} else {
			ax.securityGroup = responseSecGroup
		}
	}
}
```

russellb (approved):

PR #1154, opened 2023-06-08
rpm: Fix building srpm in aarch64 rpm jobs

### LINKED ISSUE #1142: rpm publishing is broken
The `copr-repo` environment is currently failing. The `deploy` workflow is failing to submit new rpm builds to copr.

https://github.com/nexodus-io/nexodus/deployments/activity_log?environment=copr-repo

The reason is that `quay.io/nexodus/mock:fedora-38-aarch64` was built for `linux/arm64`, but the github actions host is `linux/amd64`. These images each have a cached build root inside of them, but ideally they would be multiarch images.

russellb:
When building an srpm to submit to copr for an aarch64 build, just
build the srpm in an x86_64 build root. The arch doesn't matter when
building the srpm, and it'll be more efficient as it will match the
architecture of the host building the srpm.

When setting up the build root for aarch64 for use with `make rpm`,
install `qemu-user-static-aarch64`, as this is needed to build the
aarch64 rpm on an x86_64 host.

Closes #1142

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -137,12 +137,16 @@ jobs:
       matrix:
         include:
           - mock_root: fedora-38-x86_64
+            mock_srpm_root: fedora-38-x86_64
             srpm_distro: fc38
           - mock_root: fedora-38-aarch64
+            mock_srpm_root: fedora-38-x86_64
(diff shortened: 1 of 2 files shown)


russellb:
yolo

Issue #1146, opened 2023-06-07
Deleting a security group triggers a device reconcile


nerdalert:
### Describe the bug

Deleting a security group triggers a device reconcile

### How to Reproduce

Delete the org security group and you will see a peer reconcile

### Expected Behavior

Just need to add an exception for a security group ID change in the device cache reflect.

```go
	updatePeers := map[string]public.ModelsDevice{}
	changed := false
	for _, p := range peerMap {
		existing, ok := ax.deviceCache[p.PublicKey]
		if !ok {
			ax.addToDeviceCache(p)
			updatePeers[p.PublicKey] = p
			changed = true
		} else if !reflect.DeepEqual(existing.device, p) {
			ax.addToDeviceCache(p)
			updatePeers[p.PublicKey] = p
			changed = true
		}
		if p.Relay {
			ax.relayWgIP = p.AllowedIps[0]
		}
	}

	if changed {
		ax.buildPeersConfig()
		if len(updatePeers) > 0 {
			if err := ax.DeployWireguardConfig(updatePeers); err != nil {
				if strings.Contains(err.Error(), securityGroupErr.Error()) {
					return err
				}
				// If the wireguard configuration fails, we should wipe out our peer list
				// so it is rebuilt and reconfigured from scratch the next time around.
				ax.wgConfig.Peers = nil
				return err
			}
		}
	}

```

PR #1155, opened 2023-06-08
Ignore security groups in device reconcile

nerdalert:
- When the security group ID changes on for example, a SG ID deletion, it triggers a peer config change and redeploys the peerings.
- This patch creates a method that will ignores the two device fields, Revision and SecurityGroupId by setting them to a temporary constant and then does a reflect to check for a diff.
- The bug is described further in #1146

### DIFF
--- a/internal/nexodus/nexodus.go
+++ b/internal/nexodus/nexodus.go
@@ -702,7 +702,7 @@ func (ax *Nexodus) Reconcile() error {
 	newLocalConfig := false
 	for _, p := range peerMap {
 		existing, ok := ax.deviceCache[p.PublicKey]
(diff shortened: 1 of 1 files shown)


nerdalert:
This may not be the right approach, if there are plans for the revision field to be used for any change tracking in device reconciliation.

russellb (approved):
There's one other field that may change that could be ignored: Hostname. That should be pretty rare, though, and it's not harmful to do an extra reconciliation on a peer. I would expect it may just be some debug log noise. If it's anything more than that, I'd call that another bug.

russellb:
The rpm job failure will be fixed by #1156

PR #1156, opened 2023-06-08
fix SRPM variable name in Makefile

Jahmilli:
Updated name of variable as the variable declared was not the same as what was being referenced. 

This resulted in the following issue in the build, where we echoed an empty string.

```bash
docker run --name mock --rm --privileged=true -v /home/runner/work/nexodus/nexodus:/nexodus quay.io/nexodus/mock:$(echo  | cut -f2 -d'+') \
47
    mock --buildsrpm -D "_commit 4c837fd" --resultdir=/nexodus/dist/rpm/mock --no-clean --no-cleanup-after \
48
    --spec /nexodus/contrib/rpm/nexodus.spec --sources /nexodus/dist/rpm/ --root 
49
docker: invalid reference format.
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -331,8 +331,10 @@ jobs:
       matrix:
         include:
           - mock_root: fedora-38-x86_64
+            srpm_mock_root: fedora-38-x86_64
             srpm_distro: fc38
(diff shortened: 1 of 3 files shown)


russellb (approved):
oops! my fault. thank you for the fix!

Issue #1102, opened 2023-05-30
Investigate simple client-side load balancing for proxy mode


russellb:
Having a simple client-side load balancing would be very helpful for egress proxy rules. I'm thinking either round-robin or random and that's it. Anything fancier can be implemented with a more full-featured proxy on top.

PR #1158, opened 2023-06-08
proxy: support simple client-side load balancing for proxy mode

chirino:
closes: #1102

### DIFF
--- a/docs/user-guide/nexd-proxy.md
+++ b/docs/user-guide/nexd-proxy.md
@@ -80,6 +80,10 @@ flowchart TD
 
 Since UDP is a connectionless protocol, `nexd proxy` must maintain its own state for each UDP flow to ensure that return traffic is forwarded appropriately. These flows time out after 60 seconds of inactivity.
 
+### Proxy Load Balancing
+
+If multiple rules for share the same protocol and listener port, then the proxy will use simple round-robin load balancing of connection across the destination hosts and ports.
+
 ### Managing Rules with Nexctl
 
 In addition to configuring rules as command line flags, `nexctl` can be used to dynamically add or remove proxy rules. Rules that are added dynamically are persisted across `nexd proxy` restarts.
--- a/docs/user-guide/nexd-proxy.md
+++ b/docs/user-guide/nexd-proxy.md
@@ -82,7 +82,7 @@ Since UDP is a connectionless protocol, `nexd proxy` must maintain its own state
 
 ### Proxy Load Balancing
 
-If multiple rules for share the same protocol and listener port, then the proxy will use simple round-robin load balancing of connection across the destination hosts and ports.
+If multiple rules share the same protocol and listener port, then the proxy will use simple round-robin load balancing of connections across the destination hosts and ports.
 
 ### Managing Rules with Nexctl
 
--- a/integration-tests/proxy_test.go
+++ b/integration-tests/proxy_test.go
@@ -774,3 +774,116 @@ func TestProxyNexctlConnections(t *testing.T) {
 	require.False(strings.Contains(out, "Unreachable"))
 	require.True(strings.Contains(out, "Reachable"))
 }
+
+// TestProxyEgress tests that nexd proxy can load balance between multiple egress rules on the same listen port
+func TestProxyEgressLoadBalancer(t *testing.T) {
+	t.Parallel()
+	helper := NewHelper(t)
+	require := helper.require
+	ctx, cancel := context.WithTimeout(context.Background(), 120*time.Second)
+	defer cancel()
+
+	password := "floofykittens"
+	username, cleanup := helper.createNewUser(ctx, password)
+	defer cleanup()
+
+	// create the nodes
+	node1, stop := helper.CreateNode(ctx, "node1", []string{defaultNetwork}, enableV6)
+	defer stop()
+	node2, stop := helper.CreateNode(ctx, "node2", []string{defaultNetwork}, enableV6)
+	defer stop()
+
+	// start nexodus on the nodes
+	helper.runNexd(ctx, node1, "--username", username, "--password", password, "relay")
+	err := helper.nexdStatus(ctx, node1)
+	require.NoError(err)
+
+	node1IP, err := getTunnelIP(ctx, helper, inetV4, node1)
+	require.NoError(err)
+
+	helper.runNexd(ctx, node2, "--username", username, "--password", password, "proxy",
+		"--egress", fmt.Sprintf("tcp:80:%s", net.JoinHostPort(node1IP, "8080")),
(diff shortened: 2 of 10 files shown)


integration-tests/proxy_test.go
```diff
@@ -774,3 +775,166 @@ func TestProxyNexctlConnections(t *testing.T) {
...
+		return true, nil
+	})
+	require.NoError(err)
+	require.True(success)
+	_, _ = helper.containerExec(ctx, serverNode, []string{"killall", "python3"})
+	_, _ = helper.containerExec(ctx, serverNode, []string{"killall", "udpong"})
+	wg.Wait()
+}
+
+func TestProxyLoadBalancer(t *testing.T) {
```
russellb, line 906: Awesome job on these tests. I love how you got the coverage without duplicating code.

internal/nexodus/proxyrule.go
```diff
@@ -0,0 +1,138 @@
...
+	stored bool
+}
+
+type HostPort struct {
+	host string
+	port int
+}
+
+func (hp HostPort) String() string {
+	return fmt.Sprintf("%s:%d", hp.host, hp.port)
```
russellb, line 68: need to use `JoinHostPort()` here. That will ensure that IPv6 addresses are handled correctly.

https://pkg.go.dev/net#JoinHostPort

internal/nexodus/proxyrule.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+func (rule ProxyRule) String() string {
+	// protocol:port:destination_ip:destination_port
+	return fmt.Sprintf("%s:%d:%s:%d", rule.protocol, rule.listenPort, rule.dest.host, rule.dest.port)
```
russellb, line 73: need to use JoinHostPort() here for `dest.host` + `dest.port`. That will ensure that IPv6 addresses are handled correctly.

https://pkg.go.dev/net#JoinHostPort

integration-tests/proxy_test.go
```diff
@@ -774,3 +775,166 @@ func TestProxyNexctlConnections(t *testing.T) {
...
+				return "127.0.0.1"
+			},
+			dialIP: func(node2IP string) string {
+				return node2IP
+			},
+			workloadPlacer: func(node1, node2 testcontainers.Container) (serverNode testcontainers.Container, clientNode testcontainers.Container) {
+				return node2, node1
+			},
+		},
+	}
```
russellb, line 934: How about a couple more test cases that use IPv6 addresses? I spotted a couple of IPv6 address-handling bugs in the code, so I think it's worth adding the coverage here. I'm not sure if these tests would hit what I saw, but it's still good to add, I think.

internal/nexodus/proxyrule.go
```diff
@@ -0,0 +1,138 @@
...
+		return emptyRule, err
+	}
+
+	port, err := parsePort(parts[1])
+	if err != nil {
+		return emptyRule, err
+	}
+	if port < 1 || port > 65535 {
+		return emptyRule, fmt.Errorf("Invalid port (%d)", port)
+	}
```
russellb, line 106: This is probably old code that you just moved, but this should probably move into `parsePort()` if you don't mind while you're in here anyway.

internal/nexodus/proxyrule.go
```diff
@@ -0,0 +1,138 @@
...
+		return emptyRule, fmt.Errorf("Invalid destination host:port (%s): host cannot be empty", destHostPort)
+	}
+
+	destPort, err := parsePort(destPortStr)
+	if err != nil {
+		return emptyRule, err
+	}
+	if destPort < 1 || destPort > 65535 {
+		return emptyRule, fmt.Errorf("Invalid destination port (%d)", destPort)
+	}
```
russellb, line 125: moving the check into `parsePort()` would remove the duplication here.

russellb (changes requested):
looks awesome! relatively minor comments

internal/nexodus/userspace_proxy.go
```diff
@@ -564,25 +437,37 @@ func (proxyConn *udpProxyConn) readFromUDP(buffer []byte) (int, *net.UDPAddr, er
...
+	if proxyConn.udpProxy.proxy.key.ruleType == ProxyTypeEgress {
 		_, err = proxyConn.udpProxy.conn.WriteToUDP(buffer[:n], proxyConn.clientAddr)
 	} else {
 		_, err = proxyConn.udpProxy.goConn.WriteTo(buffer[:n], proxyConn.clientAddr)
 	}
 	return err
 }
 
+func (proxy *UsProxy) NextDest() HostPort {
+	proxy.mu.RLock()
```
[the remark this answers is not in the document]
chirino: Good catch.  I think we can still use the atomic counter.  The read lock will protect us from the proxy.rules array from changing, and the atomic will give us a copy of the incremented counter.   So we should be able to index into it ok.

docs/user-guide/nexd-proxy.md
```diff
@@ -80,6 +80,10 @@ flowchart TD
 
 Since UDP is a connectionless protocol, `nexd proxy` must maintain its own state for each UDP flow to ensure that return traffic is forwarded appropriately. These flows time out after 60 seconds of inactivity.
 
+### Proxy Load Balancing
```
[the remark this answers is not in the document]
chirino: ok see #1160

internal/nexodus/proxyrule.go
chirino: FYI: messages for `error` types should always start lower case since you often prepend more context  to them in wrapper errors.  So I'll change that too.

internal/nexodus/proxyrule.go
chirino: moving...

internal/nexodus/proxyrule.go
chirino: will do

internal/nexodus/proxyrule.go
chirino: will do

docs/user-guide/nexd-proxy.md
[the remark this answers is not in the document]
russellb: thanks!

internal/nexodus/proxyrule.go
russellb: ah, didn't know, thanks makes sense - thanks!

internal/nexodus/userspace_proxy.go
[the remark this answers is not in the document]
russellb: good point that the combo allows the use of the read lock.

integration-tests/proxy_test.go
chirino: will do.

russellb (approved):
great work!

Issue #1162, opened 2023-06-09
Add a handler call that will install a rule to drop all traffic


nerdalert:
- Since we are implicitly allowing traffic in the initial phase we need to be able to reset the tables back to default with an empty rule being installed.
- To add an explicit drop I think have a snowflake keyword rule of `drop` or `deny`

Examples:
```
# Allows all traffic (current functionality)
./nexctl \
    --host https://api.try.nexodus.127.0.0.1.nip.io --username admin --password floofykittens security-group update \
    --name="default" --description="security group testing" \
    --inbound-rules='' \
    --outbound-rules='' \
   --security-group-id="${SECURITY_GROUP_ID}" \
   --organization-id="${ORGANIZATION_ID}"

# Translates to the default table of:

table inet nexodus {
	chain nexodus-inbound {
		type filter hook input priority filter; policy accept;
		ct state established,related iifname "wg0" counter packets 0 bytes 0 accept
	}

	chain nexodus-outbound {
		type filter hook input priority filter; policy accept;
	}
}

# Deny all traffic (new functionality)
./nexctl \
    --host https://api.try.nexodus.127.0.0.1.nip.io --username admin --password floofykittens security-group update \
    --name="default" --description="security group testing" \
    --inbound-rules='deny' \
    --outbound-rules='deny' \
   --security-group-id="${SECURITY_GROUP_ID}" \
   --organization-id="${ORGANIZATION_ID}"

# Translates to the default table of:

table inet nexodus {
	chain nexodus-inbound {
		type filter hook input priority filter; policy accept;
		iifname "wg0" counter packets 0 bytes 0 drop
	}

	chain nexodus-outbound {
		type filter hook input priority filter; policy accept;
                iifname "wg0" counter packets 0 bytes 0 drop
	}
}
```

- As soon as a user adds any permit rule a deny is added after the permit, the only case not handled is a deny all which this issue is to address.

PR #1163, opened 2023-06-09
Accept empty inbound/outbound SG rules

nerdalert:
- Since we are implicitly allowing traffic in the initial phase we need to be able to reset the tables back to default with an empty rule being installed.
- The only limitation here is denying all traffic to be addressed in a future PR described in #1162


russellb (approved):
I'd like to see this covered in a test case, but lgtm

nerdalert:
> I'd like to see this covered in a test case, but lgtm

added to the issue, ty ty

PR #1166, opened 2023-06-09
added custom zap logger config

### LINKED ISSUE #1152: Timestamps in nexd logs prevent log summarization
Here is a sample `nexd` log message:

```
{"level":"info","ts":1686166202.684511,"caller":"nexodus/wg.go:233","msg":"Removed peer with key 5l0AHeRRGW1xeTeehW3Ab3P2cTfarCBTBZwGZV/9XSM="}
```

The `ts` field in the log message prevents a logging daemon from summarizing repeated messages. We should leave timestamps to logging daemons to provide.

akoserwal:
Closes #1152

Added a custom zap custom configuration to remove `ts` from the logs

### DIFF
--- a/cmd/nexd/main.go
+++ b/cmd/nexd/main.go
@@ -157,12 +157,14 @@ func main() {
 	if debug != "" {
 		logCfg := zap.NewDevelopmentConfig()
 		logLevel = &logCfg.Level
+		logCfg.EncoderConfig.TimeKey = ""
 		logger, err = logCfg.Build()
 		logger.Info("Debug logging enabled")
 	} else {
 		logCfg := zap.NewProductionConfig()
 		logLevel = &logCfg.Level
 		logCfg.DisableStacktrace = true
+		logCfg.EncoderConfig.TimeKey = ""
 		logger, err = logCfg.Build()
 	}
 	if err != nil {


akoserwal:
@russellb

cmd/nexd/main.go
```diff
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"fmt"
+	"go.uber.org/zap/zapcore"
```
russellb, line 6: Can you move this down to be with the import of `zap` ?

cmd/nexd/main.go
```diff
@@ -148,20 +149,39 @@ func nexdRun(cCtx *cli.Context, logger *zap.Logger, logLevel *zap.AtomicLevel, m
...
 	if debug != "" {
-		logCfg := zap.NewDevelopmentConfig()
-		logLevel = &logCfg.Level
+		logCfg := NewCustomConfig()
 		logger, err = logCfg.Build()
 		logger.Info("Debug logging enabled")
 	} else {
-		logCfg := zap.NewProductionConfig()
-		logLevel = &logCfg.Level
+		logCfg := NewCustomConfig()
```
russellb, line 184: How about still using zap's default production vs. debug configurations, but just setting `logCfg.EncoderConfig.TimeKey = ""` ?

russellb (changes requested):
Thanks for looking at this! I have a couple of small suggestions.

cmd/nexd/main.go
akoserwal: Yes, it makes sense. I update the PR.

russellb (approved):
Thank you!

PR #1168, opened 2023-06-09
nexd: fix a seg fault

russellb:
commit b276f998e92ffff3ae4ad36c29c145fb7d051b06
Date:   Fri Jun 9 18:29:56 2023 -0400

    nexd: fix a seg fault
    
    I observed a seg fault in a CI run against PR #1007 on the following
    line:
    
    ```go
    proxy.logger.Debugf("Connection from %s closed: %v", conn.RemoteAddr().String(), err)
    ```
    
    Based on conventions of the std library, as well as reading the code
    in wireguard-go, it appears that `conn` should never be `nil` here, as
    we only get here after `Accept()` returns `nil` for an error.
    
    However, it is possible that `conn.RemoteAddr()` returns `nil`. Here's
    the corresponding code in the `net` library:
    
    ```go
    func (c *conn) RemoteAddr() Addr {
            if !c.ok() {
                    return nil
            }
            return c.fd.raddr
    }
    ```
    
    and to further explain when it may be `nil`, here's `c.ok()`:
    
    ```go
    func (c *conn) ok() bool { return c != nil && c.fd != nil }
    ```
    
    Just in case the connection gets put in a bad state in our connection
    handler, go ahead and save off the remote address early for use in the
    log message later, if available.
    
### DIFF
--- a/internal/api/public/api_organizations.go
+++ b/internal/api/public/api_organizations.go
@@ -152,6 +152,17 @@ func (a *OrganizationsApiService) CreateOrganizationExecute(r ApiCreateOrganizat
 			newErr.model = v
 			return localVarReturnValue, localVarHTTPResponse, newErr
 		}
+		if localVarHTTPResponse.StatusCode == 409 {
+			var v ModelsConflictsError
+			err = a.client.decode(&v, localVarBody, localVarHTTPResponse.Header.Get("Content-Type"))
+			if err != nil {
+				newErr.error = err.Error()
+				return localVarReturnValue, localVarHTTPResponse, newErr
+			}
+			newErr.error = formatErrorMessage(localVarHTTPResponse.Status, &v)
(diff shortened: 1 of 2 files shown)


russellb:
I looked in wireguard-go and didn't find a code path where this is possible on an ingress proxy. Maybe it's possible on an egress proxy using the go std library, though.

russellb:
Discussion on this bug suggests it shouldn't be possible for `conn` to be `nil` using the std library if `err` is `nil`.

https://github.com/golang/go/issues/37663

For reference, here's the line from nexd in question:

```
proxy.logger.Debugf("Connection from %s closed: %v", conn.RemoteAddr().String(), err)
```

It actually seems more likely that `conn.RemoteAddr()` returned `nil`.

chirino (approved):
LGTM

PR #1169, opened 2023-06-11
rpm: Rename rpm build artifacts

russellb:
The rpms built in the build workflow were in a build artifact called
`nexodus`. It was not obvious at all that this was an archive of rpms
built from this PR. Rename the artifact to "nexodus-rpms". This lets
you install an rpm of a given PR.

I also noticed there was a typo in the job configuration that caused
`el9` to get left out of the rpm filenames for centos stream 9.


nerdalert (approved):
LGTM

Issue #1171, opened 2023-06-12
Implement device metadata API


russellb:
Implement an API for adding metadata to devices as proposed in the nexlink design document.

https://docs.nexodus.io/development/design/nexlink/#nexodus-api-changes-required

russellb:
this was merged

PR #1172, opened 2023-06-12
docs: Propose API support for device metadata

russellb:
Issue #1171

### DIFF
--- a/docs/development/design/device-metadata.md
+++ b/docs/development/design/device-metadata.md
@@ -8,31 +8,29 @@ The [nexlink design proposal](https://docs.nexodus.io/development/design/nexlink
 
 ## Proposal
 
-Instead of changing the Device table, we will create a new table called DeviceMetadata. I propose keeping it separate from the Device table while we work through how we want this to work. This table will have the following columns:
+Instead of changing the Device table, we will create a new table called `device_metadata`. I propose keeping it separate from the Device table while we work through how we want this to work. This table will have the following columns:
 
 ```go
-type Base struct {
-    ID        uuid.UUID      `gorm:"type:uuid;primary_key;" json:"id" example:"aa22666c-0f57-45cb-a449-16efecc04f2e"`
+type DeviceMetadata struct {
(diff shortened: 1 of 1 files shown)


docs/development/design/device-metadata.md
```diff
@@ -0,0 +1,55 @@
...
+    ID        uuid.UUID      `gorm:"type:uuid;primary_key;" json:"id" example:"aa22666c-0f57-45cb-a449-16efecc04f2e"`
+    CreatedAt time.Time      `json:"-"`
+    UpdatedAt time.Time      `json:"-"`
+    DeletedAt gorm.DeletedAt `gorm:"index" json:"-"`
+}
+
+type DeviceMetadataInstance struct {
+    Base
+    DeviceID uuid.UUID `json:"device_id"`
+    Key      string    `json:"key"`
```
chirino, line 24: If `Key` must be unique, why not use `DeviceID`+`Key` as the PK instead of `ID`?

docs/development/design/device-metadata.md
```diff
@@ -0,0 +1,55 @@
...
+        // Get all metadata for a device
+        private.GET("/devices/:id/metadata", api.GetDeviceMetadata)
+        // Get a specific metadata key for a device
+        private.GET("/devices/:id/metadata/:key", api.GetDeviceMetadataKey)
+        // Set a specific metadata key for a device
+        private.PATCH("/devices/:id/metadata/:key", api.UpdateDeviceMetadataKey)
+        // Delete all metadata for a device
+        private.DELETE("/devices/:id/metadata", api.DeleteDeviceMetadata)
+        // Delete a specific metadata key for a device
+        private.DELETE("/devices/:id/metadata/:key", api.DeleteDeviceMetadataKey)
```
chirino, line 45: Even in the API access pattern your not using the DeviceMetadataInstance.ID field.  Seems we should drop it.

russellb: That makes sense. Thanks for the feedback!

nerdalert (approved):
LGTM

nerdalert:
@russellb just a few hard tabs from MD lint and it will merge.

PR #1181, opened 2023-06-12
nexd: Don't attempt security groups in proxy mode

### LINKED ISSUE #1180: Update security group handling for proxy mode
Proxy mode does not support security groups, but the handlers don't account for that. The code should be updated to add proxy mode as another condition similar to checking the OS. Proxy mode should be handled just like "!= linux".

russellb:
Security groups are not supported in the userspace proxy mode. I saw
some errors when running `nexd proxy` in a Kubernetes Pod. It was
attempting to run some `nft` commands.

Closes #1180


nerdalert (approved):
LGTM

PR #1182, opened 2023-06-13
hack: Add script to get to the apiserver db

russellb:
Add a helper script that connects a postgresql client to the apiserver
database in the development environment.


russellb:
No need!  There is already `make run-sql-apiserver`

Issue #1117, opened 2023-06-04
Use a db transaction when creating a user and its organization


russellb:
While reviewing #973, I noticed that a db transaction isn't used when creating a user and its corresponding organization. This can cause some parallel attempts to create resources. You can even see a spot in the code where it's trying to do its own error handling for a concurrent operation.

```
	// maybe another concurrent request created it...
	org = models.Organization{}
	if api.db.Where("owner_id = ?", userId).First(&org).Error == nil {
		return org.ID, nil
	}
```

If we create the user and its organization atomically, we won't have to handle this condition.

Jahmilli:
Hey @russellb i'm happy to check this out :)

russellb:
> Hey @russellb i'm happy to check this out :)

Sounds good!

PR #1183, opened 2023-06-13
apiserver: Move creating a user/organization in a transaction block

Jahmilli:
Have moved the creation of the user and organization into a transaction block to support rollbacks and other benefits of transaction handling. 

Closes https://github.com/nexodus-io/nexodus/issues/1117

### DIFF
--- a/internal/handlers/organization.go
+++ b/internal/handlers/organization.go
@@ -135,14 +135,14 @@ func (api *API) CreateOrganization(c *gin.Context) {
 	}
 
 	// Create a default security group for the organization
-	sg, err := api.createDefaultSecurityGroup(ctx, org.ID.String())
+	sg, err := api.createDefaultSecurityGroup(ctx, org.ID.String(), nil)
 	if err != nil {
 		c.JSON(http.StatusInternalServerError, models.NewApiInternalError(err))
 		return
 	}
 
 	// Update the org with the new security group id, delete the security group if it fails
-	if err := api.updateOrganizationSecGroupId(ctx, sg.ID, org.ID); err != nil {
+	if err := api.updateOrganizationSecGroupId(ctx, sg.ID, org.ID, nil); err != nil {
 		if res := api.db.Delete(&sg, "id = ?", sg.ID); res.Error != nil {
 			errMsg := fmt.Sprintf("failed to update and delete the default security group with an org id: %v", err)
 			c.JSON(http.StatusInternalServerError, models.NewApiInternalError(errors.New(errMsg)))
--- a/internal/handlers/organization.go
+++ b/internal/handlers/organization.go
@@ -135,14 +135,14 @@ func (api *API) CreateOrganization(c *gin.Context) {
 	}
 
 	// Create a default security group for the organization
-	sg, err := api.createDefaultSecurityGroup(ctx, org.ID.String(), nil)
+	sg, err := api.createDefaultSecurityGroup(ctx, nil, org.ID.String())
 	if err != nil {
 		c.JSON(http.StatusInternalServerError, models.NewApiInternalError(err))
 		return
 	}
 
 	// Update the org with the new security group id, delete the security group if it fails
-	if err := api.updateOrganizationSecGroupId(ctx, sg.ID, org.ID, nil); err != nil {
+	if err := api.updateOrganizationSecGroupId(ctx, nil, sg.ID, org.ID); err != nil {
 		if res := api.db.Delete(&sg, "id = ?", sg.ID); res.Error != nil {
 			errMsg := fmt.Sprintf("failed to update and delete the default security group with an org id: %v", err)
 			c.JSON(http.StatusInternalServerError, models.NewApiInternalError(errors.New(errMsg)))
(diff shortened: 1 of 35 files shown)


Jahmilli:
Will need to have a look at how we can test the error scenarios for this, not sure if we're doing any sort of mocking or anything...

Also I've updated the `createDefaultSecurityGroup` and `updateOrganizationSecGroupId`. I can see that in `internal/handlers/organization.go` we use these and perform some manual rollback as there is no transaction used. Will look at updating this as well.

internal/handlers/user.go
```diff
@@ -39,63 +42,93 @@ func (api *API) UserIsCurrentUser(c *gin.Context) func(db *gorm.DB) *gorm.DB {
...
-	user.ID = id
-	user.UserName = userName
-	res = tx.Create(&user)
-	if res.Error == nil {
-		return api.createUserOrgIfNotExists(ctx, id, userName)
+
+	return noUUID, fmt.Errorf("can't create user record")
+}
+
+// restoreDeleteUser will restore a user if they have been deleted
```
Jahmilli, line 99: Will fix this up in next commit that will incorporate feedback on other commit as well as general PR feedback

internal/handlers/user.go
```diff
@@ -39,63 +42,93 @@ func (api *API) UserIsCurrentUser(c *gin.Context) func(db *gorm.DB) *gorm.DB {
...
-
-		// Was the user was previously deleted... lets make him active again.
-		if user.DeletedAt.Valid {
-			user.DeletedAt = gorm.DeletedAt{}
-			res = tx.Unscoped().Model(&user).Update("DeletedAt", user.DeletedAt)
-			if res.Error != nil {
-				return noUUID, res.Error
+	var uuid uuid.UUID
+
+	err := api.db.WithContext(ctx).Transaction(func(tx *gorm.DB) error {
```
chirino, line 55: use `api.transaction` instead of `api.db.WithContext(ctx).Transaction` it does the same thing an works better with cockroachdb in case we ever decide to support it.

    err := api.transaction(ctx, func(tx *gorm.DB) error {

internal/handlers/user.go
```diff
@@ -118,43 +151,55 @@ func (api *API) createUserOrgIfNotExists(ctx context.Context, userId string, use
 			ID: userId,
 		}},
 	}
-	if res = api.db.Create(&org); res.Error == nil {
 
+	if db == nil {
+		db = api.db.WithContext(ctx)
+	}
+
+	err := db.Transaction(func(tx *gorm.DB) error {
```
chirino, line 159: use api.transaction

chirino (changes requested):

chirino:
So one thing that worries me is that `createUserIfNotExists()` get call for EVERY api call.  Making things transitional might slow things down as we MAY be holding additional DB locks.  So maybe we should benchmark how this change looks under load.  Or maybe we add some caching to avoid the slow path when the user exists?

Jahmilli:
> So one thing that worries me is that `createUserIfNotExists()` gets called for EVERY api call. Making things transactional might slow things down as we MAY be holding additional DB locks. So maybe we should benchmark how this change looks under load. Or maybe we add some caching to avoid the slow path when the user exists?

Hey thanks for the callout. IMO caching would be ideal but i'm happy to take suggestions on how to move forwards on this from you guys. If you'd like I could look into storing basic info on the user in redis, i.e the user-id as the key along with the username. Could then do some basic validation against that to quickly determine whether or not we need to perform a transaction along with a list of queries.
Otherwise ill have a look at the docs on load testing and explore that in the meantime.
Just let me know the best way to proceed on this and ill find some time to check it out tomorrow.

chirino:
So thinking about it, given we don't really have any automated perf regression tests, and we haven't really tried to optimize for perf/scale yet, we should likely keep a mental note (or create an issue to remind us) that this is a possible optimization point.  Note that it's duration is being traced, so that should also help us visualize how much it's impacting perf when we get to that.

russellb (approved):
This looks OK to me. I noticed another area that needs the same type of fix, but it can be tracked as a separate issue.

internal/handlers/organization.go
```diff
@@ -135,14 +135,14 @@ func (api *API) CreateOrganization(c *gin.Context) {
...
-	sg, err := api.createDefaultSecurityGroup(ctx, org.ID.String())
+	sg, err := api.createDefaultSecurityGroup(ctx, nil, org.ID.String())
 	if err != nil {
 		c.JSON(http.StatusInternalServerError, models.NewApiInternalError(err))
 		return
 	}
 
 	// Update the org with the new security group id, delete the security group if it fails
-	if err := api.updateOrganizationSecGroupId(ctx, sg.ID, org.ID); err != nil {
+	if err := api.updateOrganizationSecGroupId(ctx, nil, sg.ID, org.ID); err != nil {
```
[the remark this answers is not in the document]
Jahmilli: Hey yeah, i noticed that too, i thought it would be best to do as a separate PR. I can check that out as well :)

russellb:
@chirino I took a look at this, but it won't auto-merge until you review it as well since you reviewed it earlier. Just FYI in case you didn't know that bot behavior.

PR #1184, opened 2023-06-14
Add peer health tracking to the device cache

russellb:
- nexd: s/Reconcile()/reconcileDeviceCache()/
- nexd: Add peer health tracking to the device cache

commit 6b15a558ce212f092c41667a0d03ebb4e3262cc6
Date:   Tue Jun 13 10:24:03 2023 -0400

    nexd: s/Reconcile()/reconcileDeviceCache()/
    
    There are now several reconcile methods. Rename this one to reflect
    what it does.
    
commit f5c350f59d7cd8bc9a8920e499de1424595f0ae3
Date:   Tue Jun 13 16:49:20 2023 -0400

    nexd: Add peer health tracking to the device cache
    
    Update `nexd` to keep track of peer connection health. The only place
    this info is used at the moment is in the output of `nexctl nexd peers
    list`. In the future, this health information can be used to help
    trigger changes to peering configuration, as we will want to try other
    peering options if the current configuration isn't working.
    
    Determing peer connection health depends on the peer information dump
    that we get from the wireguard stack. We now pull that data during
    every device cache reconciliation so we can refresh our view of the
    connection health.
    
    There are several factors that go into calculating our view of the
    peer connection health. Those factors include:
    
    - the last time a wireguard handshake occurred
    - the last time we saw tx/rx counters update
    - our keepalive interval
    - the wireguard keepalive timeout
    - the time we last updated our wireguard peer configuration
    
    The method peerIsHealthy() has several comments and debug log messages
    that explain the logic.
    
    One unfortunate piece of this change is that the device cache is now
    accessed by multiple threads, which adds a bit of complexity to the
    implementation. The unix socket control interface used by `nexctl` now
    runs a handler that accesses the device cache to get the latest health
    info. As a result, we now use a read/write mutex around reading or
    updating the device cache map.
    

russellb:
There should be no functional impact here aside from:
- Additional debug log messages related to peer connection health
- A new column in the output of `nexctl nexd peers list`

The larger benefit is that if the data seems reliable after observing this info for a while, we can use it to influence peer configuration changes.

nerdalert (approved):
LGTM

PR #1193, opened 2023-06-15
nexctl: Trim down peers list output

### LINKED ISSUE #1112: UX work to add/remove the most relevant columns for nexctl output
nexctl output is getting pretty wide in the column view. We should pair some of the columns down to what is most useful for routine debugging and relevant data most beneficial to the user. Maybe add a `--wide` option as well.

russellb:
Update the output of `nexctl nexd peers list` with a minimal list of
columns by default. To get full output, you can use the new `--full`
option. This provides output that won't line wrap in most cases.

Closes #1112


nerdalert (approved):
LGTM

Issue #1187, opened 2023-06-14
Rename any instances of the func (ax *Nexodus) to func (nx *Nexodus)


nerdalert:
Following the rename of the project a while back, we did not rename the Nexodus receiver from `ax` to `nx` everywhere.

This is a great first issue that only involves renaming the receiver everywhere in the `internal/nexodus/` path. Ping @nerdalert if you have any questions on building/testing, ty!

russellb:
I agree with the change. I suggest submitting PRs to do this in smaller chunks. One PR may be painful to rebase or for others to rebase on top of. It would also make review easier.

```
$ git grep ax internal/ | wc -l
588
```

cyneptic:
Hey i'd like to this.
I'll send PRs file by file so that its done in smaller chunks.

russellb:
I believe this is done

PR #1194, opened 2023-06-15
renames ax to nx in the receiver functions of internal/nexodus/keys*.go files

cyneptic:
resolves part of issue #1187


nerdalert (approved):
LGTM

PR #1196, opened 2023-06-15
renames ax to nx on receiver functions of internal/nexodus/ctl*.go files

cyneptic:
resolves part of issue #1187, had to rename a struct's field aswell.


russellb:
Closed per discussion with @cyneptic

PR #1199, opened 2023-06-17
Add design doc for the network router feature

nerdalert:
- More details and running code in #981


nerdalert:
The diagrams are via https://excalidraw.com. If exported as described below you can simply drag the PNG into excalidraw to edit it. The embedding is simply JSON. Worth a consideration since its much easier to draw detailed docs than other tools and its shareable via the process below.

```
export the Excalidraw drawing as a PNG and select the Embed Scene option. Excalidraw will embed the JSON inside the PNG file and the PNG file can be edited by simply dragging it onto the Excalidraw website. The PNG is completely compatible with browsers and other applications though.
```

docs/development/design/network-router.md
```diff
@@ -0,0 +1,93 @@
...
+## Summary
+
+Add the ability for device's running nexd to access non-nexd devices and act as a router for external networks.
+
+## Proposal
+
+There are two supported use cases.
+
+1. Traffic from the Nexd mesh can reach external non-nexd nodes via source NAT without any configuration required.
+2. Traffic from the Nexd mesh can reach external non-nexd nodes but require a configuration for return traffic since SNAT is disabled. This scenario will also allow external devices to initiate traffic to the mesh since the devices are required to have a route to an interface on the device connected to the physical network. With SNAT being disabled, the non-nexd device can also initiate traffic from outside the mesh going in.
```
chirino, line 14: Might be better to make #2 two different use cases even if they get implemented using the same strategy.

docs/development/design/network-router.md
```diff
@@ -0,0 +1,93 @@
...
+
+### Network Router with SNAT
+
+This is a zero touch provisioning for the non-nexodus device since the edge nexd device will be source natting the traffic.
+
+```shell
+nexd --username admin --password floofykittens router --child-prefix 10.180.0.0/20 --network-router  https://try.nexodus.127.0.0.1.nip.io
+```
```
chirino, line 22: Does this command related to the following example diagram?

docs/development/design/network-router.md
```diff
@@ -0,0 +1,93 @@
...
+
+Here is an example with text diagramming a packet flow.
+
+![no-alt-text](diagrams/network-router-snat.png)
+
```
chirino, line 27: Maybe we should show that route table entries for 192.168.0.0 network are created by nexd in Node2

docs/development/design/network-router.md
nerdalert: You suggesting a separate doc or just putting it in a different section for NAT being disabled? Ty.

docs/development/design/network-router.md
nerdalert: Thanks, updated the prefix to match.

docs/development/design/network-router.md
chirino: Just to add a third bullet to this doc.

docs/development/design/network-router.md
```diff
@@ -0,0 +1,114 @@
...
+## Summary
+
+Add the ability for device's running nexd to access non-nexd devices and act as a router for external networks.
+
+## Proposal
+
+There are two supported use cases.
+
+1. [Network Router with SNAT](#network-router-with-snat) Traffic from the Nexd mesh can reach external non-nexd nodes via source NAT without any configuration required.
+2. [Network Router with SNAT disabled](#network-router-with-snat-disabled) Traffic from the Nexd mesh can reach external non-nexd nodes but require a route to be added for return traffic since SNAT is disabled.
```
russellb, line 14:
```suggestion
1. [Network Router with SNAT](#network-router-with-snat): Traffic from the Nexd mesh can reach external non-nexd nodes via source NAT without any configuration required.
2. [Network Router with SNAT disabled](#network-router-with-snat-disabled): Traffic from the Nexd mesh can reach external non-nexd nodes but external nodes require a route to be added for return traffic since SNAT is disabled.
```

docs/development/design/network-router.md
```diff
@@ -0,0 +1,114 @@
...
+3. With SNAT disabled, a user can initiate traffic from external, non-nexd devices to initiate traffic in to the nexodus mesh.
```
russellb, line 15: The section begins with saying there are 2 supported use cases, so I wouldn't make this a number 3 on the list. You could add it as another sentence with number 2, or just as a clarifying sentence afterwards like this.

```suggestion

With SNAT disabled, a user can initiate traffic from external, non-nexd devices to initiate traffic in to the nexodus mesh.
```

docs/development/design/network-router.md
```diff
@@ -0,0 +1,114 @@
...
+
+### Network Router with SNAT
+
+This is a zero touch provisioning for the non-nexodus device since the edge nexd device will be source natting the traffic.
+
+```shell
+nexd --username admin --password floofykittens router --child-prefix 192.168.100.0/24 --network-router  https://try.nexodus.127.0.0.1.nip.io
```
russellb, line 22:
```suggestion
nexd --username admin --password floofykittens --service-url https://try.nexodus.127.0.0.1.nip.io \
    router --child-prefix 192.168.100.0/24 --network-router  
```

docs/development/design/network-router.md
```diff
@@ -0,0 +1,114 @@
...
+    }
+}
+```
+
+### Network Router with SNAT disabled
+
+There are additional configurations required to make this scenario work:
+
+- The non-nexd node needs to have a route, either static or from the underlying IGP pointing to the ingress interface on the nexd edge node.
+- It is not uncommon for there to be source/destination traffic checking enabled and blocked by default on cloud provider instances. For example, on EC2 one would need to disable [source / destination check](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-eni.html#modify-source-dest-check). The rest of the IP forwarding enablement is managed by the nexd agent.
```
russellb, line 64:
```suggestion
- It is common for there to be source/destination traffic checking enabled and blocked by default on cloud provider instances. For example, on EC2, one would need to disable [source / destination check](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-eni.html#modify-source-dest-check). The rest of the IP forwarding enablement is managed by the nexd agent.
```

russellb (changes requested):
I put some trivial edits so far.

Honestly, I felt confused after reading this because now I'm not sure I understand what `--child-prefix` is for by itself. What does it do specifying that option by itself?

I would prefer to have one design doc that covers all options to `nexd router` to keep all of this straight.

russellb:
> Honestly, I felt confused after reading this because now I'm not sure I understand what `--child-prefix` is for by itself. What does it do specifying that option by itself?

We had a live discussion about this and I understand better now. `--child-prefix` by itself does not set up any routing for the prefix. It works for either:

1. An IP range that's local to the `nexd` device, like containers running on that host.
2. An IP range that a user has set up routing for manually.

russellb:
TODO
- Add a future enhancement idea of a flag that allows routing traffic into a Nexodus network with SNAT but without specifying any prefixes to advertise

russellb (approved):

chirino (approved):
lgtm

Issue #1131, opened 2023-06-05
Revisit DB transaction handling in CreateOrganization


russellb:
While reviewing #973, I noted that creating the default security group for the organization was done outside of the main DB transaction inside of the CreateOrganization() method. Instead of trying to handle manual rollbacks when errors occur, it makes sense to see if this could all be done in a single transaction.

As the code stands now, the result of creating an organization is not guaranteed to be consistent. For example:

- If creating the default security group fails, it will return an error, though the organization will remain in the DB, just without its default security group.
- If updating the organization with the ID of the new security group fails, the API will return an error, but the organization is still created, but without its default security group
- There is a race condition where it's technically possible for the organization to get created and then changed out from under the remaining code that is trying to update it with a default security group.

Jahmilli:
Can pick this up

PR #1208, opened 2023-06-21
Create organization sg transaction

Jahmilli:
Moving create default security group logic into transaction and removing the manual error handling that was previously done.

Closes https://github.com/nexodus-io/nexodus/issues/1131


russellb (approved):
very nice!

PR #1213, opened 2023-06-22
Add *.png to the path-ignore for the build action

nerdalert:
- Prevent build actions from being triggered from a png extended file

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,6 +5,7 @@ on:
     branches: ["main"]
     paths-ignore:
       - '**/*.md'
+      - '**/*.png'
       - '**/*.gitignore'
       - '.vscode/**'
       - '.github/mergify.yml'
@@ -13,6 +14,7 @@ on:
     # Must stay in sync with the paths in .github/workflows/docs.yml and .github/mergify.yml
     paths-ignore:
       - '**/*.md'
+      - '**/*.png'
       - '**/*.gitignore'
       - '.vscode/**'
       - '.github/mergify.yml'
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -13,6 +14,7 @@ on:
     # Must stay in sync with the paths in .github/workflows/docs.yml and .github/mergify.yml
     paths-ignore:
       - '**/*.md'
+      - '**/*.png'
```
russellb, line 17: Can you update `docs.yml` and `mergify.yml` as suggested by the comment a few lines above?

russellb (changes requested):

.github/workflows/build.yml
nerdalert: @russellb correct me if Im wrong here, but those look like they are for triggering markdown-lint in doc.yml or checking for a md-lint pass in mergify?

russellb: Yes, that's the purpose of the files.

Here's the problem with the change as-is. Imagine a PR that only changes a png file.
- The build workflow does not run as a result of this change.
- Mergify will never merge the PR because the rules based on file types says the build workflow must pass.

An alternative is to make sure that the docs workflow always runs for png file changes. That requires changing docs.yml and then changing mergify.yml to have it look for the docs job passing.

I'll push a patch for you.

russellb: @nerdalert take a look at the commit I added to the branch. I wanted you to see what I meant before it merges. After you take a look, you can drop the `hold` label so this will merge.

russellb (approved):

russellb:
@mergifyio rebase

.github/workflows/build.yml
russellb: I dropped the hold, just take a look whenever :)

PR #1220, opened 2023-06-25
nexd: Optionally build with pprof support

russellb:
Go has tooling that makes profiling pretty easy. Add some build
automation and instructions for using it with nexd.


chirino (approved):
LGTM.
Maybe we should also do this for the apiserver?

russellb:
> LGTM. Maybe we should also do this for the apiserver?

Yes, good idea, done here: https://github.com/nexodus-io/nexodus/pull/1243

PR #1221, opened 2023-06-25
Implements automatically setting up nexd proxy ingress/egress rules using skupper config.

chirino:
To test out the PoC:
1. get local dev server running using `make run-on-kind`.
2. run `./hack/skupper/install-skupper.sh` to install skupper and the nexd routers.  They will be installed to both a `east` and `west` namespace.
3. run `./hack/skupper/proxy-status.sh` to see the ingress/egress rules that get created on each nexd running in proxy mode.


hack/skupper/install-skupper.sh
```diff
@@ -0,0 +1,85 @@
...
+
+# For some reason the LB ip:port did not work on my machine, so setup a port forward to
+# access the front end app:
+kubectl port-forward $(kubectl get pods -l app=frontend -o name) 8080&
+sleep 2
+open http://localhost:8080
+
+#
+# Deploy nexodus proxy routers that will proxy services based on the skupper config:
+kubectl config set-context --current --namespace west
```
russellb, line 81: One thing that bit me was that if you run two separate scripts at the same time that are nothing using `kubectl config set-context`, they'll break each other. I did that with some performance testing I was doing.

russellb:
@chirino very cool PoC -- is it posted here for easy discussion and testing?  or were you proposing that it be merged? Maybe we should create a new repo for skipper integration. We could run the config controller as a sidecar container in the nexd container that uses either nexctl or the socket interface directly for managing rules.

chirino:
Yeah. Right now it's here for easy sharing.

We can move this around to a different process if we figure out the final desires architecture.

Issue #1216, opened 2023-06-23
Brief bug summary


chirino:
### Describe the bug

This error shows up in the apiserver logs when concurrent requests for the `/api/users/me` occur for the first time from the same user.

    {"level":"error","ts":1687482540.8690588,"caller":"zap@v0.1.0/zap.go:87","msg":"can't create user record: duplicated key not allowed","status":500,"method":"GET","path":"/api/users/me" ....

### How to Reproduce

Repeatedly run the `TestConcurrentApiAccess` e2e test.

### Expected Behavior

No error should occur.  The `/api/users/me` call should be idempotent.

chirino:
We may want to change the api.transaction(...) so that if it fails the first time, we try again a second time in hopes that it was an optimistic concurrency error and the 2nd run will gracefully handle the new DB state.

PR #1222, opened 2023-06-26
apiserver: Added in retry handling for transaction for duplicate key …

Jahmilli:
Added in retriable error handling for the create user/org transaction. This is to support error case when concurrent creation requests are done.

Closes https://github.com/nexodus-io/nexodus/issues/1216

### DIFF
--- a/internal/handlers/user.go
+++ b/internal/handlers/user.go
@@ -5,10 +5,12 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"time"
 
 	"github.com/gin-gonic/gin"
 	"github.com/google/uuid"
 	"github.com/nexodus-io/nexodus/internal/models"
+	"github.com/nexodus-io/nexodus/internal/util"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
 	"gorm.io/gorm"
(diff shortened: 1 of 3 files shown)


Jahmilli:
Hey @chirino  let me know what you think of the approach but it's based off the suggestion you mentioned in https://github.com/nexodus-io/nexodus/issues/1216

internal/handlers/user.go
```diff
@@ -53,50 +55,53 @@ func (api *API) createUserIfNotExists(ctx context.Context, id string, userName s
...
-		res := tx.Unscoped().First(&user, "id = ?", id)
-
-		// If the user exists, then lets restore their status in the database
-		if res.Error == nil {
-			if user.DeletedAt.Valid {
-				user.DeletedAt = gorm.DeletedAt{}
-				if res := tx.Unscoped().Model(&user).Update("DeletedAt", user.DeletedAt); res.Error != nil {
-					return res.Error
+		// Retry the operation if we get a duplicate key error which can occur on concurrent requests when creating a user
+		return util.RetryOperationForErrors(ctx, time.Millisecond*10, 1, []error{gorm.ErrDuplicatedKey}, func() error {
```
chirino, line 59: Perhaps we should do this outside the transaction block? I'm slightly worried that you might commit a change that occurred in an operation that had failed.

chirino (changes requested):
other than that one little nit, it LGTM.

internal/handlers/user.go
Jahmilli: Good catch, definitely right, will fix this up!

chirino (approved):
LGTM.

Jahmilli:
Thanks for reviewing @chirino, I don't have write perms so i can't add/remove labels or merge, if that could be added then i think we'll be able to look at closing it! 🕺

russellb:
> I don't have write perms so i can't add/remove labels or merge

I just invited you to join the Nexodus github organization. Check your email. You'll be added to the @nexodus-io/nexodus-triage-access team which should let you manage labels.

Issue #1191, opened 2023-06-15
Cache user to improve performance of createUserIfNotExists


Jahmilli:
### Describe the Problem Statement

In `internal/handlers/user.go` the method `createUserIfNotExists` is called on every API request which continues to do a lookup on the user in the database, and create the user and an organization if it doesn't exist. Doing a db lookup is considerably more expensive than a lookup in a cache and so there's potential for a performance improvement here. I currently don't have metrics though to determine whether or not this is a significant issue atm but it would be important as the project scales. 

### Describe the Enhancement

Ideally we should be caching the user (Possibly with Redis??) so that after the user is created, follow up calls can do a fast lookup in Redis instead of having to do a db query, AND if the user does not exist, then we continue to perform the create user logic.

### Additional context

Based on the suggestion from @chirino from https://github.com/nexodus-io/nexodus/pull/1183

PR #1236, opened 2023-06-27
Cache user to improve performance of createUserIfNotExists

akoserwal:
Attempt to fix: https://github.com/nexodus-io/nexodus/issues/1191

Trying to reuse some redis session impl from: https://github.dev/nexodus-io/nexodus/pull/984

### DIFF
--- a/internal/handlers/user.go
+++ b/internal/handlers/user.go
@@ -24,10 +24,18 @@ func (api *API) CreateUserIfNotExists() gin.HandlerFunc {
 	return func(c *gin.Context) {
 		id := c.GetString(gin.AuthUserKey)
 		username := c.GetString(AuthUserName)
-		_, err := api.createUserIfNotExists(c.Request.Context(), id, username)
+		session, err := api.sessionManager.Start(c.Request.Context(), c.Writer, c.Request)
 		if err != nil {
-			_ = c.AbortWithError(http.StatusInternalServerError, err)
-			return
+			api.logger.Errorf("failed to start the redis session manager:%s", err)
+		}
+		_, ok := session.Get(id)
+		if !ok {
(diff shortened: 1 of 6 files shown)


russellb:
@akoserwal take a look at the unit tests. It looks like they're failing

akoserwal:
> @akoserwal take a look at the unit tests. It looks like they're failing

I will work on fixing the test case, I just to know, if the approach is fine.

akoserwal:
Few test cases are failing due to issues with session manager
https://github.com/go-session/redis/issues/5

internal/handlers/user.go
```diff
@@ -24,10 +24,18 @@ func (api *API) CreateUserIfNotExists() gin.HandlerFunc {
 	return func(c *gin.Context) {
 		id := c.GetString(gin.AuthUserKey)
 		username := c.GetString(AuthUserName)
-		_, err := api.createUserIfNotExists(c.Request.Context(), id, username)
+		session, err := api.sessionManager.Start(c.Request.Context(), c.Writer, c.Request)
```
chirino, line 27: This would only work for browser based sessions.  sessions manager creates a random key for each user and stores it in a browser cookie.  This key is what is used to get the session data out of redis.  

We need this cache site to work even for non-browser sessions.  So you will need to use redis directly as a map and not as a user session.

chirino (changes requested):

internal/handlers/user.go
akoserwal: @chirino, thanks for the clarification; I can update it to work for non-browser sessions.

chirino: this article might help: https://blog.petehouston.com/use-redis-to-cache-data-in-go/

akoserwal:
Create a new PR: https://github.com/nexodus-io/nexodus/pull/1245

Issue #1143, opened 2023-06-07
Failed peer configuration does not result in a retry to configure that peer


russellb:
I encountered a situation where a new peer (a linux host) in my nexodus organization was unable to ping another peer (a mac). Further debugging showed that the route for the new peer was not present. Some of the possible errors in creating this route are debug messages and nexd was not running in debug at this time.

`nexd` currently does not propagate any errors that occur when adding routes or adding wireguard configuration for a new or updated peer. If it fails, these configuration items will not be attempted again until the peer's configuration changes for some reason.

See this snippet from `DeployWireguardConfig()`

```
		ax.handlePeerRoute(peer)
		ax.handlePeerTunnel(peer)
```

Jahmilli:
Im happy to look into this if possible! Would be good to get more across `nexd`

PR #1237, opened 2023-06-28
nexd: add in error propogation for deploying wireguard config

Jahmilli:
Propagating errors when setting up Wireguard config. 
Checking route exists for Darwin prior to deleting a route.
Have also updated method variable names from `ax` to `nx` as I believe this is something that is desired and is a minor thing to do as part of this.

Closes https://github.com/nexodus-io/nexodus/issues/1143

### DIFF
--- a/internal/nexodus/nexodus.go
+++ b/internal/nexodus/nexodus.go
@@ -815,8 +815,8 @@ func (nx *Nexodus) addToDeviceCache(p public.ModelsDevice) {
 	}
 }
 
-func (ax *Nexodus) reconcileDeviceCache() error {
-	peerMap, resp, err := ax.informer.Execute()
+func (nx *Nexodus) reconcileDeviceCache() error {
+	peerMap, resp, err := nx.informer.Execute()
 	if err != nil {
 		if resp != nil {
 			return fmt.Errorf("error: %w header: %v", err, resp.Header)
@@ -825,34 +825,34 @@ func (ax *Nexodus) reconcileDeviceCache() error {
 	}
 
 	// Get the current peer configuration data from the wireguard interface
-	peerStats, err := ax.DumpPeersDefault()
+	peerStats, err := nx.DumpPeersDefault()
 	if err != nil {
-		if ax.TunnelIP != "" {
+		if nx.TunnelIP != "" {
 			// Unexpected to fail once we have local interface configuration
-			ax.logger.Warnf("failed to get current peer stats: %w", err)
+			nx.logger.Warnf("failed to get current peer stats: %w", err)
 		}
 		peerStats = make(map[string]WgSessions)
 	}
 
-	ax.deviceCacheLock.Lock()
-	defer ax.deviceCacheLock.Unlock()
+	nx.deviceCacheLock.Lock()
+	defer nx.deviceCacheLock.Unlock()
 
 	// Get our device cache up to date
 	newLocalConfig := false
 	for _, p := range peerMap {
 		// Update the cache if the device is new or has changed
-		existing, ok := ax.deviceCache[p.PublicKey]
-		if !ok || !ax.isEqualIgnoreSecurityGroup(existing.device, p) {
-			if p.PublicKey == ax.wireguardPubKey {
+		existing, ok := nx.deviceCache[p.PublicKey]
+		if !ok || !nx.isEqualIgnoreSecurityGroup(existing.device, p) {
+			if p.PublicKey == nx.wireguardPubKey {
 				newLocalConfig = true
 			}
-			ax.addToDeviceCache(p)
-			existing = ax.deviceCache[p.PublicKey]
+			nx.addToDeviceCache(p)
+			existing = nx.deviceCache[p.PublicKey]
 		}
 
 		// Store the relay IP for easy reference later
 		if p.Relay {
-			ax.relayWgIP = p.AllowedIps[0]
+			nx.relayWgIP = p.AllowedIps[0]
 		}
 
 		// Keep track of peer connection stats for connection health tracking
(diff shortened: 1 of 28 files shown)


internal/nexodus/nexodus.go
```diff
@@ -873,42 +873,42 @@ func (ax *Nexodus) reconcileDeviceCache() error {
...
-		if err := ax.DeployWireguardConfig(updatePeers); err != nil {
+		if err := nx.DeployWireguardConfig(updatePeers); err != nil {
 			if strings.Contains(err.Error(), securityGroupErr.Error()) {
 				return err
 			}
 			// If the wireguard configuration fails, we should wipe out our peer list
 			// so it is rebuilt and reconfigured from scratch the next time around.
-			ax.wgConfig.Peers = nil
+			nx.wgConfig.Peers = nil
 			return err
```
Jahmilli, line 905: @russellb @nerdalert Would be good to get your thoughts around this, now that we have a few more places within `DeployWireguardConfig` that could fail, should we still continue to do this or should we be looking at what updatedPeers failed to be configured and only remove them from the peer list?

nerdalert: @Jahmilli I think that definitely sounds more ideal. If you remove that peer from the device cache it will force an attempt to re-add it on the next peer listing.  That said, I'm not sure that would get triggered with the peer listing watch. @chirino could answer that.

If it failed on the security group, you can remove the security group ID from the peer in device cache and it would get triggered since security groups are on a poller rather than a watch currently. Sorry its not a super clean answer here. If it feels overly complex we can pop a new issue and tackle it in another PR. Thanks!!

russellb: I would suggest leaving it as-is and considering further changes in later changes. The current behavior is the safest.  "We failed to apply the config; start over and try again."

Whether it is this change or a peer-specific change, none of that will trigger a retry. The retry will happen either when a device change comes in from watching the API, or when the reconcile timer is hit. There's still a timer that runs this code every 20 seconds, just in case.

internal/nexodus/nexodus.go
```diff
@@ -815,8 +815,8 @@ func (nx *Nexodus) addToDeviceCache(p public.ModelsDevice) {
 	}
 }
 
-func (ax *Nexodus) reconcileDeviceCache() error {
-	peerMap, resp, err := ax.informer.Execute()
+func (nx *Nexodus) reconcileDeviceCache() error {
```
russellb, line 818: Don't worry about it for this PR, but my preference as a best practice would be to split out all of the renaming into its own commit. That makes the review of the logic changes easier, as the reviewer can look at each commit individually.

internal/nexodus/utils.go
```diff
@@ -46,6 +48,31 @@ func RunCommand(cmd ...string) (string, error) {
 	return string(output), nil
 }
 
+// RunCommandPipeline runs a pipeline of commands in a shell
+func RunCommandPipeline(commands ...string) (string, error) {
```
russellb, line 52: I would prefer a different implementation for the pipeline that doesn't use `sh`.  See "Piping Between Processes" on this blog post for an example: https://www.dolthub.com/blog/2022-11-28-go-os-exec-patterns/

My reason is that it's less likely to introduce a security issue later. By using `sh` here, if anyone accidentally passes user-supplied input here that can be interpreted by `sh`, we have a shell-injection vulnerability. As a made-up example, imagine some new code uses an organization name as part of a command and also forgets to quote it properly. What happens if someone renames their organization `; rm -rf / ;` ?  😬 

I don't see any such chance with the code as-is, but changing the implementation would avoid the possibility in the future.

russellb (changes requested):

internal/nexodus/nexodus.go
Jahmilli: Hey will keep that in mind in the future, definitely realised it was a mistake when i saw how many changes were made as i went through this

internal/nexodus/utils.go
Jahmilli: Good callout, will fix it up!

Jahmilli: @russellb  have updated based on feedback. After looking through the PR i realise the scope has sort of creeped a bit more than the initial issue (changing variable naming, doing validation on routes for Darwin etc) which is a bit of a mistake... Happy to split these changes up into separate PRs if you'd be more comfortable with it, just let me know.

russellb: Thanks for asking, but don't worry about it on this one. I'd rather that effort go toward something else. :)

internal/nexodus/utils_darwin.go
```diff
@@ -3,15 +3,49 @@
...
+	defer r.Close()
+	ns := exec.Command("netstat", "-r", "-n")
+	ns.Stdout = w
+	if err = ns.Start(); err != nil {
+		return false, err
+	}
+	defer func() {
+		_ = ns.Wait()
+	}()
+	w.Close()
```
russellb, line 36: If ns.Start() fails, the write end of the pipe is never closed. I see the blog post I sent you has the same thing. It looks like a bug to me. What do you think? I would move `w.Close()` to be `defer w.Close()` right after `defer r.Close()`.

internal/nexodus/wg_deploy.go
```diff
@@ -33,13 +33,17 @@ func (ax *Nexodus) DeployWireguardConfig(updatedPeers map[string]public.ModelsDe
...
 			continue
 		}
-		ax.handlePeerRoute(peer)
-		ax.handlePeerTunnel(peer)
+		if err := nx.handlePeerRoute(peer); err != nil {
+			return err
+		}
+		if err := nx.handlePeerTunnel(peer); err != nil {
+			return err
+		}
```
russellb, line 44: I think a more optimistic way to handle this is to finish looping through all `updatedPeers` even if we encounter a failure with one of them. Configuring other peers may still work. You could just save off the first error you hit to return after the loop.

russellb (changes requested):

internal/nexodus/utils_darwin.go
```diff
@@ -3,15 +3,49 @@
...
+
+	// #nosec -- G204: Subprocess launched with a potential tainted input or cmd arguments (gosec)
```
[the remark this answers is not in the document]
Jahmilli: Definitely fair, i thought because im doing the validation above then it's alright to keep this otherwise i would have looked for a different approach but not too opinionated on this.

internal/nexodus/wg_deploy.go
```diff
@@ -14,32 +14,37 @@ var (
...
+		if err = nx.handlePeerTunnel(peer); err != nil {
+			nx.logger.Errorf("Failed to handle peer tunnel: %v", err)
+		}
 	}
 
-	ax.logger.Debug("Peer setup complete")
-	return nil
+	nx.logger.Debug("Peer setup complete")
+	return err
```
russellb, line 49: If there are two updated peers and the first fails, but the second succeeds, we won't return an error here. Maybe create a `firstError` variable that stores `err` the first time an error occurs and then return that here at the end.

I think we should return an error if any peer update fails and not just the last one.

russellb (changes requested):

internal/nexodus/wg_deploy.go
Jahmilli: Ahh this was a dumb mistake 😵  You're right though, have fixed in latest commit

russellb: not dumb. The error handling I'm suggesting is a little funky

russellb (approved):

nerdalert (approved):
LGTM ty!

PR #1243, opened 2023-07-03
Add pprof support to the apiserver image

russellb:
Also document how to get a profile from a running apiserver.


nerdalert (approved):
LGTM

PR #1245, opened 2023-07-04
Cache user to improve performance of createUserIfNotExists

akoserwal:
Using redis to cache the user

closes: https://github.com/nexodus-io/nexodus/issues/1191
closes: https://github.com/nexodus-io/nexodus/pull/1236

### DIFF
--- a/cmd/apiserver/main.go
+++ b/cmd/apiserver/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"github.com/redis/go-redis/v9"
 	"log"
 	"net/http"
 	"os"
@@ -201,10 +202,11 @@ func main() {
 				EnvVars: []string{"NEXAPI_COOKIE_KEY"},
 			},
 			&cli.StringFlag{
-				Name:    "redis-server",
-				Usage:   "Redis host:port address",
-				Value:   "redis:6379",
-				EnvVars: []string{"NEXAPI_REDIS_SERVER"},
+				Name:     "redis-server",
+				Usage:    "Redis host:port address",
+				Value:    "redis:6379",
+				EnvVars:  []string{"NEXAPI_REDIS_SERVER"},
+				Required: true,
 			},
 			&cli.IntFlag{
 				Name:    "redis-db",
@@ -235,7 +237,12 @@ func main() {
 
 				store := inmem.New()
 
-				api, err := handlers.NewAPI(ctx, logger.Sugar(), db, ipam, fflags, store, signalBus)
+				redisClient := redis.NewClient(&redis.Options{
+					Addr: cCtx.String("redis-server"),
+					DB:   cCtx.Int("redis-db"),
+				})
+
+				api, err := handlers.NewAPI(ctx, logger.Sugar(), db, ipam, fflags, store, signalBus, redisClient)
 				if err != nil {
 					log.Fatal(err)
 				}
--- a/internal/handlers/user.go
+++ b/internal/handlers/user.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/redis/go-redis/v9"
 	"net/http"
 	"time"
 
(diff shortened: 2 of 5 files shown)


cmd/apiserver/main.go
```diff
@@ -235,7 +237,12 @@ func main() {
 
 				store := inmem.New()
 
-				api, err := handlers.NewAPI(ctx, logger.Sugar(), db, ipam, fflags, store, signalBus)
+				redisClient := redis.NewClient(&redis.Options{
+					Addr: cCtx.String("redis-server"),
+					DB:   cCtx.Int("redis-db"),
```
chirino, line 242: This will make the client use the same DB and key space as the session store. We need to be careful that the keys we store for users don't conflict with the session keys.  Later I make a recommendation to add a key prefix to our keys.. that might be enough to distinguish our keys from the sessions store.. not sure.

internal/handlers/user.go
```diff
@@ -22,14 +22,24 @@ const AuthUserName string = "_nexodus.UserName"
...
 
+// CacheExp Zero expiration means the key has no expiration time.
+const CacheExp time.Duration = 0
+
 func (api *API) CreateUserIfNotExists() gin.HandlerFunc {
 	return func(c *gin.Context) {
 		id := c.GetString(gin.AuthUserKey)
 		username := c.GetString(AuthUserName)
-		_, err := api.createUserIfNotExists(c.Request.Context(), id, username)
+		cachedUsername, err := api.redis.Get(c.Request.Context(), id).Result()
```
chirino, line 32: I would recommend adding a prefix to the cache key like "user:" or "create_user:" so that we can more confidently use the same Redis database as a cache for other things in the future.

internal/handlers/user.go
```diff
@@ -22,14 +22,24 @@ const AuthUserName string = "_nexodus.UserName"
...
 		if err != nil {
-			_ = c.AbortWithError(http.StatusInternalServerError, err)
-			return
+			api.logger.Infof("failed to find user in the cache:%s", err)
```
chirino, line 34: A cache miss will be a common scenario.  If the error is because of a cache miss, I would recommend debug logging level, otherwise if the error is due to a redis system error, a warn level would be better.

chirino (changes requested):

internal/handlers/user.go
```diff
@@ -22,14 +23,30 @@ const AuthUserName string = "_nexodus.UserName"
...
-		_, err := api.createUserIfNotExists(c.Request.Context(), id, username)
+		prefixId := fmt.Sprintf("%s:%s", CachePrefix, id)
+		cachedUsername, err := api.redis.Get(c.Request.Context(), prefixId).Result()
 		if err != nil {
-			_ = c.AbortWithError(http.StatusInternalServerError, err)
-			return
+			if errors.Is(err, redis.Nil) {
+				api.logger.Debugf("user id doesn't exits in the cache:%s", err)
+			}
+			api.logger.Warnf("failed to find user in the cache:%s", err)
```
chirino, line 40: Should this be in an else branch?

internal/handlers/user.go
```diff
@@ -22,14 +23,30 @@ const AuthUserName string = "_nexodus.UserName"
...
+		}
+		if cachedUsername == "" {
+			uuid, err := api.createUserIfNotExists(c.Request.Context(), id, username)
+			if err != nil {
+				_ = c.AbortWithError(http.StatusInternalServerError, err)
+				return
+			}
+			prefixId = fmt.Sprintf("%s:%s", CachePrefix, uuid)
```
chirino, line 48: You could reuse the prefixId

internal/handlers/user.go
akoserwal: I agree

internal/handlers/user.go
akoserwal: fixed

cmd/apiserver/main.go
akoserwal: updated

internal/handlers/user.go
akoserwal: added

chirino (approved):
Lgtm

akoserwal:
@chirino please can you take a look again?

chirino (approved):
Lgtm

russellb:
@mergifyio rebase

PR #1248, opened 2023-07-05
Proposal for device registration flow using keycloak dynamic client registration

akoserwal:


docs/development/design/device-registration.md
```diff
@@ -0,0 +1,47 @@
+# Device registration flow using dynamic client registration in keycloak
+
+> [Issue #1209](https://github.com/nexodus-io/nexodus/issues/1209)
+
+## Summary
+* Device registration through access a token
```
russellb, line 6: @akoserwal take a look at these markdown-lint errors.

akoserwal:
> @akoserwal could you update the registration flow diagram so that it's more clear which methods in `nexctl` or `nexd` would initiate those API calls. And maybe even sketch up what those API URL paths would be?

@chirino Added steps with nexd command. let me know if it helps.

docs/development/design/device-registration.md
```diff
@@ -0,0 +1,47 @@
...
+### Solution
+
+With the help of the [keycloak dynamic client registration](https://medium.com/keycloak/dynamic-client-registration-in-keycloak-4dd1c5cd5e69) feature, I am proposing to generate an initial access token with a configurable max count and expiration date.
+
+![design-registration flow](./diagrams/design-registration-d1.png)
+
+#### Steps to register device
+
+1. Register a initial access token with a count number and expiration date.
+    1. command: `nexd --username --password generate-intial-token --count 10 (clients) --exp date`
```
chirino, line 26: We use `nexctl` (not `nexd`)  to operate against the control plane.   So I think a command like this would make more sense:
`nexctl registration-token create --count 10 (clients) --exp date`

docs/development/design/device-registration.md
```diff
@@ -0,0 +1,47 @@
...
+    2. Store the intial access token in `/var/lib/nexd/apitoken.json`
+2. Register a service account for device using the initial access token
+    1. command: `nexd --intial-token https://try.nexodus.127.0.0.1.nip.io`
```
chirino, line 29: I would rename `inital-token` to `registration-token` since it's more descriptive of what the token is for.

docs/development/design/device-registration.md
```diff
@@ -0,0 +1,47 @@
...
+    2. Store the client-id and client-secret in `/var/lib/nexd/apitoken.json`
+3. Device registration
+    1. command: `nexd --clientid <sa> --secret <secret> https://try.nexodus.127.0.0.1.nip.io`
```
chirino, line 32: I don't think that we should make the user manage the subsequent client id and secret the nexd gets in step 2.  Step 2 should persist the client id/secret to disk (to the same place we currently store the oauth refresh tokens), and on a `nexd` restart, load the previously stored client id/secret and skip what would be done in step two.  That way nexd can be started with the same arguments on every startup.

chirino (changes requested):

docs/development/design/device-registration.md
akoserwal: I agree

chirino (approved):
lgtm

Issue #1231, opened 2023-06-26
CLI flags inconsistent `nexctl --host` vs `nexd --service-url`


chirino:
### Describe the bug

CLI flags are inconsistent between `nexctl --host` and `nexd --service-url`.  They seem to be for the same purpose.  

### How to Reproduce

See usage info for both commands.

### Expected Behavior

Both commands should use the same flag.

russellb:
I vote `--service-url`, though the behavior and defaults are not quite the same.

The default in `nexctl` has an `api` prefix while the default for `nexd` does not. It would be nice to make the behavior of the options the same as part of making the option names the same.

chirino:
> I vote `--service-url`, though the behavior and defaults are not quite the same.
> 
> The default in `nexctl` has an `api` prefix while the default for `nexd` does not. It would be nice to make the behavior of the options the same as part of making the option names the same.

I agree with all the above.

PR #1249, opened 2023-07-06
nexctl: use `--service-url` instead of `--host`.

chirino:
- also set better usage description for the command.
- udpates docs usage.
- `—host` is still supported but deprecated.
- rename controllerUrl field to apiURL.

fixes: #1231

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -78,7 +78,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
         with:
-          node-version: 16
+          node-version: 19
       - run: npm ci
         working-directory: ui
       - run: npx prettier --check .
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -76,15 +76,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: actions/setup-node@v3
-        with:
-          node-version: 19
-      - run: npm ci
-        working-directory: ui
-      - run: npx prettier --check .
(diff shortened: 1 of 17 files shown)


ui/src/App.tsx
```diff
@@ -36,13 +36,13 @@ const authProvider = goOidcAgentAuthProvider(backend);
 const baseDataProvider = simpleRestProvider(
   `${backend}/api`,
   fetchJson,
-  "X-Total-Count"
+  "X-Total-Count",
```
russellb, line 39: It looks like changes to the UI broke the ui-lint job

russellb (approved):
lgtm other than the CI failure

ui/src/App.tsx
chirino: I just ran the `make all -j`, and it generated this.. going to exclude it from the PR.. but not sure why what's getting generated is not right.

chirino:
sad: now the generate step fails because it's generating the UI files which breaks the lint.

.github/workflows/build.yml
```diff
@@ -76,15 +76,8 @@ jobs:
...
-        with:
-          node-version: 16
-      - run: npm ci
-        working-directory: ui
-      - run: npx prettier --check .
-        working-directory: ui
-      - run: npm run build
-        working-directory: ui
+      - name: UI Lint
+        run: make ui-lint ui
```
russellb, line 80: There's no `ui` Makefile target. I think it passes because a `ui` directory exists. I would drop it from here, though.

russellb (approved):
I noticed a minor thing, but I'd rather not force another CI run over it

PR #1259, opened 2023-07-10
api: Fix a couple of issues with org create

russellb:
I got a 500 error creating a new organization in my development
environment. This commit addresses a couple of issues I found while
looking into this.

1. Ensure that the default IPAM namespace has been created. The 500
   error I hit was because it didn't exist. It gets created when a
   user first logs in. Since this was an existing environment, I
   wasn't going through that code path. Now this code does the same
   thing and ensures it exists.

2. Once it worked, I noticed that the security group ID was empty in
   the result I received from the API. Fix the code to put the
   security group ID on the org result returned from the API.

3. Add error logging that I put in place to see what was causing my
   first error.

### DIFF
--- a/internal/handlers/organization.go
+++ b/internal/handlers/organization.go
@@ -3,6 +3,9 @@ package handlers
 import (
 	"errors"
 	"fmt"
+	"net/http"
+	"strconv"
+
 	"github.com/gin-gonic/gin"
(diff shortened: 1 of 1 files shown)


nerdalert (approved):
LGTM

PR #1260, opened 2023-07-10
signalbus: Change msg to debug

russellb:
This message comes up every 90 seconds even when the system is sitting
idle. It seems better as a debug message.


nerdalert (approved):
LGTM

PR #1263, opened 2023-07-13
Add service url to net router docs

nerdalert:


docs/user-guide/network-routers.md
```diff
@@ -11,13 +11,13 @@ The situations where this can be useful are wide-ranging. They can span from con
 - Site-A Configuration: You simply specify the network prefix you want to advertise and Nexodus will use the physical interface that contains the default route to connect to the non-Nexodus nodes.
 
 ```terminal
-nexd router --child-prefix 192.168.1.0/24 --network-router
+nexd router --child-prefix 192.168.1.0/24 --network-router --service-url https://try.nexodus.io
```
russellb, line 14: For most people this will not be necessary. The default for production builds is to use that URL.

I also don't think putting the option at this spot is valid. It has to come before `router`.

My general preference is to not include the option in the docs unless we're specifically documenting about how to specify which service to use. That makes the docs a little less verbose and focused on the options under discussion.

russellb (changes requested):

docs/user-guide/network-routers.md
nerdalert: thanks! didn't realize it defaults to it ><

Issue #1185, opened 2023-06-14
IPAM resource leak when deleting organizations


russellb:
This is something I noticed while reviewing #1007.

It looks like we only release the IPv4 prefix when deleting an organization. The IPv6 prefix and the IPAM namespace appear to not be cleaned up. I have not done any validation on this yet. This is just an observation from reading the code.

In the CreateOrganization() handler, we have:

```go
		// Create the organization in IPAM
		if err := api.ipam.CreateNamespace(ctx, org.ID); err != nil {
			return err
		}

		if err := api.ipam.AssignPrefix(ctx, org.ID, request.IpCidr); err != nil {
			return err
		}

		if err := api.ipam.AssignPrefix(ctx, org.ID, request.IpCidrV6); err != nil {
			return err
		}
```

and later in the DeleteOrganization() handler:

```go
	orgCIDR := org.IpCidr

	if orgCIDR != "" {
		if err := api.ipam.ReleasePrefix(c.Request.Context(), org.ID, orgCIDR); err != nil {
			c.JSON(http.StatusInternalServerError, models.NewApiInternalError(fmt.Errorf("failed to release ipam organization prefix: %w", err)))
			return
		}
	}
```

akoserwal:
@russellb is this still an issue, or can be closed?

akoserwal:
@russellb I will work on it

PR #1265, opened 2023-07-14
release prefix for cidrv6 when deleting org

akoserwal:
Address issues: https://github.com/nexodus-io/nexodus/issues/1185

### DIFF
--- a/internal/handlers/organization.go
+++ b/internal/handlers/organization.go
@@ -541,12 +541,19 @@ func (api *API) DeleteOrganization(c *gin.Context) {
 	}
 
 	orgCIDR := org.IpCidr
+	orgCIDRV6 := org.IpCidrV6
 
-	if orgCIDR != "" {
+	if org.PrivateCidr {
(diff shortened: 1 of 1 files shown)


internal/handlers/organization.go
```diff
@@ -541,12 +541,19 @@ func (api *API) DeleteOrganization(c *gin.Context) {
...
 	orgCIDR := org.IpCidr
+	orgCIDRV6 := org.IpCidrV6
 
 	if orgCIDR != "" {
 		if err := api.ipam.ReleasePrefix(c.Request.Context(), ipamNamespace, orgCIDR); err != nil {
 			c.JSON(http.StatusInternalServerError, models.NewApiInternalError(fmt.Errorf("failed to release ipam organization prefix: %w", err)))
 			return
 		}
 	}
+	if orgCIDRV6 != "" {
```
chirino, line 552: This and the previous IF statements should also be in an `if org.PrivateCidr  { .... }` block.  Otherwise, we may try to delete the shared cidrs.

chirino (approved):
Lgtm

PR #1278, opened 2023-07-21
Consolidate all nexd persistent state into a single json file in thes state dir.

chirino:
Default the state dir and the socket dir under the user’s home directory if not being run as root.


chirino:
Also need to also migrate the previous state files to this new single one.

russellb (changes requested):
I believe I gave similar feedback on your original PR, but I think it still applies.

This PR treats the wireguard keys as internal state, but I don't think that's correct.

Wireguard keys are the only way we have to control which device `nexd` comes up as. If someone wants to pre-create a device, the keys are what must be used to come up as that device. We may have better ways to manage device identity in the future, but right now, control over the keys is all we have.

Because of the above, I view `/etc` as the correct base dir for the keys (configuration) vs `/var/lib` (variable state information).

The linux filesystem hierarchy standard says this about variable state:

> Users must never need to modify files in /var/lib to configure a package's operation, and the specific file hierarchy used to store the data must not be exposed to regular users.

https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch05s08.html

For now, wireguard keys are still something that users may need to modify.

chirino:
I don't understand the workflow that currently exists that would require a user to manually configure their wg keys. At least it's not documented anywhere. If your proposing something new, could we document the use case first to see if it's something we want to support going forward.

russellb:
> I don't understand the workflow that currently exists that would require a user to manually configure their wg keys. At least it's not documented anywhere. If your proposing something new, could we document the use case first to see if it's something we want to support going forward.

This is slightly different than what I described before, but these docs provide wireguard keys as configuration for each nexd pod. The reason is to ensure that `nexd` comes up as the same device as a restart. A PVC would work, but it's much nicer not requiring one. Removing key config means I would need a PVC (with the current nexd code).

https://docs.nexodus.io/user-guide/scenarios/nexd-proxy-in-kube/

This could be extended to run `nexd` outside of kube with those keys once to ensure a nexodus device exists for each keypair. Then you know the tunnel IPs associated with each kube Deployment of nexd ahead of time, which lets you be more declarative with the kube deployments.

In some dev scripts, I'm having to `kubectl exec` into each Pod to figure out their tunnel IPs, and then I'm using that in configuration for another Deployment. I'd rather have the nexodus Device details upfront.

Another use case is running two nexd proxy Deployments. If there are rules on both sides referencing the other side, you can't generate valid manifests without knowing the tunnel IPs up front.

tl;dr perhaps is I want to know my nexodus Device details prior to writing a manifest to run `nexd` in Kube.

Right now, controlling the keys is the only way to do this. I would agree that it's not great, but I don't want to remove it in favor of something that doesn't exist yet.

chirino:
Well, until we 

> A PVC would work, but it's much nicer not requiring one. Removing key config means I would need a PVC (with the current nexd code).

Ok got it, so it's mostly about supporting a cheaper way to maintain nexd state in a Kube env.  Don't think this change really affects how you get/discover the device tunnel IP.  If nexd could directly store it's state in a kube secret, would that better solve the issue?

russellb:
> Well, until we
> 
> > A PVC would work, but it's much nicer not requiring one. Removing key config means I would need a PVC (with the current nexd code).
> 
> Ok got it, so it's mostly about supporting a cheaper way to maintain nexd state in a Kube env. Don't think this change really affects how you get/discover the device tunnel IP. If nexd could directly store it's state in a kube secret, would that better solve the issue?

A couple of thoughts ...

1) It would remove the option of being able to discover the tunnel IP prior to deploying into Kube. Right now you can do that by pre-creating devices associated with a keypair.

2) I think this could create a race condition where devices could be leaked. If `nexd` creates a device, but gets killed prior to stashing it in a Secret, that device is orphaned in Nexodus. I'm not sure how to avoid that.

chirino:
> A couple of thoughts ...
> 
> 1. It would remove the option of being able to discover the tunnel IP prior to deploying into Kube. Right now you can do that by pre-creating devices associated with a keypair.

Maybe... A user could backup the state from a node he manually pre-creates the device on and restore it to kube.

> 2. I think this could create a race condition where devices could be leaked. If `nexd` creates a device, but gets killed prior to stashing it in a Secret, that device is orphaned in Nexodus. I'm not sure how to avoid that.

Won't happen.  After the wg keys are generated, the state is stored.  This happens before the device is registered with the API server.

russellb:
> > A couple of thoughts ...
> > 
> > 1. It would remove the option of being able to discover the tunnel IP prior to deploying into Kube. Right now you can do that by pre-creating devices associated with a keypair.
> 
> Maybe... A user could backup the state from a node he manually pre-creates the device on and restore it to kube.

True, but then we're violating the definition of variable state from the filesystem spec, since users aren't supposed to touch it.

I suppose another way to address this is to just add a `--device-id` arg to `nexd`. That way, you can be explicit about which Device `nexd` should operate as. It can update the keys as needed.

Maybe this would go away in the future with better per-device token support, but it would be good enough for now to satisfy my concerns.

What do you think? If you're OK with it, I don't mind adding it.

> > 2. I think this could create a race condition where devices could be leaked. If `nexd` creates a device, but gets killed prior to stashing it in a Secret, that device is orphaned in Nexodus. I'm not sure how to avoid that.
> 
> Won't happen. After the wg keys are generated, the state is stored. This happens before the device is registered with the API server.

Makes sense, thanks.

chirino:
> True, but then we're violating the definition of variable state from the filesystem spec, since users aren't supposed to touch it.

They are not modifying those files.  They are just moving the files to where the process moved to.  This is the same as a DB storing its internal files, you backing them up and then restoring them on different a machine.

> I suppose another way to address this is to just add a `--device-id` arg to `nexd`. That way, you can be explicit about which Device `nexd` should operate as. It can update the keys as needed.

> Maybe this would go away in the future with better per-device token support, but it would be good enough for now to satisfy my concerns.
> 

Yeah.. I think would work once we have per device clientId/passwords to ensure that API requests from a different device can't be used to modify other devices.  

> What do you think? If you're OK with it, I don't mind adding it.

Isn't it just simpler to backup/restore the 1 state file for the device?   Maybe you want a more formal backup/restore process?    Like a `--restore-from-dir=/mnt/backup` option that only kicks in if the state does not exist?

russellb:
> > True, but then we're violating the definition of variable state from the filesystem spec, since users aren't supposed to touch it.
> 
> They are not modifying those files. They are just moving the files to where the process moved to. This is the same as a DB storing its internal files, you backing them up and then restoring them on different a machine.

Good point. Makes sense.

> > I suppose another way to address this is to just add a `--device-id` arg to `nexd`. That way, you can be explicit about which Device `nexd` should operate as. It can update the keys as needed.
> 
> > Maybe this would go away in the future with better per-device token support, but it would be good enough for now to satisfy my concerns.
> 
> Yeah.. I think would work once we have per device clientId/passwords to ensure that API requests from a different device can't be used to modify other devices.
> 
> > What do you think? If you're OK with it, I don't mind adding it.
> 
> Isn't it just simpler to backup/restore the 1 state file for the device? Maybe you want a more formal backup/restore process? Like a `--restore-from-dir=/mnt/backup` option that only kicks in if the state does not exist?

I think I'm envisioning a workflow that doesn't fully exist (number 2 below).

1. What would work now, run nexd once to initialize state. Get the tunnel IPs from the Nexodus API or nexctl. Use that state to ensure nexd comes up as the expected Device.

2. Pre-create a Device in the Nexodus API without running nexd. This is possible in the API, but not in `nexctl`. If we added device create to nexctl, it would work. Specify the device ID from the newly created device as an arg to `nexd`.

Both of these would work to know the details up front for interdependent instances of `nexd` running in Kube clusters, but the second option is much nicer since it doesn't require running `nexd` somewhere up front to create the state file.

I think merging your changes would be fine, and supporting workflow 2 here would be OK until we have a better alternative. I do think per-device auth would replace what I'm talking about, eventually.

integration-tests/proxy_test.go
```diff
@@ -679,17 +681,15 @@ func TestProxyNexctl(t *testing.T) {
...
-		  ],
-		  "ingress": [
-			"tcp:4242:127.0.0.1:8080"
-		  ]
-		}`, node1IP), out)
+	s := state.State{}
+	err = json.Unmarshal([]byte(data), &s)
+	require.NoError(err)
+	require.Equal(state.ProxyRulesConfig{
+		Egress:  []string{fmt.Sprintf("tcp:80:%s:4242", node1IP)},
```
russellb, line 690: minor nit, `net.JoinHostPort` is best practice for joining and IP and port number. It's fine as-is here since it's an IPv4 address, but copy/paste of this code and reuse with an IPv6 address would break.

internal/nexodus/keys.go
```diff
@@ -2,68 +2,86 @@ package nexodus
...
 	return nil
+
 }
 
-// readKeyFile reads the contents of a key file
-func readKeyFile(logger *zap.SugaredLogger, keyFile string) string {
-	if !FileExists(keyFile) {
-		return ""
+// loadLegacyKeys should not be needed after everyone has upgraded to the latest nexd.
+func (ax *Nexodus) loadLegacyKeys() (string, string, error) {
```
russellb, line 45: For new code, can you use `nx`?  `ax` is the old variable name used back when it was apex. I've been trying to migrate to `nx` as we go.

russellb (approved):
I had two minor nits, but neither are important enough to block on.

It's also not clear to me if the `--public-key` and `--private-key` options to `nexd` still work, or even if they still worked prior to this change. If you could take a look, that'd be awesome. I assume we want to drop them whether they work or not.

Issue #1144, opened 2023-06-07
mac: nexd should not assume utun8 is available


russellb:
I ran into a case where `utun8` was already in use on my mac, so `nexd` failed to start because it hard codes `utun8`. It needs to adapt to be able to use an interface not already in use.

PR #1279, opened 2023-07-21
Draft: checking network interface is avaliable instead of using default

akoserwal:
Address: https://github.com/nexodus-io/nexodus/issues/1144

Would this approach work? I couldn't find a better way to check if the network interface is used or unused.

### DIFF
--- a/internal/nexodus/nexodus.go
+++ b/internal/nexodus/nexodus.go
@@ -1011,5 +1011,5 @@ func (ax *Nexodus) defaultTunnelDev() string {
 	if ax.userspaceMode {
 		return ax.defaultTunnelDevUS()
 	}
-	return defaultTunnelDevOS()
+	return avaliableTunnelDevOS()
 }
--- a/internal/nexodus/utils_darwin.go
+++ b/internal/nexodus/utils_darwin.go
@@ -8,6 +8,7 @@ import (
 	"net"
 	"os"
 	"os/exec"
+	"regexp"
 	"strings"
 
 	"github.com/vishvananda/netlink"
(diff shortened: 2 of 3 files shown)


nerdalert:
@akoserwal thanks for taking a look at it! I think how you're going about it makes perfect sense.

- My questions are probably more around the issue from @russellb and best practices for the utun interface on OSX. I fully admit to spinning a wheel and picking utun8 😵‍💫. As far as I could ever tell it was basically don't use one in use but on my machines I have always seen at least `utun0` in use, but often `utun[0-2]` so ¯\\_(ツ)_/¯ 
- If I was to take a wild guess, I would guess that the utun8 interface in the issue could have been another instance of nexd, since the utun interface gets removed when nexd/wireguard-go exits. I think we check for an existing process now so it might have been a manually started wireguard-go interface. 
- In the case of Linux, the kernel based interface does not get removed when nexd exits. So if you were to restart nexd, a new driver interface would be created, e.g. `wg1` that would not be functional since all of the peers would still be tied to `wg0`. The only alternative would be to try and figure out if the old wg(n) interface was our creation and delete it similar to `removeExistingInterface()`
- **TLDR**; Should this be for OSX/Windows only since they are user space implementations that cleanup old interfaces when nexd previously exited?

> Worth noting is whatever iface gets picked, is attempted to be removed further down in the code, primarily for Linux devices. That could probably be more elegant but its so fast to converge I've never seen any delay and its nice to be completely declarative to clean up any old peers/routes that are no longer operative.

```
	ax.tunnelIface = ax.defaultTunnelDev()

	if ax.relay {
		ax.listenPort = WgDefaultPort
	}

	if err := ax.checkUnsupportedConfigs(); err != nil {
		return nil, err
	}

	if err := prepOS(logger); err != nil {
		return nil, err
	}

	// remove orphaned wg interfaces from previous node joins
	ax.removeExistingInterface()
```

Thanks!

russellb:
It was a while ago when I hit this issue and I didn't do myself any favors by filing the issue with so little detail. I can try to recreate it, though. I believe it happened when I was using some other VPN app at the same time -- probably OpenVPN for work.

nerdalert:
From everything I gathered I never saw a way to remove the interface without unless the process tied to it such as wireguard-go exits or the darwin host reboots. I don't even think Darwin exposes the ability to remove physical interfaces without the physical interfaces being removed.

Ideally if an existing process is running, we don't let the daemon start. This is the scenario for Linux currently via nexd but that isn't supported in Darwin so you can have multiple copies running (opened #1291 for that).

I think in general, checking if there is a nexd process is already running, and if it isn't grabbing an available utun interface is just fine. Same for windoze. The process check doesn't need to be part of this since it's pretty much a user error if they are trying to start multiple instances of nexd.

For Linux, the current workflow should probably be left as is unless we were to add a trap for host cleanup (which there are scenarios we might need this in the future but thats outside of the scope of this, e.g. nftables in particular).

ty!

akoserwal:
> From everything I gathered I never saw a way to remove the interface without unless the process tied to it such as wireguard-go exits or the darwin host reboots. I don't even think Darwin exposes the ability to remove physical interfaces without the physical interfaces being removed.
> 
> Ideally if an existing process is running, we don't let the daemon start. This is the scenario for Linux currently via nexd but that isn't supported in Darwin so you can have multiple copies running (opened #1291 for that).
> 
> I think in general, checking if there is a nexd process is already running, and if it isn't grabbing an available utun interface is just fine. Same for windoze. The process check doesn't need to be part of this since it's pretty much a user error if they are trying to start multiple instances of nexd.
> 
> For Linux, the current workflow should probably be left as is unless we were to add a trap for host cleanup (which there are scenarios we might need this in the future but thats outside of the scope of this, e.g. nftables in particular).
> 
> sorry for the bikeshedding here Abhishek 😅 ty!

@nerdalert not at all, thank you so much for adding the details and helping with the discussion.

akoserwal:
@russellb @nerdalert Do you think we need this?

russellb:
> @russellb @nerdalert Do you think we need this?

we need something on mac. I can't run `nexd` on my mac sometimes if `utun8` is already in use.

PR #1281, opened 2023-07-23
Add a no summary e2e option for debugging raw output

nerdalert:


Makefile
```diff
@@ -344,6 +344,10 @@ e2e: e2eprereqs dist/nexd dist/nexctl image-nexd ## Run e2e tests
 	gotestsum --format standard-quiet -- \
 		-race --tags=integration ./integration-tests/... $(shell [ -z "$$NEX_TEST" ] || echo "-run $$NEX_TEST" )
 
+.PHONY: e2e-no-summary
+e2e-no-summary: e2eprereqs dist/nexd dist/nexctl image-nexd ## Run e2e tests with verbose output
+	go test -race -v --tags=integration ./integration-tests/... $(shell [ -z "$$NEX_TEST" ] || echo "-run $$NEX_TEST" )
```
russellb, line 349: I'm not a huge fan of duplicating the `go test` arguments. An alternative would be to rewrite the `e2e` target like this.

First add this to the existing if/else block for `NOISY_BUILD`

```
ifeq ($(NOISY_BUILD),)
    ...
    GO_TEST=gotestsum --format standard-quiet --
else
    ...
    GO_TEST=go test -v
endif
```

and then change the `e2e` target:

```
.PHONY: e2e
e2e: e2eprereqs dist/nexd dist/nexctl image-nexd ## Run e2e tests
	$(GO_TEST) -race --tags=integration ./integration-tests/... $(shell [ -z "$$NEX_TEST" ] || echo "-run $$NEX_TEST" )
```

but maybe what would be even better is to use `gotestsum` in both cases, but with a different output format.  See this page: https://github.com/gotestyourself/gotestsum#output-format

```
ifeq ($(NOISY_BUILD),)
    ...
    GOTESTSUM_FMT=standard-quiet
else
    ...
    GOTESTSUM_FMT=standard-verbose
endif
```

and then change the `e2e` target:

```
.PHONY: e2e
e2e: e2eprereqs dist/nexd dist/nexctl image-nexd ## Run e2e tests
	gotestsum --format $(GOTESTSUM_FMT) \
                -race --tags=integration ./integration-tests/... $(shell [ -z "$$NEX_TEST" ] || echo "-run $$NEX_TEST" )
```

I think that still provides full output, but perhaps without it interleaved? That would be nice, but I haven't tried it myself to see for sure.

russellb (changes requested):

Makefile
nerdalert: WFM, done!

Makefile
```diff
@@ -340,8 +342,8 @@ dist/.generate: $(SWAGGER_YAML) dist/.ui-fmt | dist
 	$(CMD_PREFIX) touch $@
 
 .PHONY: e2e
-e2e: e2eprereqs dist/nexd dist/nexctl image-nexd ## Run e2e tests
-	gotestsum --format standard-quiet -- \
+e2e: e2eprereqs dist/nexd dist/nexctl image-nexd ## Run e2e verbose tests
```
russellb, line 345:
```suggestion
e2e: e2eprereqs dist/nexd dist/nexctl image-nexd ## Run e2e tests
```

russellb (approved):
one unnecessary doc string change in there, but otherwise lgtm!

PR #1293, opened 2023-07-28
Add exit node client/server support

nerdalert:
- Enables the ability to funnel all device traffic through a designated exit node


nerdalert:
TODO:
- [ ] Add docs and user facing diagrams
- [ ] Add exit node client enable flag for enabling an exit node client. Currently enabled and disabled with `nexctl exit-node enable` and `nexctl exit-node disable`
- [ ] Add e2e tests and e2e container mocking exit origin and client nodes.

internal/nexodus/nexodus.go
```diff
@@ -566,6 +586,11 @@ func (nx *Nexodus) Start(ctx context.Context, wg *sync.WaitGroup) error {
 		for _, proxy := range nx.proxies {
 			proxy.Start(ctx, wg, nx.userspaceNet)
 		}
+		if nx.exitNode.exitNodeClientEnabled {
+			if err := nx.ExitNodeClientSetup(); err != nil {
+				nx.logger.Errorf("failed to enable this device as an exit-node client: %v", err)
+			}
+		}
```
russellb, line 593: This functionality needs to be integrated into the reconcile loop instead of just once at startup or when triggered via `nexctl`.

internal/nexodus/utils_darwin.go
```diff
@@ -142,3 +142,8 @@ func isIPv6Supported() bool {
 
 	return true
 }
+
+// getDefaultGatewayIPv4 not currently implemented for darwin
+func getDefaultGatewayIPv4() (string, error) {
+	return "", nil
```
russellb, line 148: might as well return an error

internal/nexodus/utils_windows.go
```diff
@@ -121,3 +121,8 @@ func isIPv6Supported() bool {
 
 	return true
 }
+
+// getDefaultGatewayIPv4 not currently implemented for windows
+func getDefaultGatewayIPv4() (string, error) {
+	return "", nil
```
russellb, line 127: might as well return an error

internal/nexodus/policy_windows.go
```diff
@@ -11,3 +13,13 @@ func (nx *Nexodus) processSecurityGroupRules() error {
...
+
+// runNftCmd for windows build purposes
+func runNftCmd(logger *zap.SugaredLogger, cmd []string) (string, error) {
+	return "", nil
+}
+
+// nfTableDrop for windows build purposes
+func (nx *Nexodus) nfTableDrop(table string) error {
+	return nil
+}
```
russellb, line 25: how about returning errors in case they're ever called by accident for some reason

internal/nexodus/policy_darwin.go
```diff
@@ -11,3 +13,13 @@ func (nx *Nexodus) processSecurityGroupRules() error {
...
+
+// runNftCmd for Darwin build purposes
+func runNftCmd(logger *zap.SugaredLogger, cmd []string) (string, error) {
+	return "", nil
+}
+
+// nfTableDrop for Darwin build purposes
+func (nx *Nexodus) nfTableDrop(table string) error {
+	return nil
+}
```
russellb, line 25: return errors in case these are ever called by accident

internal/nexodus/nexodus_linux.go
```diff
@@ -92,11 +93,14 @@ func (nx *Nexodus) setupInterfaceOS() error {
 		return fmt.Errorf("%w", interfaceErr)
 	}
 	defer util.IgnoreError(c.Close)
+
+	wgFwMark := 51820
```
russellb, line 97: Can you put this in a constant with some docs in the code that explain what this is for? You can also just reference the design doc to avoid repeating yourself.

internal/nexodus/nexodus.go
```diff
@@ -518,6 +532,12 @@ func (nx *Nexodus) Start(ctx context.Context, wg *sync.WaitGroup) error {
 		}
 	}
 
+	if nx.exitNode.exitNodeOriginEnabled {
+		if err := nx.exitNodeOriginSetup(); err != nil {
+			return fmt.Errorf("failed to setup this device as an exit-node: %w", err)
+		}
+	}
```
russellb, line 539: This will need to be integrated into the reconcile loop instead of only at startup to ensure that the proper configuration stays in place.

russellb (changes requested):

internal/nexodus/exit_node.go
```diff
@@ -0,0 +1,234 @@
...
+
+	for _, deviceEntry := range nx.deviceCache {
+		for _, allowedIp := range deviceEntry.device.AllowedIps {
+			if util.IsDefaultIPv4Route(allowedIp) {
+				// assign the device advertising a default route as the exit node server/origin node
+				nx.exitNode.exitNodeOrigins[0] = wgPeerConfig{
+					PublicKey:           deviceEntry.device.PublicKey,
+					Endpoint:            deviceEntry.device.EndpointLocalAddressIp4,
+					AllowedIPs:          deviceEntry.device.AllowedIps,
+					PersistentKeepAlive: "25",
```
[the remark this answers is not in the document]
nerdalert: See https://www.wireguard.com/quickstart/ 

```
This is called persistent keepalives. When this option is enabled, a keepalive packet is sent to the server endpoint once every interval seconds. A sensible interval that works with a wide variety of firewalls is 25 seconds. Setting it to 0 turns the feature off, which is the default, since most users will not need this, and it makes WireGuard slightly more chatty. This feature may be specified by adding the PersistentKeepalive = field to a peer in the configuration file, or setting persistent-keepalive at the command line. If you don't need this feature, don't enable it. But if you're behind NAT or a firewall and you want to receive incoming connections long after network traffic has gone silent, this option will keep the "connection" open in the eyes of NAT.
```

internal/nexodus/nexodus_linux.go
nerdalert: Yeah, this looks goofy because `FirewallMark *int` and Go, you can't take the address of a constant. Added a const.

```
const 	wgFwMark         = 51820

	fwMark := wgFwMark

	err = c.ConfigureDevice(nx.tunnelIface, wgtypes.Config{
		PrivateKey:   &privateKey,
		ListenPort:   &listenPort,
		ReplacePeers: true,
		Peers:        nil,
		FirewallMark: &fwMark,
	})
```

internal/nexodus/utils.go
```diff
@@ -207,3 +208,47 @@ func parseIPfromAddrPort(addrPort string) string {
 
 	return ip.Addr().String()
 }
+
+// getInterfaceByIP returns the name of the interface that has the given IP address assigned to it.
+func getInterfaceFromIP(ip string) (string, error) {
```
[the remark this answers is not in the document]
nerdalert: Since its for endpoints and we only support v4 now. Added to docs.

internal/nexodus/exit_node.go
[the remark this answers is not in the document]
russellb: in `wg_deploy.go` we have:

```
const (
	persistentKeepalive = "20"
)
```

We should use a consistent default. You can just use that one. Change it to 25 if you'd like.

integration-tests/routing_test.go
```diff
@@ -0,0 +1,182 @@
...
+	"unicode"
+
+	"github.com/docker/docker/api/types/container"
+	"github.com/stretchr/testify/require"
+	"github.com/testcontainers/testcontainers-go"
+)
+
+// TestExitNode performs the following exit node tests:
+// 1. Start two nodes, one is an exit node client and the other is an exit node origin/server.
+// 2. Curl an Internet address from the exit node client to ensure connectivity via the origin server.
```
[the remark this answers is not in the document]
nerdalert: Will file an issue. I can't think of a way to do it without creating another container but I want to keep it simple if that wfy until all the potential bugs are caught.

internal/nexodus/nexodus.go
nerdalert: Good call. This is big enough I'd like to do in a followup. The code changes are relatively minimal, but I do want to allocate additional time testing the scenario (particularly exit node client side) of it flipping between 5g/wired.

https://github.com/nexodus-io/nexodus/issues/1408

internal/nexodus/exit_node.go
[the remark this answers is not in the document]
nerdalert: Ty!

internal/nexodus/nexodus.go
nerdalert: Agree, tracked in #1408 ty ty.

internal/nexodus/exit_node.go
```diff
@@ -0,0 +1,234 @@
...
+func (nx *Nexodus) ExitNodeClientSetup() error {
+	nx.exitNode.exitNodeClientEnabled = true
+
+	// Lock deviceCache for read
+	nx.deviceCacheLock.RLock()
+	defer nx.deviceCacheLock.RUnlock()
+
+	for _, deviceEntry := range nx.deviceCache {
+		for _, allowedIp := range deviceEntry.device.AllowedIps {
+			if util.IsDefaultIPv4Route(allowedIp) {
```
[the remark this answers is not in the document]
nerdalert: Done! #1409

integration-tests/routing_test.go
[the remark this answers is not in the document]
russellb: OK, I'm fine with an issue for this.

russellb (approved):

PR #1320, opened 2023-08-07
fix: `make clear-db` and `make recreate-db`

chirino:
They were leaving the service in a bad state where clients would get API errors.  We now made redis non-peristent and bounce it after clearing the DB.


nerdalert (approved):
LGTM, ty!

PR #1338, opened 2023-08-22
Don’t compile in the kstore code introduced in dd822f7

chirino:
That PR tripled the size of the nexd binary, so we decided to not  compile it in by default to rethink how to accomplish the same goal a  different way.

To build with the kstore code run: `NEXODUS_BUILD_TAGS="kubernetes" make clean dist/nexd`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,6 +38,9 @@ NEXODUS_BUILD_FLAGS:=
 ifneq ($(NEXODUS_PPROF),)
     NEXODUS_BUILD_FLAGS+=-tags pprof
 endif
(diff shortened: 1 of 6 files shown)


russellb:
I'm OK with this for now. I'm not excited about another dimension in the build matrix, though. I think I'd rather see it refactored into a separate binary eventually. What do you think?

chirino:
Opened issue #1339 to remind us to drop the bloat at some point so we can enable the kstore by default again.

russellb (approved):

PR #1342, opened 2023-08-24
Add wireguard userspace tools to the nexd image

nerdalert:
- Adds about 30m for the wg user space binary. Will be useful to debug user issues and QA early on. We currently get information such as peer listings and stats directly from the wg interface but if nexd isn't running those stats aren't available. Since nexd does not rely on this it is removable at any time but would like to have it available for time saving for installs for now.


nerdalert:
Also had an OC moment and alphabetized Containerfile.dev

russellb (approved):

Issue #1343, opened 2023-08-24
On Windows fix the `stun dialing timed out dial udp4`


nerdalert:
### Describe the bug

On Windows, you cannot currently access the socket bound by wireguard to source STUNs.

```
{"level":"error","caller":"stun/client.go:15","msg":"stun dialing timed out dial udp4 :65013->172.253.120.127:19302: bind: An attempt was made to access a socket in a way forbidden by its access permissions."}
```

- Immediate need is to suppress the error log for windows and/or skip the STUN for windows OS.

- A short term resolution is for Windows, STUN from a random unbound port and only look for reflexive IP address changes as opposed to IP:Port binding changes. If a reflexive IP change is detected, rejoin the mesh so that the new peer address is advertised.

- A longer term resolution is to STUN via similar techniques as Darwin or Linux.

### How to Reproduce

-

### Expected Behavior

-

nerdalert:
Also suppress this log for windows.

```
{"level":"error","caller":"nexodus/route_windows.go:21","msg":"route add failed: no windows route added: failed to run \"netsh int ipv4 add route 100.64.0.1/32 wg0\": exit status 1 (The object already exists.\r\n\r\n\r\n)"}
{"level":"error","caller":"nexodus/wg_deploy.go:42","msg":"Failed to handle peer route: no windows route added: failed to run \"netsh int ipv4 add route 100.64.0.1/32 wg0\": exit status 1 (The object already exists.\r\n\r\n\r\n)"}
{"level":"error","caller":"nexodus/nexodus.go:681","msg":"Failed to reconcile state with the nexodus API server: no windows route added: failed to run \"netsh int ipv4 add route 100.64.0.1/32 wg0\": exit status 1 (The object already exists.\r\n\r\n\r\n)"}
```

PR #1345, opened 2023-08-25
Cleaning up windows logging and routing

nerdalert:
- Supressed debug logs from failed STUN lookups since windows is currently unable to resuse the wg port for STUNs.
- Filled out some no-ops in unimplemented routing functions v4/v6.
- More details in #1343


russellb (approved):
lgtm, but i'm holding it for a moment so I can fix the issue ID in the commit message to reference the correct bug

PR #1346, opened 2023-08-25
Support scripts for scale testing

vishnoianil:
* Moved scale related scripts to scale directory
* Add script to check connectivity within each org
* Add script to scrap error from each container running nexd


nerdalert (approved):
LGTM ty

PR #1358, opened 2023-08-30
Playground manifest configuration updates to support 10K devices

vishnoianil:
Commit contain following changes that is required to scale control plan to support 10K devices.

- Use pgbouncer for postgres database connection pooling to ensure clients doesn't run out of connections for requests.
- Update apiserver, ipam and auth deployment to use pgbouncer rather than directly using postgres db.
- Update base auth deployment to explicitly set resource limits, so that overlays can override it.
- Add patroni configuration to base database deployment, to configure max_connections and shared_buffer for database server for scale scenario. Base deployment is set to default values, individual overlays can overwrite it.
- Configure PGBouncer to specify pool size and reserve pool per database, rather than per database server.
- Add envoy files in playground overlay, to update with scale configuration for playground overlay. Playground overlay envoy configuration includes circuit breaker configuration and custom timeout values. 
- Update limitador config to increase the request rate limits.
- Update playground kustomization.yaml to overwrite various limits for scaling playground deployment.

### DIFF
--- a/deploy/nexodus/base/apiserver/deployment.yaml
+++ b/deploy/nexodus/base/apiserver/deployment.yaml
@@ -24,12 +24,12 @@ spec:
               valueFrom:
                 secretKeyRef:
                   name: database-pguser-apiserver
-                  key: host
+                  key: pgbouncer-host
             - name: NEXAPI_DB_PORT
               valueFrom:
                 secretKeyRef:
                   name: database-pguser-apiserver
-                  key: port
+                  key: pgbouncer-port
             - name: NEXAPI_DB_NAME
               valueFrom:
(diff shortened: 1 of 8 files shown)


vishnoianil:
@nerdalert This PR will trigger update for the qa deployment as well, because the pgbouncer is added to the base database deployment (with default values), so pgbouncer will be deployed in qa namespace as well. It should not break anything in qa, as well as in prod deployment (when we promote it.)

vishnoianil:
Looking into integration job failure.

nerdalert (approved):
LGTM

nerdalert:
Restarted e2e

PR #1361, opened 2023-08-31
docs: Add dark mode quickstart overview diagram

russellb:
This probably isn't worth the effort for every diagram we have, but
excalidraw can export images in dark mode, as well. To help make the
quickstart page a bit more polished, automatically switch to a dark
mode diagram when being viewed in that mode.


nerdalert (approved):
LGTM

PR #1373, opened 2023-09-05
Add latency to nexd peers ping outputs

nerdalert:


nerdalert:
Example Output:


```
[05:19:56] (='o'=) [nexctl]$ sudo ./nexctl nexd peers ping
HOSTNAME                     WIREGUARD ADDRESS     LATENCY     CONNECTION STATUS
ip-10-180-9-76               100.64.0.1            41.73ms     ✓ Reachable
ub201                        100.64.0.2            5.29ms      ✓ Reachable
ub202                        100.64.0.3            4.98ms      ✓ Reachable
```

Also changed the return to explicitly return the second ICMP probe since the first can sometimes have a higher latency in general on pings.

Pretty sure this won't break US probing but I didn't test, ty!

russellb (approved):

PR #1376, opened 2023-09-06
Add the option to disable security groups at runtime

nerdalert:


vishnoianil:
LGTM

russellb (commented):
Do you see a use case for this if API performance was not a concern?

nerdalert:
> Do you see a use case for this if API performance was not a concern?

I do in the scenario a user didn't want to install the nftables userspace tool until we migrate that to kernel api calls.

russellb:
> > Do you see a use case for this if API performance was not a concern?
> 
> @russellb good point! I do in the scenario a user didn't want to install the nftables userspace tool until we migrate that to kernel api calls. That said Im not super attached to it or anything if you think its overthinking things.

It feels weird to offer "turn security off." I feel like if policy is defined, we should enforce it or fail.

If it's for dev/test use cases, then I'd be ok with it as a hidden option.

PR #1391, opened 2023-09-12
Implements a shared informer API for nexd

chirino:
Depends on #1382, this provides a shared informer so that a single REST api call can be used to watch multiple types of events like device or security group changes.


chirino:
This PR giving me inconsistent e2e test failures.. Running it against CI to see if it's also seeing the same thing.

russellb (approved):

PR #1394, opened 2023-09-12
rpm: drop aarch64 rpm build for epel9

russellb:
This isn't working, and it looks like nobody uses it anyway. Just drop
it for now.


nerdalert (approved):
LGTM

PR #1400, opened 2023-09-14
Keycloak needs atleast default memory/cpu

vishnoianil:
During scale testing in playground namespace the value were set to minimum, and playground overwrote it with high value. Prod/Qa doesn't overwrite the value, so keycloak atleast needs default cpu/memory limits to run.


nerdalert (approved):
LGTM

PR #1403, opened 2023-09-14
Separate oidc check auth into a seperate endpoint

### LINKED ISSUE #1402: Race condition in UI login
### Describe the bug

Web UI does a double login due to a race in LoginEnd. 

### How to Reproduce

Login to a fresh session. The first login generally fails, that triggers the LoginEnd endpoint that will redirect to the LoginStart endpoint. By that time AuthCheck will be true and the login goes through.

### Expected Behavior

*

### Deployment Environment

_No response_

### Additional Context

_No response_

nerdalert:
- Resolves having to login twice to get into the UI by adding a CheckAuth endpoint to be used by react-admin checkauth instead of /web/login/end which was creating a race condition since AuthProvider RA calls checkauth multiple times during a login along with a long running timer which was triggering login/end after login as well.
- As a side benefit this resolves another issue where a user is kicked out of the web ui after a few minutes and forced to login again.
- Additional client side web console logging added as well.
- Closes #1402


nerdalert:
Logout is still not honoring what should be 30 minutes according to the idle timer in KC `"ssoSessionIdleTimeout": 1800,`. Will figure it out on another PR since logout needs some attention as well.

russellb (approved):

PR #1407, opened 2023-09-14
nexd: Refactor peering method choices

russellb:
Define more concretely the concept of a peering method. Each method
has a name, a set of prerequisites that must be met to use it, and an
implementation of generating wireguard configuration using that
method.

There should be no functional changes other than an updated debug log
message that will now indicate which peering method is in use for a
peer.


russellb:
@mergifyio rebase

nerdalert (approved):
LGTM - peering looked good in ec2 qa 🎉  https://github.com/nexodus-io/nexodus/actions/runs/6202214199/job/16840503685

PR #1423, opened 2023-09-18
Copy over the qa/prod envoy changes to the dev envoy config 

chirino:
and go back to using a single envoy config.


deploy/nexodus/overlays/playground/files/envoy.yaml
```diff
@@ -1,393 +0,0 @@
...
-          "@type": type.googleapis.com/envoy.extensions.transport_sockets.tls.v3.UpstreamTlsContext
-
-    # upstream server: apiserver
-    - name: apiserver
-      connect_timeout: 10s
-      type: STRICT_DNS
-      dns_refresh_rate: 1s
-      dns_lookup_family: V4_ONLY
-      lb_policy: ROUND_ROBIN
-      circuit_breakers:
```
vishnoianil, line 53: @chirino we also need to override these circuit breaker related configuration in the kustomization.yaml for playground (with these specific value). Also it's good to add those config in prod/qa kustomization.yaml with default values, so that we can tweak them in case we want to scale out the prod/qa.

chirino: Btw.. I did copy the circuit breaker config, see: https://github.com/nexodus-io/nexodus/pull/1423/files#diff-1adf750eb6b72619179551461e72d56737aaeba856689dff25f962d6f8138815R53

vishnoianil: you mean bump for dev/qa as well to high value even if they don't need to handle any scale?

chirino: yes. might as well keep them all the same unless have a solid reason to make them different.

chirino: oh yeah... and if we do want to have per overlay Enovy config, perhaps we should carry it as a patch that gets applied during the build to create the overlay version. That way changes to the based envoy configuration automatically roll into overlays.

vishnoianil: > yes. might as well keep them all the same unless have a solid reason to make them different.

Currently only playground has a bumped up values for circuit breaker configurations, so as far as we can use those values in base it should be fine. Generally when we bump up these values, the memory and cpu requirements increases for envoy, and if we bump up these values in base, we probably will have to bump-up resource limits for each overlay to accommodate. Currently with 10K devices, we don't have to worry about it, as the default envoy limits (500m/1000Mi) seems to work fine.

vishnoianil: > oh yeah... and if we do want to have per overlay Enovy config, perhaps we should carry it as a patch that gets applied during the build to create the overlay version. That way changes to the based envoy configuration automatically roll into overlays.

Currently that's pretty much we are doing (all individual envoy's are augmentation on base envoy), but yes patch is any option as well.

vishnoianil (approved):

PR #1424, opened 2023-09-18
Upgrade to latest sqlite version.

chirino:


russellb:
According to go-sqlite3's github readme:

> Latest stable version is v1.14 or later, not v2.
> NOTE: The increase to v2 was an accident. There were no major changes or features.

russellb (approved):

PR #1425, opened 2023-09-18
fix: don’t use the gorm Scopes() method, it’s behavior is more complex than we realized.

chirino:
this also lets up bump the gorm version and still pass the unit tests.

We wanted all the db middleware passed to the scopes method to always get applied, but it seems there are edge cases where the they are not always applied.


russellb:
Could you update the commit message and PR description with a bit more info on what happened? What did we intend, what happened instead, what's the alternative people should use?

russellb (approved):

PR #1429, opened 2023-09-19
systemd: Fix support for multiple nexd arguments

russellb:
The way args were passed to `nexd` made it so all arguments were
passed as a single argument to the process. This change makes it work
as expected.


nerdalert (approved):
LGTM

PR #1431, opened 2023-09-19
nexd: dynamic peering

### LINKED ISSUE #965: Peering failure due to invalid local connectivity assumption

### LINKED ISSUE #1322: Nested containers on EC2 instances peering issue when NACL permits the WG UDP port on both sides.
This is a scenario depicted in the diagram. This breaks when all UDP is permitted, you start seeing ICMP port unreachables. The nested containers on two separate VMs can peer via wireguard when the ACLs are not permitting the WG UDP ports. It breaks when the WG UDP ports are explicitly allowed.  Note: This does not occur when wg/nexd is running on the ec2 instance itself, only nested in containers and only when the Wireguard UDP port is explicitly opened in the secGroup. Its not a huge issue since you would rarely blindly open all UDP traffic. We can also do a workaround with `--listen-port` to lock the node down to a particular port or range that could be blocked in the secgroup if for some reason all UDP ephemeral ports needed to be opened.

I scratched my head on this for a few days and am getting diminishing returns at this point. The workaround is just to not explicitly open  the UDP port in the EC2 instance security group but I would like to get the root cause of this at some point. Its as much a wireguard thing as anything. This can be replicated if the hosts are in the same VPC or different VPCs. The common issue is it only happens when the WG UDP ports are explicitly open. 

You can replicate it without Nexodus to rule things out with just Podman or Docker on two different ec2 nodes. You don't need a STUN to open the ports since they will open with a 1:1 port mapping if there are no conflicts which is generally the case.

```
### Start two NEXD containers or a container with wireguard installed:
docker run -d --network bridge             --cap-add SYS_MODULE             --cap-add NET_ADMIN             --cap-add NET_RAW             --sysctl net.ipv4.ip_forward=1             --sysctl net.ipv6.conf.all.disable_ipv6=0             --sysctl net.ipv6.conf.all.forwarding=1             quay.io/nexodus/nexd sleep 100000

### host-1 ec2 instance (x.x.x.x is the ip of host-2) :

wg genkey | tee privatekey | wg pubkey > publickey
sudo ip link add wg0 type wireguard
sudo ip link set mtu 1420 up dev wg0
sudo ip -4 address add 10.200.0.1 dev wg0
sudo ip -4 route add 10.200.0.2/32 dev wg0

sudo wg set wg0 listen-port 33333 private-key /privatekey
sudo wg set wg0 peer "LO5qZmwcWty6Hsc0uMRNb4Q8fnpcqKJg1UQrsOGuuEk=" allowed-ips 10.200.0.2/32 persistent-keepalive 25  endpoint x.x.x.x:44444

### host-2 ec2 instance (y.y.y.y is the ip of host-2) :

wg genkey | tee privatekey | wg pubkey > publickey
sudo ip link add wg0 type wireguard
sudo ip link set mtu 1420 up dev wg0
sudo ip -4 address add 10.200.0.2 dev wg0
sudo ip -4 route add 10.200.0.1/32 dev wg0

sudo wg set wg0 listen-port 44444 private-key /privatekey
sudo wg set wg0 peer "6g83cD/Bpno8//rikITyg8ku2Bx/uPuXIbQaqpXJ5SQ=" allowed-ips 10.200.0.1/32 persistent-keepalive 25  endpoint y.y.y.y:33333
```

image

russellb:
The most common scenario addressed by this change is that after
failing to peer via a peer's local IP or reflexive IP, we can fall
back to using a relay if one is available.

More generally, these changes give a chosen peering method a chance to
work and then fall back to alternative methods. For more examples:

- An attempt to peer via local IP will fall back to peering via 
  reflexive IP

- An attempt to peer via reflexive IP will fall back to using a relay.

- If there is not currently a healthy relay available, it will fall
  back to the beginning and choose the first available method to try 

These retries will keep going until something works, or it will stay
with the relay if one is available. When using a relay, other methods
will be retried after the API sends us changes to a device that
impact peering configuration.

All of this is based on peering health inferred via wireguard stats,
not testing connectivity to each device via ICMP or something else.

Closes #965
Closes #1322

### DIFF
--- a/cmd/nexctl/connectivity.go
+++ b/cmd/nexctl/connectivity.go
@@ -47,8 +47,8 @@ func cmdConnStatus(cCtx *cli.Context, family string) error {
 
 	if err == nil {
 		w := newTabWriter()
-		fs := "%s\t%s\t%s\t%s\n"
-		fmt.Fprintf(w, fs, "HOSTNAME", "WIREGUARD ADDRESS", "LATENCY", "CONNECTION STATUS")
+		fs := "%s\t%s\t%s\t%s\t%s\n"
+		fmt.Fprintf(w, fs, "HOSTNAME", "WIREGUARD ADDRESS", "LATENCY", "CONNECTION STATUS", "PEERING METHOD")
 
 		keys := make([]string, 0, len(result))
 		for k := range result {
@@ -65,7 +65,7 @@ func cmdConnStatus(cCtx *cli.Context, family string) error {
 			if v.IsReachable {
 				status = fmt.Sprintf("%s Reachable", checkmark)
 			}
-			fmt.Fprintf(w, fs, v.Hostname, k, v.Latency, status) // Added v.Latency to the output
+			fmt.Fprintf(w, fs, v.Hostname, k, v.Latency, status, v.Method) // Added v.Latency to the output
 		}
 
 		w.Flush()
--- a/internal/nexodus/ctlconnectivty.go
+++ b/internal/nexodus/ctlconnectivty.go
@@ -50,10 +50,10 @@ func (nx *Nexodus) connectivityProbe(family string) map[string]KeepaliveStatus {
 	peerStatusMap := make(map[string]KeepaliveStatus)
 
 	if !nx.relay {
-		for _, value := range nx.deviceCache {
+		nx.deviceCacheIterRead(func(value deviceCacheEntry) {
 			// skip the node sourcing the probe
 			if nx.wireguardPubKey == value.device.PublicKey {
-				continue
+				return
 			}
 			var nodeAddr string
 			pubKey := value.device.PublicKey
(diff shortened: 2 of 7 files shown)


russellb:
It took me a while to get this working reliably in the qa-dev CI environment, but it passed there the last 6 runs in a row. I would still merge with caution and the understanding that it could cause new problems.

nerdalert (approved):
LGTM, thanks for the work here, hard fought lines of code :clap: One of us should update the peering design doc at some point.

PR #1441, opened 2023-09-20
qa-dev: dump connectivity test info for all nodes

russellb:

### DIFF
--- a/.github/workflows/qa-dev.yml
+++ b/.github/workflows/qa-dev.yml
@@ -133,8 +133,8 @@ jobs:
       - name: Mesh Connectivity Results
         run: |
           set -e
-          cat ./ops/ansible/aws/connectivity-results.txt
-          if grep -iq 'Unreachable' ./ops/ansible/aws/connectivity-results.txt || grep -iq 'Failed' ./ops/ansible/aws/connectivity-results.txt; then
+          cat ./ops/ansible/aws/*-connectivity-results.txt
+          if grep -iq 'Unreachable' ./ops/ansible/aws/*-connectivity-results.txt || grep -iq 'Failed' ./ops/ansible/aws/*-connectivity-results.txt; then
             echo "Connectivity results contain 'Unreachable or Failed' nodes, check the connectivity results and artifacts for details. Failing the job"
             exit 1
           else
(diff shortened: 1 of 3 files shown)


russellb:
@nerdalert what's the best way to test this? I put it on a branch I was testing (#1431) but qa-dev seems to run the ansible stuff from main.

nerdalert:
@russellb The workflow change would run from main but the playbooks should be off the branch. For the initial heavy lifting I will do a qa fork and vet everything there. For this lets just merge it and fix any changes on followups. The forking approach is a pita to setup since you need secrets etc.

Im guessing the connectivity test is happening on all the non-relay-only `{ neoxdusNodes }` nodes but the output isnt get cat'ted to the screen since its in the workflow. 

Oh one other thing, peers ping doesnt work from the relay node. I forgot to figure out why in case you see that. 

Less pretty but an alternative validation if you dont trust peers ping output which sometimes I dont is to just check the peers dump logs the artifact logs. Here is one from your 1431 job:

```
APVCvd3N1NL30OWlQoVCu5JIe/JmiEyvOsoQnvkW8Fg=	ALbIZkx0RCVnvPJav3JHybYaFOBC5UXSRr5XHNWzPm0=	44994	0xca6c
1JYzP5CvsTdK5gE0tlf3y3PYK9464Rlts97GFmq1P2w=	(none)	54.86.238.74:38575	100.64.0.3/32,200::3/128	1695241079	180	304	20
CdXGwKEyKMM7u4YeixFth91P5uP1rXUgd4W2sEDevWI=	(none)	54.242.27.75:45952	100.64.0.2/32,200::2/128	1695241078	212	272	20
MjwRizqTpDZHBpUNy+U9ToAdCu+4zCWC46hcaR9xqjw=	(none)	54.234.178.18:54950	100.64.0.5/32,200::5/128	1695241079	124	212	20
kr3kFCriteG7lF2jxFNzyZwyJjYUxQADn4nSgYfIw3g=	(none)	3.94.79.102:38158	100.64.0.4/32,200::4/128	1695241079	124	212	20
```

Approving and we can just fix anything on another so you get the workflow change it. That said it looks like it should work!

nerdalert (approved):

PR #1442, opened 2023-09-21
qa-dev: Fix globbing in workflow

russellb:
The failure showed that the wildcard was not interpreted. Use another
method that will hopefully expand the wildcard.

The failure:

> cat: './ops/ansible/aws/*-connectivity-results.txt': No such file or directory


nerdalert (approved):
LGTM

PR #1443, opened 2023-09-21
feat: apiserver will now track a device’s online status 

chirino:
by tracking if it has an active long poll request against the events endpoint.


nerdalert (approved):
LGTM

PR #1444, opened 2023-09-21
nexd: fixes #1426 If your machine's time is set to 15 mins into the future, nexd will fail to login

### LINKED ISSUE #1426: If your machine's time is set to 15 mins into the future, nexd will fail to login
### Describe the bug

If your machine's time is set to 15 mins into the future, nexd will fail to login.  It will get an error similar to:

```
dist/nexd --service-url https://qa.nexodus.io/
{"level":"info","ts":"2023-09-18T18:24:47.692-0400","caller":"nexd/main.go:108","msg":"Starting node agent with wireguard driver"}
Your device must be registered with Nexodus.
Your one-time code is: IOCQ-CVKT
Please open the following URL in your browser to sign in:
https://auth.qa.nexodus.io/realms/nexodus/device?user_code=IOCQ-CVKT
Authentication succeeded.
{"level":"warn","ts":"2023-09-18T18:24:58.088-0400","caller":"nexodus/nexodus.go:409","msg":"client api error - retrying: oidc: token is expired (Token Expiry: 2023-09-18 18:13:13 -0400 EDT)"}
```

### How to Reproduce

change your machines time to be 15 mins in the future.

### Expected Behavior

Nexd should eith cope with machine times that are wrong, or it should provide a better error message so the user knows that his machine time is incorrect.

### Deployment Environment

_No response_

### Additional Context

_No response_

chirino:
We now compare the local time with server time to compute if the client needs to adjust for client time skew.

### DIFF
--- a/internal/client/apiclient.go
+++ b/internal/client/apiclient.go
@@ -69,9 +69,29 @@ func NewAPIClient(ctx context.Context, addr string, authcb func(string), options
 }
 
 func createOAuthHttpClient(ctx context.Context, apiClient *public.APIClient, opts *options, authcb func(string)) (*http.Client, error) {
+	startTime := time.Now()
 	resp, _, err := apiClient.AuthApi.DeviceStart(ctx).Execute()
 	if err != nil {
 		return nil, err
+	}
+
+	// knowing the right time is needed to validate if the JWT has expired
+	serverTimeFn := time.Now
+
+	// Can we use the server's reported time to detect if our time is skewed?
+	if !resp.ServerTime.IsZero() {
+
+		// account for the time spend sending the request to the server...
+		rtt := time.Since(startTime)
+		timeSkew := resp.ServerTime.Sub(startTime.Add(rtt / 2))
+
+		timeSkewGrace := 5 * time.Second
+		if timeSkew > timeSkewGrace || timeSkew < -timeSkewGrace {
+			// adjust for the time skew...
+			serverTimeFn = func() time.Time {
+				return time.Now().Add(timeSkew)
+			}
+		}
 	}
 
 	provider, err := oidc.NewProvider(ctx, resp.Issuer)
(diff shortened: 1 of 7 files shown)


vishnoianil:
> Is there a way to simulate time skew with containers so we can create a tests case?

i think we can set time in the container? I assume that requires SYS_TIME cap and probably need to disable NTP on host.

internal/client/apiclient.go
```diff
@@ -47,18 +47,39 @@ func NewAPIClient(ctx context.Context, addr string, authcb func(string), options
...
 	}
 
+	// knowing the right time is needed to validate if the JWT has expired
+	serverTimeFn := time.Now
+
+	// Can we use the server's reported time to detect if our time is skewed?
+	if !resp.ServerTime.IsZero() {
+
+		// account for the time spend sending the request to the server...
+		rtt := time.Since(startTime)
```
vishnoianil, line 63: does it make sense to log a warning about this time skew? because there is a good possibility that the device clock can get skewed out of grace period and never reconcile ?

chirino: We don't pass a logger instance to the client, so that's why I didn't log it.  But with this PR we use a new Now function implementation that adjusts for the time skew which will allow reconciles to succeed.

chirino:
> > Is there a way to simulate time skew with containers so we can create a tests case?
> 
> i think we can set time in the container? I assume that requires SYS_TIME cap and probably need to disable NTP on host.

Does that change the time in all containers or just one container? Because for e2e tests I want my sever's container time to be different than my nexd container's time.  But they are both on the same host.

russellb:
I can't remember what the original symptom was. What did the failure look like before this change?

chirino:
nexd would fail to log into the API as it would verify the OAuth jwt and see that it had expired (due to the clock skew).

russellb (approved):

russellb:
> Is there a way to simulate time skew with containers so we can create a tests case?

you could wrap the time methods in question and have them adjust behavior based on an env var being set, like `NEXD_SIMULATE_CLOCK_SKEW=true` or something.

vishnoianil (approved):

chirino:
@russellb wrapping like that would complicate the code more than just manually testing on a machine with clock skew.  Maybe we can force the clock skew on one of the qa-dev job machines.

PR #1447, opened 2023-09-22
qa-dev: Collect mesh connectivity from all nodes

russellb:
A previous change tried to do this, but missed a spot that the
filenames needed to change to be unique per host.


nerdalert (approved):
LGTM

PR #1449, opened 2023-09-23
Add Static Favicons

nerdalert:
- Removed favicon since it doesn't appear we are using it.


russellb:
Could we just save off the favicon and drop this dependency completely?

nerdalert:
> Could we just save off the favicon and drop this dependency completely?

Well, I went digging for a while and I can't find anywhere the ico was getting rendered. The main logo render in the toolbar is handled by the `CustomAppBar` component from `logo.png` in tree.

Going to leave the commented piece in `vite.config.ts` until I make sure there aren't any residuals after a week or so. Random icons in bookmarks and things like that in case I'm missing something it is actually doing. 

Removed favicon and rolled back the build options. Also leaving `cross-env ` to pass variables at build time as needed so I dont have to figure it out again. Its no overhead, just a means of expressing ENVs to vite.

russellb:
Pretty sure a favicon gets served that’s rendered from the logo … I’ll take a look and review Monday!

nerdalert:
> Pretty sure a favicon gets served that’s rendered from the logo … I’ll take a look and review Monday!

- ty ty, Check out:
https://github.com/nexodus-io/nexodus/blob/main/ui/src/layout/AppBar.tsx#L16

- and later in the file:
https://github.com/nexodus-io/nexodus/blob/main/ui/src/layout/AppBar.tsx#L50

nerdalert:
I capitulated and did a static when I noticed an icon in the bookmark that was the default react logo that react-admin must have been generating. Now I am getting the Nex logo in the bookmark bar. Im officially out of the logo business for the weekend.

image

russellb (approved):
Thank you for sorting this out!  Dropping this dependency is very helpful, too.

PR #1463, opened 2023-09-25
Add token refresh to client side web ui

nerdalert:
- This should resolve the hard logout after the initial token expires after 300s. The logic is the OIDC token Refresh() kicks off 5sec before the exp in the keycloak token is set to expire and gets updated with a new refresh token. More details on this method at: https://marmelab.com/blog/2020/07/02/manage-your-jwt-react-admin-authentication-in-memory.html
- Tokens are still managed on the server-side so the client side is not dealing with storage. There are pros and cons for both methods.
- Safari private-mode is the only scenario that doesn't work here, but Safari is notorious for its strictness in cookies so it is likely something related.
- This handles multiple logins from separate browsers on the same host etc.

### DIFF
--- a/internal/api/public/api_auth.go
+++ b/internal/api/public/api_auth.go
@@ -216,6 +216,27 @@ func (a *AuthApiService) ClaimsExecute(r ApiClaimsRequest) (map[string]interface
 			body:  localVarBody,
 			error: localVarHTTPResponse.Status,
 		}
+		if localVarHTTPResponse.StatusCode == 401 {
+			var v string
+			err = a.client.decode(&v, localVarBody, localVarHTTPResponse.Header.Get("Content-Type"))
+			if err != nil {
+				newErr.error = err.Error()
+				return localVarReturnValue, localVarHTTPResponse, newErr
+			}
+			newErr.error = formatErrorMessage(localVarHTTPResponse.Status, &v)
+			newErr.model = v
+			return localVarReturnValue, localVarHTTPResponse, newErr
+		}
+		if localVarHTTPResponse.StatusCode == 500 {
+			var v string
+			err = a.client.decode(&v, localVarBody, localVarHTTPResponse.Header.Get("Content-Type"))
+			if err != nil {
+				newErr.error = err.Error()
+				return localVarReturnValue, localVarHTTPResponse, newErr
+			}
+			newErr.error = formatErrorMessage(localVarHTTPResponse.Status, &v)
+			newErr.model = v
+		}
 		return localVarReturnValue, localVarHTTPResponse, newErr
 	}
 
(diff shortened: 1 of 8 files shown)


nerdalert:
Putting on hold until tomorrow to let client sides run overnight to see if there are any long running issues.

internal/api/public/api_auth.go
```diff
@@ -493,30 +538,71 @@ func (a *AuthApiService) RefreshExecute(r ApiRefreshRequest) (*http.Response, er
...
+		if localVarHTTPResponse.StatusCode == 500 {
+			var v string
+			err = a.client.decode(&v, localVarBody, localVarHTTPResponse.Header.Get("Content-Type"))
+			if err != nil {
+				newErr.error = err.Error()
+				return localVarReturnValue, localVarHTTPResponse, newErr
+			}
+			newErr.error = formatErrorMessage(localVarHTTPResponse.Status, &v)
+			newErr.model = v
+		}
```
russellb, line 592:
```suggestion
		if localVarHTTPResponse.StatusCode == 400 || localVarHTTPResponse.StatusCode == 401 || localVarHTTPResponse.StatusCode == 500 {
			var v string
			err = a.client.decode(&v, localVarBody, localVarHTTPResponse.Header.Get("Content-Type"))
			if err != nil {
				newErr.error = err.Error()
				return localVarReturnValue, localVarHTTPResponse, newErr
			}
			newErr.error = formatErrorMessage(localVarHTTPResponse.Status, &v)
			newErr.model = v
			return localVarReturnValue, localVarHTTPResponse, newErr
		}
```

internal/api/public/api_auth.go
```diff
@@ -416,6 +437,27 @@ func (a *AuthApiService) LogoutExecute(r ApiLogoutRequest) (*ModelsLogoutRespons
...
```
russellb, line 460:
```suggestion
		if localVarHTTPResponse.StatusCode == 401 || localVarHTTPResponse.StatusCode == 500 {
			var v string
			err = a.client.decode(&v, localVarBody, localVarHTTPResponse.Header.Get("Content-Type"))
			if err != nil {
				newErr.error = err.Error()
				return localVarReturnValue, localVarHTTPResponse, newErr
			}
			newErr.error = formatErrorMessage(localVarHTTPResponse.Status, &v)
			newErr.model = v
			return localVarReturnValue, localVarHTTPResponse, newErr
		}
```

internal/api/public/api_auth.go
```diff
@@ -604,6 +690,27 @@ func (a *AuthApiService) UserInfoExecute(r ApiUserInfoRequest) (*ModelsUserInfoR
...
```
russellb, line 713:
```suggestion
		if localVarHTTPResponse.StatusCode == 401 || localVarHTTPResponse.StatusCode == 500 {
			var v string
			err = a.client.decode(&v, localVarBody, localVarHTTPResponse.Header.Get("Content-Type"))
			if err != nil {
				newErr.error = err.Error()
				return localVarReturnValue, localVarHTTPResponse, newErr
			}
			newErr.error = formatErrorMessage(localVarHTTPResponse.Status, &v)
			newErr.model = v
			return localVarReturnValue, localVarHTTPResponse, newErr
		}
```

russellb (changes requested):

internal/api/public/api_auth.go
nerdalert: @russellb Ahoy! Those are all openapi generated files.

russellb: @nerdalert lol sorry.

russellb (approved):

PR #1464, opened 2023-09-27
Add a health field to the device UI table

nerdalert:


nerdalert:
There is a conditional added that will show online since only if the device is online. Otherwise, the field is not present.

image

I added a basic icons-material for online/offline thats only slight prettier than the default checkmark. Also removed the distance array item from the top level devicelist until its functional but left it in the device drilldown view.

image

russellb (approved):
A tooltip that explains how the online status is determined would be a nice small improvement in the future

PR #1469, opened 2023-09-28
Aliased predefined ports in sec group rules edit UI

nerdalert:
- The predefined ports were getting passed as-is instead of being normalized to a proper protocol, e.g., MYSQL needs to be mapped to TCP to be punched into netfilter. The secgroup rules validation patch brought this to the surface. The ports do not grey out as they did, but I added a todo to do it in another patch to avoid regression here.
- Remove ALL TRAFFIC as an option and replace with tooltips
- Since we are implicitly allowing all traffic it makes more sense to give the user a message to remove all rules to allow all traffic.
- Adds a tooltip explaining adding a rule will implicitly begin dropping all other traffic that isn't explicitly allowed.
- Screencaps for all of the changes in comments.

### DIFF
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -21,7 +21,7 @@
         "web-vitals": "^3.4.0"
       },
       "devDependencies": {
-        "@mui/material": "^5.14.10",
+        "@mui/material": "^5.14.11",
         "@types/jest": "^29.5.5",
         "@types/react": "^18.2.22",
         "@types/react-dom": "^18.2.7",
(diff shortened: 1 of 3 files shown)


nerdalert:
This breaks the blurring of the port range cell in this screen cap, but makes it work again. Added a todo to go back and fix that but Im trying to avoid regression from what I have working here:

image

Now looks like this (only difference is the port field is modifiable) But it renders properly with an alias in edit:

image

In the view tab you will see the actual port number rather than the alias. That should be reasonably to resolve in another patch in SecGroupTable using the same functions from edit to alias.

image

Here is what the API payload looks like with the fix in:

```json
  "group_name": "default",
  "group_description": "security group testing",
  "inbound_rules": [
    {
      "ip_protocol": "tcp",
      "from_port": 5432,
      "to_port": 5432,
      "ip_ranges": [
        "200::/64"
      ]
    },
    {
      "ip_protocol": "tcp",
      "from_port": 3306,
      "to_port": 3306
    },
    {
      "ip_protocol": "tcp",
      "from_port": 445,
      "to_port": 445,
      "ip_ranges": []
    }
  ],
  "outbound_rules": [
    {
      "ip_protocol": "ipv4",
      "from_port": 0,
      "to_port": 0,
      "ip_ranges": [
        ""
      ]
    }
  ]
}
```

nerdalert:
Screencap of the tooltips added in the second commit.

image

image

russellb (approved):

Issue #1470, opened 2023-09-28
Security Group todos:


nerdalert:
Add SecGroup issues here please:
- [ ] `ALL ICMP` currently only allows icmpv4. Probably shoud break it into `ALL ICMPv4` and `ALL ICMPv6` respectively.
- [ ]  Address family mixing will get fairly complicated. We should consider making the protocols explicit. An example is permit that passes validation but bombs in the datapath is `add rule inet nexodus nexodus-inbound meta nfproto ipv4 ip saddr 0200::/8 udp dport 123-456 iifname wg0 counter accept" failed: exit status 1`. For early going keeping address families separated might make the most sense even though there is some support.
- [ ] The following isn't getting instantiated in the data path: `[{"from_port":123,"ip_protocol":"ipv4","ip_ranges":["100.64.0.0/10"],"to_port":456}]`. tcp works but ipv6 || ipv4 are missing. Add e2e with the fix.

PR #1472, opened 2023-09-28
Security group agent side rule fixes

nerdalert:
- broke out a v4/v6 case statement that should have been two different cases.
- Added support for a missing agent implementation of a ipv4 or ipv6 rule with an ip range and ports specified using the transport header "th" option as follows:
```json
{
  "from_port": 123,
  "ip_protocol": "ipv4",
  "ip_ranges": [
    "100.64.0.0/10"
  ],
  "to_port": 456
},
{
  "ip_protocol": "ipv6",
  "from_port": 123,
  "to_port": 456,
  "ip_ranges": [
    "0200::/8"
  ]
}
```
- tracked in #1470

### DIFF
--- a/internal/nexodus/policy_linux.go
+++ b/internal/nexodus/policy_linux.go
@@ -188,7 +188,6 @@ func (nx *Nexodus) nfPermitProtoPortAddrV4(chain string, rule public.ModelsSecur
 	} else {
 		srcOrDst = destAddr
 	}
-
 	switch rule.IpProtocol {
 	case protoIPv4:
 		// if the specified proto is ipv4 that specifies an L3 address and does not specify ports.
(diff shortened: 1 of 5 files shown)


internal/nexodus/policy_linux.go
```diff
@@ -201,6 +200,19 @@ func (nx *Nexodus) nfPermitProtoPortAddrV4(chain string, rule public.ModelsSecur
 					return err
 				}
 			}
+		} else if rule.FromPort != 0 && rule.ToPort != 0 {
```
russellb, line 203: what if one port is `0` and one isn't? That case is ignored. Does validation in the API catch that already? Even if so, it's probably worth logging an error here to indicate that something unexpected happened.

internal/nexodus/policy_linux.go
```diff
@@ -290,6 +302,19 @@ func (nx *Nexodus) nfPermitProtoPortAddrV6(chain string, rule public.ModelsSecur
 					return err
 				}
 			}
+		} else if rule.FromPort > 0 && rule.ToPort > 0 {
```
russellb, line 305: same comment as above

russellb (changes requested):
Have you looked at what test cases could be added to catch whatever was missed here?

internal/nexodus/policy_linux.go
nerdalert: Good catch, I changed the port validation in the API server so it would never reach the agent. ty.

internal/nexodus/policy_linux.go
nerdalert: Same, adjusted in the api-server, ty.

nerdalert:
> Have you looked at what test cases could be added to catch whatever was missed here?

roger, its in there, tests were why it was in draft. Ty for taking a look.

internal/nexodus/policy_linux.go
russellb, line 203: Here you use `!= 0`, but later in the patch you use `> 0`.

Ensuring it's a positive integer is probably better, but I won't hold the patch over it.

russellb (approved):

PR #1473, opened 2023-09-28
apiserver: all internal server errors are now logged

chirino:
The error response includes a trace id to find the related log message.


russellb (approved):
thank you!

PR #1481, opened 2023-09-30
Resolves the token refreshing for the web components

nerdalert:
- This modifies some token endpoints to send the tokens to the client. This is the generally recommended strategy rather than tracking through cookies on the server side. It improves multi-tab/window issues that can arise. The refresh endpoint was modified to a POST since it needs to be able to accept the refresh token from the JWT. This doesn't affect the device workflow since we don't currently refresh existing tokens through that API.
- The flow is the client receives the access token and a refresh token. Added a RefreshManager class that will send token refresh requests on a timer along with the useEffects getting fired on browser events by react-admin which also trigger CheckAuth and thus a refresh.
- On Refresh() the client sends the refresh token which extends the token grant exp time another 10m. At some point we need to add support for improved web-based auth (which wasn't the initial need) as the refresh should technically be assigning a new token as opposed to just extending the exp field. The token exchange in place exchanges a one-time code for a token as opposed to swapping an old token for a new token. Not a blocker but something to pop an issue on and address over time.
- Having the tokens on the client solves the issue the first attempt failed at of being able to determine the state while async calls are getting fired in the background by react.
- Now we can easily determine the state based on the token on the client.
- Code example from the RA maintainers that the intial custom provider was based off and this borrows from that on refresh timers https://github.com/marmelab/ra-example-oauth/blob/master/app/src/authProvider.ts
- In memory token storage is another longer term consideration but can get fairly complicated for multiple browsers running in tandem. For now, local_storage serves our needs.


ui/src/providers/AuthProvider.tsx
```diff
@@ -1,6 +1,9 @@
 import { AuthProvider, UserIdentity } from "react-admin";
+import { RefreshManager } from "./RefreshManager";
 
 const cleanup = () => {
+  RefreshManager.stopRefreshing();
+  console.log("Cleanup Called");
```
russellb, line 6: should probably be console.debug() ? no big deal though

russellb (approved):

PR #1484, opened 2023-10-02
qa-dev: Collect nexd logs later in the job

russellb:
While reviewing qa-dev results, I noticed that the nexd logs looked
incomplete. This was because the logs were collected too soon. For
example, the main nodes logs are collected before starting the relay,
the relay-only node, and running the connectivity checks. This moves
log collection to the very end.


russellb:
@nerdalert This hasn't been tested, so I'll fix any mistakes in main after I see the job run. I wanted you to take a look and see if you think it's generally correct, at least.

nerdalert (approved):

PR #1503, opened 2023-10-05
apiserver: add apis to manage registration tokens.

chirino:


chirino:
I'll work on using these new APIs from nexd in a separate PR.

russellb (approved):
I'm comfortable with the design. It looks safe to merge now since it's not used by nexd yet.

PR #1505, opened 2023-10-06
nexd: Handle invalid tokens at startup

russellb:
This patch improves the behavior of nexd at startup when it has an
invalid auth token stored. The most common time to see this is when
switching a host to use a different service endpoint, such as
switching between dev, qa, and prod.

This makes nexd clear the invalid auth token and restart the auth
process. The backoff behavior has been adjusted from constant to an
exponential backoff. This ensures the first retry happens quickly,
since in this common scenario the first will fail, but the second
attempt will succeed, so we want to get to that second try quickly. If
the failure is something else, then we want to start backing off.


chirino (approved):
LTFM

PR #1515, opened 2023-10-09
Move nexctl connection status to the last column for aligment

nerdalert:
```
HOSTNAME     WIREGUARD ADDRESS     LATENCY     PEERING METHOD     CONNECTION STATUS
ub-arm       100.64.0.5            6.83ms      direct-local       ✓ Reachable

```


russellb (approved):
lgtm

As discussed in chat, it's probably worth exploring one of the libs out there for fancy tables. Here are a couple
- http://github.com/rodaine/table
- https://github.com/jedib0t/go-pretty

PR #1520, opened 2023-10-15
Better logging for privileges notification for nexctl nexd cmds

nerdalert:
Change the logging for `nexctl nexd` privileges error from:

```
nexctl nexd status
2023/10/15 19:08:54 Failed to get nexd version: Failed to connect to nexd: dial unix nexd.sock: connect: no such file or directory
```

To:

```
nexctl nexd status
'nexctl nexd' command must be run with sudo on macOS
```


cmd/nexctl/local_unix.go
```diff
@@ -265,6 +266,13 @@ func init() {
 			},
 		},
 	})
+	// Apply privilege check only for 'nexctl nexd' subcommand
+	if len(os.Args) > 1 && os.Args[1] == "nexd" {
+		if err := hasPrivileges(); err != nil {
+			fmt.Println(err)
+			os.Exit(1)
+		}
+	}
```
russellb, line 275: something like this might be a little cleaner, by hooking into the CLI lib:

```
diff --git a/cmd/nexctl/local_unix.go b/cmd/nexctl/local_unix.go
index 6a873d06..0cae0f88 100644
--- a/cmd/nexctl/local_unix.go
+++ b/cmd/nexctl/local_unix.go
@@ -25,6 +25,13 @@ func init() {
                                Required:    false,
                        },
                },
+               Before: func(cCtx *cli.Context) error {
+                       if err := hasPrivileges(); err != nil {
+                               fmt.Println(err)
+                               return err
+                       }
+                       return nil
+               },
                Subcommands: []*cli.Command{
                        {
                                Name:   "version",
```

russellb (approved):

PR #1538, opened 2023-10-20
Add registration token cli

chirino:
This also improves the registartion token APIs a bit.


russellb (approved):
looking great! thanks for all of your work on this!

PR #1552, opened 2023-10-24
Add PacketFilter policy support for macOS

nerdalert:
- Adds v4/v6 policy support via PF for darwin.
- Creates a PF anchorfile in `/etc/pf.anchors/io.nexodus` and copies the main pf.conf for loading to avoid modifying a system file.

### DIFF
--- a/docs/user-guide/security-groups.md
+++ b/docs/user-guide/security-groups.md
@@ -2,7 +2,7 @@
 
 ## Overview
 
-Nexodus Security Groups are virtual firewalls for your Nexodus instances to control inbound and outbound traffic. They act as a white list, only allowing through the traffic that you specify is allowed. Each security group includes a set of rules that filter traffic coming into and out of the instance.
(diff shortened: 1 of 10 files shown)


nerdalert:
This is pretty much ready to go. I just need some diverse testing. For example, UTM was a major headache as QEMU loads the sharing anchor in what I am guessing via memory rather than an anchor file as I don't think there is even an API for that. It may just be that way for all `com.apple.internet-sharing` but there is virtually no documentation and what it is either unanswered or weird Apple Developer gated issues like [this](https://developer.apple.com/forums/thread/656877).

One known caveat is a PF limitation of v4/v6 support for address families with a port defined without an L4 port defined. E.g., `port only applies to tcp/udp`. This can be normalized on the agent side, either in this PR or a followup. From what I've seen so far we have parity with Linux policy once that last issue is fixed.

russellb (approved):
I'll run with this and let you know if I hit anything!

PR #1553, opened 2023-10-24
nexd: Add unit tests for peering method fallback and fix some bugs

russellb:
- nexd: Fix delay of peering fallback
- nexd: adjust rebuildPeerConfig() return values
- nexd: Expand peering unit tests
- nexd: Fix peering reset when no relays are present
- nexd: Add 3rd level peering fallback unit tests

commit 8de00491a5b700bc5da9074cafa365b9817a11d0
Date:   Mon Oct 23 16:50:16 2023 -0400

    nexd: Add 3rd level peering fallback unit tests
    
    This change introduces tests for a 3rd level of peering fallback. It
    covers cases like:
    
        direct -> reflexive -> relay
    
    or if no relay was present:
    
        direct -> reflexive -> direct
    
    It also ensures that peering doesn't change once we fall back to a
    healthy relay.
    
commit e2cebf9f618173b14268e7b847bcc84f9d4b18a7
Date:   Tue Oct 24 10:49:21 2023 -0400

    nexd: Fix peering reset when no relays are present
    
    The code that intended to make peering start over at the beginning
    would only work after connecting with a relay and then that relay goes
    away. If no relay was available, we would never go back to the
    beginning of the peering method list.
    
    This change also introduces a "none" peering method to be more
    explicit when we do not think it is currently possible to reach a
    peer via any means. Previously, "via-relay" was used for both
    "reachable by a relay" and "reachable IF a relay was present".
    
commit 8e8711df7f04c54441f7fde52ef47b978e303f12
Date:   Mon Oct 23 13:15:08 2023 -0400

    nexd: Expand peering unit tests
    
    Expand the peering method selection unit tests to cover the first
    level of fallback. It ensures that enough time is given before the
    fallback and that the expected second choice occurs.
    
commit a707e681be56db5cb4a3314fbb1eacf6a818cb56
Date:   Mon Oct 23 13:35:58 2023 -0400

    nexd: adjust rebuildPeerConfig() return values
    
    When no new method is chosen, just return the current settings instead
    of blank. This doesn't change any behavior, but makes it a bit easier
    to write some tests.
    
commit 5b0c4c2b1b5f827f69829381319001a9d28528e3
Date:   Mon Oct 23 13:13:09 2023 -0400

    nexd: Fix delay of peering fallback
    
    The code intended to delay dropping back to the next preferred peering
    method if a given method was previously working and then stopped. The
    intent was to wait 3 minutes in that case, instead of waiting much
    less time for a method that had not yet worked at all.
    
    The code was checking the wrong timestamp for this. The problem was
    discovered while writing unit tests for this code.
    
### DIFF
--- a/internal/nexodus/nexodus.go
+++ b/internal/nexodus/nexodus.go
@@ -82,8 +82,6 @@ type userspaceWG struct {
 const ()
 
 type peerHealth struct {
-	// the time we started tracking this data
-	startTime time.Time
 	// the last tx bytes value for this peer
 	lastTxBytes int64
 	// the time of the last tx bytes update
@@ -1010,7 +1008,7 @@ func (nx *Nexodus) reconcileDeviceCache() error {
 		existing.endpoint = curStats.Endpoint
 		existing.peerHealthy = nx.peerIsHealthy(existing)
 		if existing.peerHealthy {
-			existing.startTime = now
+			existing.peerHealthyTime = now
 		}
 		nx.deviceCache[p.PublicKey] = existing
 	}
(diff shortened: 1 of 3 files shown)


nerdalert (approved):
LGTM

PR #1555, opened 2023-10-28
Add qa-dev to promote for promotion validation

### LINKED ISSUE #1554: Add qa-dev to the promote workflow
We need basic testing prior to promote. qa-dev does a real world deployment on EC2 with relay-only and full mesh peers which makes it a reasonable gating job to start with.
https://github.com/nexodus-io/nexodus/actions/workflows/qa-dev.yml

nerdalert:
Closes #1554

GH doesn't return a job ID from a dispatch so this timestamps and queries qa-dev for runs after the timestamp. This should work in all scenarios since qa-dev is in a concurrency group and shouldn't conflict with other dispatches. More on the API limitation [here](https://github.com/orgs/community/discussions/9752).

I validated the API calls in a fork and they are functional. Going to draft this until Monday when we promote next.

### DIFF
--- a/.github/workflows/promote.yml
+++ b/.github/workflows/promote.yml
@@ -7,6 +7,62 @@ on:
   workflow_dispatch:
 
 jobs:
+  run-qa-dev:
+    # Dispatch the qa-dev workflow to validate main prior to promotion
+    runs-on: ubuntu-latest
+
+    env:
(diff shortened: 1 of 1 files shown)


russellb (approved):
Cool!

After this we should be able to auto trigger promote after successfully updating QA

PR #1556, opened 2023-10-30
Exit node fixes

nerdalert:
- Fixes an issue where exit nodes were being duplicated in `exitNodeOrigins` if the endpoint address changed. It now keys off the public key to avoid duplication.
```
# Resolves this:
nexctl nexd exit-node list
ENDPOINT ADDRESS        PUBLIC KEY
192.168.64.2:56997      Wneb+xx6481iaWB2eohZmEttZzar+G7R6A0hoBYkJBQ=
192.168.64.2:55374      Wneb+xx6481iaWB2eohZmEttZzar+G7R6A0hoBYkJBQ=
```
- Adjust nexctl to display the local node if it is an exit node. Currently, if the local node is an exit node it is omitted from the nexctl output of `nexctl nexd exit-node list`. It only adjusts the nexctl output since it is operating as intended since it is using peer data from the api-server.

### DIFF
--- a/internal/nexodus/exit_origin.go
+++ b/internal/nexodus/exit_origin.go
@@ -61,3 +61,21 @@ func addExitOriginForwardRule(logger *zap.SugaredLogger) error {
 
 	return nil
 }
+
+func (nx *Nexodus) updateExitNodeOrigins(newPeer wgPeerConfig) {
+	nx.exitNode.exitNodeExists = true
+	// check if the exit node already exists and update its details
+	found := false
+	for i, existingPeerConfig := range nx.exitNode.exitNodeOrigins {
+		if existingPeerConfig.PublicKey == newPeer.PublicKey {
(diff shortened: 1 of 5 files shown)


internal/nexodus/exit_origin.go
```diff
@@ -61,3 +61,21 @@ func addExitOriginForwardRule(logger *zap.SugaredLogger) error {
...
+			found = true
+			break
+		}
+	}
+
+	// If no existing entry with the same public key was found, append the new exit node
+	if !found {
+		nx.exitNode.exitNodeOrigins = append(nx.exitNode.exitNodeOrigins, newPeer)
+	}
+}
```
russellb, line 81: Using a map instead of a list would simplify this a good bit.

russellb (approved):
left one future suggestion inline

PR #1568, opened 2023-11-01
Update relay status if device rejoins as a relay

### LINKED ISSUE #367: Write unit test to cover the scenario reported in issue #183
#183

### LINKED ISSUE #1566: Rejoining device as a relay, doesn't mark it relay.
### Describe the bug

If you connect the device as a normal device, but if you reconnect the same device as a relay, it doesn't update it's relay flag.

### How to Reproduce

Connect nexd. Disconnect and reconnect it using `nexd relay` , and check the web-ui (or debug log), it won't update the relay flag.

### Expected Behavior

If the device rejoin as a Relay, it should be marked as a Relay. If we don't want to allow any device to convert to relay, we should deny onboarding of the device if it comes back as a relay.

### Deployment Environment

_No response_

### Additional Context

_No response_

vishnoianil:
Fixes #1566 
Fixes #367

### DIFF
--- a/internal/models/device.go
+++ b/internal/models/device.go
@@ -54,5 +54,6 @@ type UpdateDevice struct {
 	Hostname        string     `json:"hostname" example:"myhost"`
 	Endpoints       []Endpoint `json:"endpoints" gorm:"type:JSONB; serializer:json"`
 	Revision        *uint64    `json:"revision"`
+	Relay           *bool      `json:"relay"`
 	SecurityGroupId *uuid.UUID `json:"security_group_id"`
 }
--- a/integration-tests/features/device-api.feature
+++ b/integration-tests/features/device-api.feature
@@ -153,7 +153,56 @@ Feature: Device API
       }
       """
 
-    # Bob gets an should see 1 device in the device listing..
+    # Bob can update if device become relay.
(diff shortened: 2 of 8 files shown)


russellb:
Can you add test coverage? It should be a pretty small addition to `integration-tests/features/device-api.feature`

internal/models/device.go
```diff
@@ -54,4 +54,5 @@ type UpdateDevice struct {
 	Hostname       string     `json:"hostname" example:"myhost"`
 	Endpoints      []Endpoint `json:"endpoints" gorm:"type:JSONB; serializer:json"`
 	Revision       *uint64    `json:"revision"`
+	Relay          bool       `json:"relay"`
```
chirino, line 57: Can we make this a *bool so that updating the relay status is optional?

chirino (changes requested):

vishnoianil:
> Can you add test coverage? It should be a pretty small addition to `integration-tests/features/device-api.feature`

Done

internal/models/device.go
vishnoianil: Done

russellb (approved):

chirino (approved):
Lgtm

PR #1569, opened 2023-11-01
e2e: Add more flexibility to the proxy LB test

russellb:
I've seen cases where this test fails because the counters are not 5
and 5. This was the last one:

    proxy_test.go:906: TCP responses: apples=4, bananas=6
    proxy_test.go:907: UDP responses: carrot=5, potato=5

I can see that curl failed once and had to retry. In that case, we'll
hit the other endpoint.

Make the test more flexible for when this happens. Make sure we hit
each endpoint at least 3 times and that the sum is still 10.


chirino (approved):
Lgtm

PR #1571, opened 2023-11-01
nexd: Add child prefix routing through relays

### LINKED ISSUE #1536: Nexd relay doesn't advertise child prefixes to its peers
Here is an example dump from a relay. I think adding the child-prefix to the peer will enable reachability from spoke relay-only nodes:

```
$ sudo wg show wg0 dump
eJHkc5w+wPKuWACwK3MfQ0qT/tgRg4p/lFKiJEl5zFk=	IAIMKPbWNCs9GDnH03RnmwkBdObcgpvdimY8+QOWTQM=	50336	0xca6c
4kFJVTNiDZ2opwm+JC0TfHUDV4dH0D6EM9dyVFl8EHk=	(none)	3.85.122.146:51820	100.64.0.0/10,200::/64	1697648726	2308	2476	20
```

I verified that all child-prefixes need to be added to each peers to enable forwarding through the relay. An ex:

```
wg set wg0 peer "4kFJVTNiDZ2opwm+JC0TfHUDV4dH0D6EM9dyVFl8EHk=" allowed-ips 100.64.0.0/10,200::/64,172.24.100.0/24 persistent-keepalive 25  endpoint 3.85.122.146:51820
```

russellb:
Prior to this change, the only way to reach a child-prefix CIDR was by
peering directly with the device exposing that prefix.

This change updates relay peer configuration to include child-prefix
CIDRs for peers that are reachable via the relay instead of directly.

Closes #1536

### DIFF
--- a/integration-tests/integration_test.go
+++ b/integration-tests/integration_test.go
@@ -409,8 +409,8 @@ func TestHubVPC(t *testing.T) {
 	err := helper.nexdStatus(ctx, node1)
 	require.NoError(err)
 
-	helper.runNexd(ctx, node2, "--username", username, "--password", password, "--vpc-id", vpcID)
-	helper.runNexd(ctx, node3, "--username", username, "--password", password, "--vpc-id", vpcID)
+	helper.runNexd(ctx, node2, "--username", username, "--password", password, "--vpc-id", vpcID, "--relay-only")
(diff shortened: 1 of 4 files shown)


russellb:
This could use some additional testing. My only testing so far is via the unit tests included in the PR.

russellb:
Marked as a draft as e2e shows that it's broken.

russellb:
tests are passing now!

nerdalert (approved):
LGTM

PR #1577, opened 2023-11-02
nexd: Remove --stun option

### LINKED ISSUE #1574: Combining --stun with --exit-node produces errors on linux
### Describe the bug

if on linux: 

`NEXD_ARGS="--service-url https://try.nexodus.io router --exit-node"`
It appears to succeed

if
`NEXD_ARGS="--stun --service-url https://try.nexodus.io router --exit-node"`

the following error line is produced
` {"level":"fatal","ts":"2023-11-02T05:12:56.321Z","caller":"nexd/main.go:202","msg":"failed to setup this device as an exit-node: failed to add nftables rule nexodus-exit-node: nft command: nft \"add rule inet nexodus-exit-node postrouting oifname  masquerade\" failed: exit status 1\n"}`

### How to Reproduce

Install nexd on centos stream 9 with dnf - note `nexd` on centos stream 9 should specify `nftables` as a dependency.

alter the `NEXD_ARGS` in `/etc/sysconfig/nexodus` and restart the service.

### Expected Behavior

EIther:
1. A logical error message.
2. Something not to 

### Deployment Environment

Machine is a centos9 streams vm on IBM Cloud with a 2 core 8GB footprint.

system is deployed with a permissive security group. 

### Additional Context

_No response_

russellb:
This option appears to be left over from a time when we only kept
track of one endpoint address per device. We would either just use a
local address or we would use one from stun. We now intentionally keep
track of both in the Endpoints list.

The code for `--stun` now did something that didn't make any sense. If
`--stun` is used, it stores the reflexive IP from stun in the "local"
endpoint as well as the stun endpoint.

As far as I can tell, there is no longer any value to this option. We
are always using stun and this option didn't turn that on or off.

By storing the reflexive IP from stun as our local address, it broke
exit node when used with the `--stun` option.

Closes #1574


nerdalert (approved):
LGTM

PR #1583, opened 2023-11-05
ui: fix security groups after API changes

russellb:
The paths for the security-groups API changed in our recent round of
API updates. This change adjusts the security-groups page ti use the
new APIs.


chirino (approved):
Lgtm

PR #1586, opened 2023-11-05
apiserver: implement a per user concurrency limiter

chirino:


internal/routers/middleware.go
```diff
@@ -153,25 +162,51 @@ func ValidateJWT(ctx context.Context, o APIRouterOptions, jwksURI string, nexodu
...
+		}()
+
+		prefixId := fmt.Sprintf("%s:%s", handlers.CachePrefix, idpUserID)
+		cachedUserId := ""
 
-		if cachedUserId == "" {
-			userId, err := o.Api.CreateUserIfNotExists(c.Request.Context(), idpUserID, idpUserName)
+		// for now just use the concurrency limiter to serialize the cache lookup and create user if not exists
+		// in the future we can use the concurrency limiter to limit the number of concurrent requests to  other
+		//  requests in the apiserver.  We may need different concurrency levels for things like device requests.
```
russellb, line 185: It may be worth noting here that this is enforced on a per-process basis. It doesn't prevent concurrent requests across replicas. This looks good to help prevent some DoS issues, but should not be relied upon for system-wide serialization.

russellb:
```suggestion
		//  requests in the apiserver.  We may need different concurrency levels for things like device requests.
		// NOTE: This does not prevent concurrent access across the system. It is only enforced on a per
		// apiserver process basis. With apiserver replicas in place, concurrent access in this code path can
		// still occur. The change here will still limit the number of db connections from here at a time, but they
		// will not necessarily be serialized.
```

russellb (approved):

Issue #1582, opened 2023-11-04
Don't allow deleting the default org and vpc.


chirino:
The default organization, vpc, and security groups should always exist. They should only be deleted when their owner is deleted.

chirino:
implemented.

PR #1588, opened 2023-11-05
apiserver: add tests and fix bugs related to the new vpc resource

chirino:
Implements #1582 for vpcs
Deleting vpcs was not working at all
We now prohibit deleting vpcs if devices still exist for it. Add support for updating the vpc description.


nerdalert (approved):
LGTM

PR #1589, opened 2023-11-05
apiserver: add tests and #1582 for security groups.

chirino:
Implements #1582 for security groups.
We now sort by the group name instead of by id.


chirino:
Need to think about this one a bit... TestSecurityGroupsExtended expects to be able to delete the default security group.

PR #1604, opened 2023-11-07
docs: Use paths for root user in help text

russellb:
When displaying the default unix socket path, use the privileged mode
default, as that is the common case.


nerdalert:
We could capture both Darwin and Linux with `/var/run/nexd.sock`. Both symlink to different directories, linux. `/run/nexd` and OSX `/private/var/run/nexd.sock`. No big deal, just throwing it out there!

chirino (approved):
LGTM

PR #1606, opened 2023-11-07
chore: Use ShouldBindJSON everywhere

dave-tucker:
BindJSON is infallible. It sets the status to 400 and aborts. We need ShouldBindJSON since in all cases we are doing our own error handling.


dave-tucker:
Closing in favour of #1607

PR #1615, opened 2023-11-08
nexctl: Fix device list --full

russellb:
This option went boom because of fields that no longer exist.


nerdalert (approved):
LGTM, ty, I'll add tests for it in e2e. Thanks!

PR #1616, opened 2023-11-08
nexd: Handle when sec group Id changes on Device

russellb:
Handle the case when the security group Id is changed for the local
device. Prior to this change, it would be ignored until another field
changed on the Device. This was because the deviceUpdated() method did
not check the security group Id.

Some additional handling was required to ensure that the security
group update caused an immediate reconcile of the security group
settings, since that happens outside of the device reconcillation
method.

In passing, take deviceUpdated() off of the Nexodus struct, as it
didn't use anything from there.


chirino (approved):
lgtm

PR #1617, opened 2023-11-08
apiserver: don’t allow deleting default security groups anymore.  

chirino:
Simplifies the data model a bit.


internal/handlers/device.go
```diff
@@ -578,7 +578,7 @@ func (api *API) CreateDevice(c *gin.Context) {
 			SymmetricNat:    request.SymmetricNat,
 			Hostname:        request.Hostname,
 			Os:              request.Os,
-			SecurityGroupId: vpc.SecurityGroupId,
+			SecurityGroupId: vpc.ID,
```
russellb, line 581: Since we support changing the security group on a device, we should also support creating it with an Id other than the default.  Your patch doesn't create any new problems here. It just made me notice it.

russellb (approved):

PR #1620, opened 2023-11-09
feat: add an initial spike of playwright UI tests

chirino:
Lets see if we can figure out a nice workflow to use these for e2e  and qa testing.


nerdalert (approved):
LGTM

PR #1630, opened 2023-11-10
Support setting security-group-id on nexd device create, and reg-key create fixes #1622

### LINKED ISSUE #1622: Support creating a device with a different sec group Id
Right now, the create device API call does not allow you to create the device with a custom security group. It will automatically set the default security group for the VPC the device is going into. We should add this as an optional field to the create request and also add it as an option to `nexd`.

chirino:
* Improve the ui to allow editing of devices and reg keys.
* Make all the Nexodus create args part of an Options struct.


chirino:
fixes #1622

russellb:
> fixes #1622

if you put this in the PR description / commit message, it will get closed automatically

chirino:
> > fixes #1622
> 
> if you put this in the PR description / commit message, it will get closed automatically

thanks or the tip!

internal/handlers/reg_key.go
```diff
@@ -48,49 +49,157 @@ func (api *API) CreateRegKey(c *gin.Context) {
...
 	}
 
-	// The vpc has to be owned by the user.
-	var vpc models.VPC
-	db := api.db.WithContext(ctx)
-	if res := api.VPCIsOwnedByCurrentUser(c, db).
-		First(&vpc, "id = ?", request.VpcID.String()); res.Error != nil {
-		c.JSON(http.StatusNotFound, models.NewNotFoundError("vpc"))
+	// use a wg private key as the token, since it should be hard to guess.
+	token, err := wgtypes.GeneratePrivateKey()
```
russellb, line 53: Unrelated to this PR, but I wonder if this is more expensive to generate than a UUID.

russellb (commented):
one small comment inline, but it's unrelated to this change

russellb (approved):

internal/handlers/reg_key.go
chirino: It likely is, but I would assume it's more random too which is highly desired.

chirino: looked it up: 
> Randomly generated UUIDs have 122 random bits.  
so that's about 16 bytes vs 32 bytes for wgkeys.

PR #1632, opened 2023-11-10
nexd: Don't assign tunnel IP if it conflicts with host

### LINKED ISSUE #24: Check for overlap of CIDR on the system net interfaces
When a user runs the agent to join a mesh, ensure there is not overlap with existing IP on the agent machine interfaces for OSX/Linux (maybe routing tables in the case of child prefix Linux only).

russellb:
If either the v4 or v6 tunnel IPs conflict with a CIDR found on an
existing host interface.

I also moved setupInterface() to wg_deploy.go because that was the
only place it was used and it seems like an appropriate enough place
for it.

Closes #24


nerdalert (approved):
LGTM

PR #1634, opened 2023-11-11
Check for existing nexd process on macOS

nerdalert:
- Adjusts the nexd.sock path to be the common symlink for linux and macOS


chirino (approved):
lgtm

PR #1637, opened 2023-11-11
feat: initial spike of importing coredns for dns server capabilities.

chirino:


nerdalert (approved):
LGTM, double checked the nexd binary and it's still right at 20mb with the added deps 🎉

PR #1638, opened 2023-11-11
fixes #1514 - When a device is deleted it is marked deleted but still  has a unique key in the device table.

### LINKED ISSUE #1514: When a device is deleted it is marked deleted but still has a unique key in the device table
### Describe the bug

- After deleting a device,  rejoining will result in a 500 due to unique key constraints unless the device comes with a new pub_key

### How to Reproduce

Delete a device with the ui or with nexctl and then rejoin.

### Expected Behavior

-

### Deployment Environment

_No response_

### Additional Context

The work around is to remove the nexd cache and have a new public_key be generated.

`sudo rm /var/lib/nexd/state.json`

Some potential options include:
- deleting the record if it has a soft delete status `deleted_at: xxxx` when a device attempts to reuse the key, 
- returning a more informative error that lets the agent know to regenerate a key.
- Not a fix for the particular issue but periodically purging the soft deletes makes sense as well.

chirino:
We now null out unique fields of the device record when we delete it.

### DIFF
--- a/integration-tests/features/device-api.feature
+++ b/integration-tests/features/device-api.feature
@@ -259,6 +259,71 @@ Feature: Device API
         "id": "${device_id}",
         "vpc_id": "${vpc_id}",
         "os": "linux",
+        "public_key": "",
+        "relay": false,
+        "revision": ${response.revision},
+        "symmetric_nat": true,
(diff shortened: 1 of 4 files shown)


nerdalert (approved):
LGTM, ty!

PR #1640, opened 2023-11-12
cleanup: simplify feature flag checks.

chirino:
Reduces code duplication.


russellb (approved):
Thanks!

We can probably drop both of these flags and remove even more code. I don't think either flag is useful anymore.

PR #1658, opened 2023-11-15
apiserver: introduce a new private api to fully delete soft deleted records

chirino:
* we now generate an API doc for private apis.
* enable ready/live kube probes for the apiserver

To use, run curl in the apiserver like so:

    kubectl -n nexodus exec -it $(kubectl -n nexodus get pods -l 'app.kubernetes.io/name=apiserver' -o name) -- curl 'http://127.0.0.1:8080/private/gc?retention=24h’


nerdalert (approved):
LGTM

PR #1679, opened 2023-11-20
nexctl: clean up api access commands to be simpler and more consistent.

chirino:
Also:
* don’t show output using log, since the timestamps are not needed for interactive commands.
* we more consistently validate all uuids before doing api calls.


nerdalert (approved):
LGTM

PR #1710, opened 2023-11-29
ui: Color the online status icon

russellb:
When a device is online, make the checkmark green. A black checkmark
wasn't super clear if that meant it was good or not.

image
image
image
image


russellb:
Online:




Offline:

PR #1716, opened 2023-12-01
Add accept invitations to the UI

nerdalert:
- Adds the custom fetch handler since it's beyond what react-admin can handle.

### DIFF
--- a/internal/api/public/model_models_user_info_response.go
+++ b/internal/api/public/model_models_user_info_response.go
@@ -12,6 +12,8 @@ package public
 
 // ModelsUserInfoResponse struct for ModelsUserInfoResponse
 type ModelsUserInfoResponse struct {
+	Email             string `json:"email,omitempty"`
+	EmailVerified     bool   `json:"email_verified,omitempty"`
 	FamilyName        string `json:"family_name,omitempty"`
 	GivenName         string `json:"given_name,omitempty"`
 	Picture           string `json:"picture,omitempty"`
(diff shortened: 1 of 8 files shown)


nerdalert:
I'm getting 404's from the accept. Need to add api-server logging to narrow things down in `apiGroup.POST("/invitations/:id/accept", api.AcceptInvitation)`:


```
POST https://api.try.nexodus.127.0.0.1.nip.io/api/invitations/d46d1c84-1d64-46ec-896f-faadcafcc2d7/accept 404 (Not Found)

index-9b4c174a.js:273 Error accepting invitation: Could not fetch https://api.try.nexodus.127.0.0.1.nip.io/api/invitations/d46d1c84-1d64-46ec-896f-faadcafcc2d7/accept, status: 404
```

image

chirino:
The invitation UserId get set under a few conditions:
1) When you add the invitation, you can provide the UserID: https://github.com/nexodus-io/nexodus/blob/main/internal/handlers/invitations.go#L69
2) If the user exists in the system and has the email associated with him, then we set the UserID: https://github.com/nexodus-io/nexodus/blob/main/internal/handlers/invitations.go#L89C17-L89C17
3) If an invitation for the email address exists when a user is first created with the associated email: https://github.com/nexodus-io/nexodus/blob/main/internal/handlers/user.go#L112

Point 3 requires that the Oauth JWT claims contains that user email.  This should be the case if you login using google auth.  It's not the case for manually created users against Keycloak.

How are you testing? With google auth?

PR #1717, opened 2023-12-02
feat: add support for sending emails from the apiserver

chirino:
The emails support plain text and html messages along with inline images and normal attachments.


nerdalert (approved):
LGTM

PR #1729, opened 2023-12-05
apiserver: send an invitation email when an invitation is created.

chirino:
Email sending is only enabled in the qa env for now to test it more.


chirino:
@russellb our deployment docs just explain our current deployment offerings, dev, qa, prod. etc.
Are you asking for docs for how to enable the email notification system on those, or are you asking for how to deploy to a custom environment?

chirino:
Enabling int the other envs is going to be really easy, it's just a matter of if the `kustomization.yaml` for the env is loading the `- sealed-secret-smtp.yaml` or not.

russellb:
ok, yeah, i don't see an obvious place to extend.  I was thinking about this doc: https://docs.nexodus.io/deployment/nexodus-service/

but it really doesn't cover very much other than how to run the kind env.

chirino:
@russellb added a little blurb there... I think it should be enough to get folks going in the right direction.

PR #1747, opened 2023-12-14
Correctly seal the secrets for prod/qa namespaces

vishnoianil:


vishnoianil:
YOLOing this to enable argo-cd auto-sync.

PR #1855, opened 2024-01-24
nexctl: Don't use spinner for parseable output

russellb:
The output of the `nexctl nexd peers ping` command outputs a friendly
spinner while it runs the ping check with all of the peers. This
change turns the spinner off if the output format in use is one of the
parseable formats.


nerdalert (approved):
LGTM

PR #1857, opened 2024-01-25
docs: Fix a rendering error

russellb:
The text was not interpreted as a comment when written this way.
Dropping the `1.` makes it render as intended.


nerdalert (approved):
lgtm

PR #1907, opened 2024-02-12
Remove vscode devcontainer

russellb:
This feature is not actively used and has caused some confusion with
new contributors trying to run the full test stack in the dev
container, which we don't expect to work.


chirino:
I think it should be removed from the https://docs.nexodus.io/development/tooling/ doc too

russellb:
> I think it should be removed from the https://docs.nexodus.io/development/tooling/ doc too

Good catch!  Removed. I searched for "devcontainer" but not "dev-container" !

PR #1916, opened 2024-02-15
frontend: Redirect to the authorization server from the apiserver

chirino:
This simplifies the apiserver logic, and improves security since the Web browser never get access to the access/refresh tokens.

* No longer need to configure the redirect URL on the apiserver
* the /web/login/start and /web/login/end endpoints are now simpler  GET operations that use redirects.  
* /web/refresh now does NOT send the tokens back to the frontend
* Keycloak is configured to redirect to the apiserver instead of the front end.


chirino:
@chirino I've not verified it, but this simplification might fix the double login issue.

nerdalert (approved):
LGTM!!

PR #1919, opened 2024-02-17
Change same site to none in /login/start

nerdalert:
On QA and Prod when seeing 400s returned in the initial login redirect this might allow the login flow.

### DIFF
--- a/pkg/oidcagent/handlers.go
+++ b/pkg/oidcagent/handlers.go
@@ -94,11 +94,12 @@ func (o *OidcAgent) LoginStart(c *gin.Context) {
 		"redirect", query.Redirect,
 	)
 
-	c.SetSameSite(http.SameSiteStrictMode)
-	c.SetCookie("redirect", query.Redirect, int(time.Hour.Seconds()), "/", "", c.Request.URL.Scheme == "https", true)
-	c.SetCookie("failure", query.Redirect, int(time.Hour.Seconds()), "/", "", c.Request.URL.Scheme == "https", true)
-	c.SetCookie("state", state, int(time.Hour.Seconds()), "/", "", c.Request.URL.Scheme == "https", true)
-	c.SetCookie("nonce", nonce, int(time.Hour.Seconds()), "/", "", c.Request.URL.Scheme == "https", true)
+	c.SetSameSite(http.SameSiteNoneMode)
+	c.SetCookie("redirect", query.Redirect, int(time.Hour.Seconds()), "/", "", true, true)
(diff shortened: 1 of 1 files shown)


nerdalert:
@chirino just throwing this out to see what you think. I don't see it on a KIND deploy since I'm using a local account that never call google OAUTH as opposed to prod which will call a third party site. Loosening SameSite is the only thing I can think of there. Fact that you don't see it makes me wonder if it's local to me but Im seeing that 400 so there is something up. We can get others to see what they get. This change doesn't break the KIND env for me but still leaving in draft as I'm reaching just based on the third-party cookie msg below:
image

Here is the 400 log from prod. Doesn't really say much but I'm guessing its from something like `redirectURL, err := c.Cookie("redirect")` in login/end:

```
{"level":"info","ts":1708144272.9217975,"caller":"zap@v0.2.0/zap.go:91","msg":"/web/login/end","status":400,"method":"GET","path":"/web/login/end","query":"state=OL4GIOOv8PFzOitXceooSw&session_state=81733717-2843-4c74-be66-4221ed23f3b7&code=f5d54079-62f1-4355-967c-bd3d87ff1de3.81733717-2843-4c74-be66-4221ed23f3b7.cf93f410-e964-42aa-a244-795280fea76a","ip":"10.131.15.245","user-agent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/121.0.0.0 Safari/537.36","latency":0.002112183,"time":"2024-02-17T04:31:12Z","traceID":"f387fe18276be4446c74fd5d33ae2769"}
```

chirino (approved):
Lgtm.  Weird that I was not seeing those fail.  Is my browser not as secure as yours?

chirino:
Thinking about it. The redirect from keycloak to the apiserver should be considered the a same site request since they share the same tld+1 and scheme. The only cross site action happening is between keycloak and google.

chirino:
Have you tried disabling all your chrome plugins?

chirino:
here's my Chrome network tab results.  I enabled preserve logs, and you can I don't get those errors:

image

nerdalert:
@chirino good call on extensions. Still no joy. I did try on a Windows VM. I got the bad redirect with a 400 on the first login, and then every time after it works ¯\_(ツ)_/¯. Can replicate that with incognito.

The same site was a hail mary based on the redirect to `Navigated to https://accounts.google.com/o/oauth2/v2/auth?scope=openid+profile+email`.

chirino:
incorporated into https://github.com/nexodus-io/nexodus/pull/1924

PR #1924, opened 2024-02-19
frontend: redirect back to original window URL after login.

chirino:
This should help users navigate to an org invitation from the  invitation email link.


nerdalert (approved):
LGTM

PR #1927, opened 2024-02-20
Dependency bumps

nerdalert:


nerdalert:
@chirino the redis changes here are from a breakage upstream. It is getting set as a default option in future releases but this just goes ahead and sets it `DisableIndentity: true,`. Patch just merged upstream: `https://github.com/redis/go-redis/pull/2915`

chirino (approved):
lgtm

PR #1939, opened 2024-02-22
apiserver: support multiple Organization owners

chirino:
* added a roles []string field to user_organizations table, can be set to owner|member, allows for additional roles in the future.
* remove the organization owner_id field, as we can find owners using the roles field.


nerdalert (approved):
LGTM

PR #1940, opened 2024-02-24
fix frontend: invitation create org pick list was empty.

chirino:
* reduce code duplication in handlers
* show roles in org member listing


nerdalert (approved):
LGTM

PR #1945, opened 2024-02-27
Increase the timeout in DNS unit tests

nerdalert:
- Noticed some failures on this test, bumping the timeout from the default 2s to 5s see if it resolves the flakes.


vishnoianil:
LGTM

chirino (approved):
lgtm

PR #1947, opened 2024-02-28
apiserver: fix reg-keys api to use new org ownership roles

chirino:
* allow org members to create RegKeys that uses resources in those orgs
* Added API tests for the above.


nerdalert (approved):
LGTM

PR #1953, opened 2024-03-05
nexctl: add symmetric nat to full device list output

russellb:


nerdalert (approved):
LGTM

Issue #1956, opened 2024-03-05
SymmetricNAT is never set back to false for a device


russellb:
Once a device sets symmetric_nat to true in its Device record, it will never get set back to false once network conditions change again. I hit this for the Device record for my laptop, which gets moved around between different networks regularly.

To reproduce:

1. connect with nexd normally on a network without symmetric NAT
2. restart nexd with `--relay-only` to force the record to be updated with symmetric_nat: true
3. restart nexd again without `--relay-only`
4. Observe that the Device record still has symmetric_nat set to true. It should be false at this point.

PR #1957, opened 2024-03-06
Ensure device symmetricNAT field is updated

vishnoianil:
with add/update device requests

Openapi client generates the model with json tag
omitempty, that removes the symmetricNAT flag in
the json request body if it's false.

Fixes: #1956

### DIFF
--- a/integration-tests/features/device-api.feature
+++ b/integration-tests/features/device-api.feature
@@ -202,6 +202,55 @@ Feature: Device API
       }
       """
 
+    # Bob can update if device is no more behind symmetric NAT.
+    When I PATCH path "/api/devices/${device_id}" with json body:
+      """
+      {
+        "symmetric_nat": false
+      }
+      """
+    Then the response code should be 200
+    And the response should match json:
+      """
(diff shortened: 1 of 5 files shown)


chirino:
I don't think this will fix the problem.  The issue is in the generated OpenAPI client models.

integration-tests/features/device-api.feature
```diff
@@ -202,6 +202,55 @@ Feature: Device API
       }
       """
 
+    # Bob can update if device is no more behind symmetric NAT.
+    When I PATCH path "/api/devices/${device_id}" with json body:
+      """
+      {
+        "symmetric_nat": false
```
russellb, line 209: In my testing, this was not present in the request at all

chirino: Yeah.. this PR changes AddDevice model, but that's not used in the PATCH operation.

chirino:
fixed in #[1958](https://github.com/nexodus-io/nexodus/pull/1958)`

PR #1960, opened 2024-03-06
apiserver/nexd: improve the meta data event apis so that we can use

chirino:
An informer from nexd to access its data.

* Added the ability to create an informer for a Get operation
* Added a util.JsonStringer(x) that you can use for deferred json formatting.


nerdalert (approved):
LGTM

PR #1979, opened 2024-03-14
Fixes e2e failures on mac

chirino:
* e2e: avoid using os.env to configure test options as that can cause a data race in concurrent tests.
* add make e2e-setup target that gets you setup to run e2e tests.
* don’t allow direct local connections when the `--peer-only` cli option is used.
* fx data races in derp proxy restart

### DIFF
--- a/internal/nexodus/wg_peers.go
+++ b/internal/nexodus/wg_peers.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"github.com/nexodus-io/nexodus/internal/client"
 	"net"
+	"net/netip"
 	"reflect"
 	"runtime"
 	"strconv"
@@ -75,7 +76,7 @@ var wgPeerMethods = []wgPeerMethod{
 		// We are behind the same reflexive address as the peer, try direct, local peering
 		name: peeringMethodDirectLocal,
 		checkPrereqs: func(nx *Nexodus, device client.ModelsDevice, reflexiveIP4 string, healthyRelay bool, _ bool) bool {
-			return !nx.relay && !device.GetRelay() && nx.nodeReflexiveAddressIPv4.Addr().String() == parseIPfromAddrPort(reflexiveIP4)
+			return !nx.relay && !nx.relayOnly && !device.GetRelay() && nx.nodeReflexiveAddressIPv4.Addr().String() == parseIPfromAddrPort(reflexiveIP4)
 		},
 		buildPeerConfig: buildDirectLocalPeer,
 	},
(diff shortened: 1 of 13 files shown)


internal/nexodus/wg_peers.go
```diff
@@ -268,21 +269,7 @@ func (nx *Nexodus) buildPeersConfig() map[string]client.ModelsDevice {
...
-
-				if cmManual {
-					// If the cert mode is manual, we need to set the local DNS entry
-					// to point it to the provided hostname
-					ip, _, err := net.SplitHostPort(derpAddr)
-					if err != nil {
-						nx.logger.Errorf("Failed to parse stun address %s : %v", derpAddr, err)
-					}
-
-					err = SetLocalDerpDnsEntry(ip, hostname)
```
vishnoianil, line 281: @chirino I don't see this being set anywhere else (context/dnscache), so wondering how does it will resolve the dns request, we don't explicitly set it?

chirino: See: https://github.com/nexodus-io/nexodus/pull/1979/files#diff-7b699c477aec224ffe7d58ab48015134c0c67b55615982ccc78c62cfbdec0553R300-R303

We just set them using the extra hosts option when we spin up the test containers.

vishnoianil: yeah I saw that, but thats covers the test part. But we do need this when we deploy this in cluster. If user onboard it's own private derp relay using it's own certificates (--certmode manual), the dns won't be resolved by public relay, so currently we are using configuring that dns entry in the /etc/host. Sorry if you are already doing that in the PR and i missed it.

vishnoianil: We can totally ditch setting up entry in /etc/host, and asked user to do it by itself as well, but because we have a better visibility to reflexive address on the onboarded relay, it's makes things bit convenient for user, and they don't have to run the stun client etc to figure out the address.

vishnoianil: As you suggested on slack, i think we can use http dialer context to resolve the dns to the reflexive ip address of the onboarded derp relay, but unfortunately most of the connect related call is in derp package.

vishnoianil: I am just wondering if we can hook our own dns resolved to the derp http client 
https://github.com/nexodus-io/nexodus/blob/main/pkg/derp/derphttp/derphttp_client.go#L55

chirino: Yeah, I think I'm fine with letting users setup DNS themselves for now.  Using the stun discovered reflective IP is not a sure way to get the IP:port of the derp proxy so I think we should not do that as the primary resolution method.  If we could fallback to that (when the user did not setup DNS), then maybe that would be ok.

chirino: I updated the PR to do that.  We now set a fallback resolver that uses the reflexive IP of the derp relay.  So we should still have the previous functionality.  Sadly, the current e2e tests can't exercise it since we can't depend on using the reflexive IP on OSX.

vishnoianil (approved):

PR #1994, opened 2024-03-18
ci: Stop publishing rpms to the deprecated copr repo

### LINKED ISSUE #1139: Rename copr rpm repository
[Fedora copr](https://copr.fedorainfracloud.org/) allows users to manage custom rpm repositories and serves as a build service for those repos. Repos are namespace under fedora accounts. There's no concept of a shared organization. Because of this, the Nexodus rpm copr repo is `russellb/nexodus`. It would be better to make it `nexodus-io/nexodus`. That requires creating a new Fedora account for `nexodus-io` and maintaining the repos under that user, instead.

russellb:
We switched from russellb/nexodus to nexodus/nexodus in Nov 2023. The
key expired for the deprecated repo, so it's a good reminder to just
retire it.

Closes #1139


chirino (approved):
Lgtm

PR #2072, opened 2024-07-23
fix: Add patch version to go.mod

dave-tucker:
Without this go1.22 on Linux complains that it can't download the correct version of Go.


russellb (approved):
lgtm if you can get CI working ... or just merge manually if you want and CI failures should be fixed elsewhere

